PR #3, opened 2021-12-10
Setup GitHub actions

### LINKED ISSUE #2: Using CI job for uploading this compnents
It will be more convenient to upload these components to the registry with the Github Action.

tore-espressif:
Three notable changes (one per commit):
1. Test: `test_app` is added that compiles tests of testable components (runner will be set up later)
2. Miscellaneous: Add pre_commit and editorconfig for user's convenience
3. CI: Build `test_app` on PR and upload components on push to `master`

Todo list:

- [x] Set-up GitHub secret for component registry access

Closes #2


.github/workflows/upload_component.yml
```diff
@@ -0,0 +1,20 @@
...
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  upload_components:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
kumekay, line 12: Is there a reason for using this old version of checkout action? I would suggest using v2.  `actions/checkout@v2`  which is more flexible.

.github/workflows/upload_component.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          submodules: 'true'
```
kumekay, line 14: Don't we need `submodules: 'recursive'` here? 
I would suggest recursive here even if current components don't need it, because it can help avoiding Ooops in future

kumekay (approved):
Couple of tiny notes, LGTM, thank you @tore-espressif

.github/workflows/upload_component.yml
tore-espressif: `checkout@v2` requires git version >= 2.18 if submodules are fetched (https://github.com/actions/checkout#checkout-v2).

in idf docker image v4.3 this is not met :(

it is fixed in v4.4 https://github.com/espressif/esp-idf/commit/57b243a6999c8b1a4c31cf8ce6e6d715c51bb8c3

.github/workflows/upload_component.yml
tore-espressif: fixed!

.github/workflows/upload_component.yml
kumekay: Ok I understand this limitation  for `build_test.yml` workflow
But in this workflow, it runs directly in VM `runs-on: ubuntu-latest`, so it shouldn't be a problem.
It's ok however, to keep v1 here too.

tore-espressif: True, fixed

PR #4, opened 2021-12-31
Feature: Add nghttp and sh2lib components

mahavirj:
- Add nghttp (HTTP/2 C Library) component
- Add sh2lib (HTTP2 + TLS abstraction) component

### DIFF
--- a/nghttp/README.md
+++ b/nghttp/README.md
@@ -1,0 +1,1 @@
+nghttp2/README.rst
--- a/nghttp/idf_component.yml
+++ b/nghttp/idf_component.yml
@@ -1,0 +1,5 @@
+version: "1.41.0"
+description: "nghttp2 - HTTP/2 C Library"
+url: https://github.com/espressif/idf-extra-components/tree/master/nghttp
+dependencies:
+  idf: ">=5.0"
--- a/nghttp/port/private_include/config.h
+++ b/nghttp/port/private_include/config.h
@@ -1,0 +1,29 @@
+#ifndef __HAVE_CONFIG_H_
+#define __HAVE_CONFIG_H_
+
+#define _U_
+
+#define SIZEOF_INT_P 2
+
+#include "stdio.h"
+#include "stdlib.h"
+#include "string.h"
+
+#if (!defined(nghttp_unlikely))
+#define nghttp_unlikely(Expression)	!!(Expression)
+#endif
+
+#define nghttp_ASSERT(Expression) do{if (!(Expression)) printf("%d\n", __LINE__);}while(0)
+
+#define CU_ASSERT(a) nghttp_ASSERT(a)
+#define CU_ASSERT_FATAL(a) nghttp_ASSERT(a)
+
+#define NGHTTP_PLATFORM_HTONS(_n)  ((uint16_t)((((_n) & 0xff) << 8) | (((_n) >> 8) & 0xff)))
+#define NGHTTP_PLATFORM_HTONL(_n)  ((uint32_t)( (((_n) & 0xff) << 24) | (((_n) & 0xff00) << 8) | (((_n) >> 8)  & 0xff00) | (((_n) >> 24) & 0xff) ))
+
+#define htons(x) NGHTTP_PLATFORM_HTONS(x)
+#define ntohs(x) NGHTTP_PLATFORM_HTONS(x)
+#define htonl(x) NGHTTP_PLATFORM_HTONL(x)
+#define ntohl(x) NGHTTP_PLATFORM_HTONL(x)
+
+#endif // __HAVE_CONFIG_H_
--- a/sh2lib/idf_component.yml
+++ b/sh2lib/idf_component.yml
@@ -1,0 +1,6 @@
+version: "1.0.0"
+description: HTTP2 TLS Abstraction Layer
+url: https://github.com/espressif/idf-extra-components/tree/master/sh2lib
+dependencies:
+  idf: ">=5.0"
+  espressif/nghttp: ">=1.41.0"
--- a/sh2lib/sh2lib.h
+++ b/sh2lib/sh2lib.h
@@ -1,0 +1,268 @@
+/*
+ * SPDX-FileCopyrightText: 2027-2021 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "esp_tls.h"
+#include <nghttp2/nghttp2.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * This is a thin API wrapper over nghttp2 that offers simplified APIs for usage
+ * in the application. The intention of this wrapper is to act as a stepping
+ * stone to quickly get started with using the HTTP/2 client. Since the focus is
+ * on simplicity, not all the features of HTTP/2 are supported through this
+ * wrapper. Once you are fairly comfortable with nghttp2, feel free to directly
+ * use nghttp2 APIs or make changes to sh2lib.c for realising your objectives.
+ *
+ * TODO:
+ * - Allowing to query response code, content-type etc in the receive callback
+ * - A simple function for per-stream header callback
+ */
+/**
+ * @brief Handle for working with sh2lib APIs
+ */
+struct sh2lib_handle {
+    nghttp2_session *http2_sess;   /*!< Pointer to the HTTP2 session handle */
+    char            *hostname;     /*!< The hostname we are connected to */
+    struct esp_tls  *http2_tls;    /*!< Pointer to the TLS session handle */
+};
+
+/**
+ * @brief sh2lib configuration structure
+ */
+struct sh2lib_config_t {
+    const char *uri;                    /*!< Pointer to the URI that should be connected to */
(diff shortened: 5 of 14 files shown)


mahavirj:
Is there any way to upload these components (to say test registry) for test purpose?

nghttp/idf_component.yml
```diff
@@ -0,0 +1,5 @@
+version: "1.41.0"
+description: "nghttp2 - HTTP/2 C Library"
+url: https://github.com/espressif/idf-extra-components/tree/master/nghttp
+dependencies:
+  idf: ">=5.0"
```
tore-espressif, line 5: Is this 'just in case' requirement, or we really don't support it with IDF versions 4.x ?

nghttp/idf_component.yml
```diff
@@ -0,0 +1,5 @@
+version: "1.41.0"
```
tore-espressif, line 1: `nghttp2ver.h` files says v1.22.0

tore-espressif (commented):
Hi @mahavirj , thanks for the PR, few questions:

1. Why is the `nghttp` component different from this https://github.com/espressif/esp-idf/tree/master/components/nghttp ? Do you plan to leave `nghhtp` in esp-idf?
2. Could you please add your new components to this line? https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18
They will be automatically uploaded then

tore-espressif:
> Is there any way to upload these components (to say test registry) for test purpose?

We have a staging server https://staging.components.espressif.tools/ for testing. But your API token that you have for the production server will not work. you can ask @kumekay for access token

nghttp/idf_component.yml
mahavirj: This was auto-generated file which had not been updated for long time. I have added update to this file now.

mahavirj:
> Hi @mahavirj , thanks for the PR, few questions:
> 
>     1. Why is the `nghttp` component different from this https://github.com/espressif/esp-idf/tree/master/components/nghttp ? Do you plan to leave `nghhtp` in esp-idf?

Yes, IDF component will be removed.

> 
>     2. Could you please add your new components to this line? https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18
>        They will be automatically uploaded then

Fixed, thanks.

nghttp/idf_component.yml
mahavirj: We will be splitting `nghttp` from ESP-IDF into 2 component, `http_parser` and `nghttp` (vanilla version). This will be part of ESP-IDF v5.0 and hence have kept this requirement.

igrr: One minor complication I haven't thought of before, how do we support projects which should be buildable both with 4.x and 5.x, and which depend on this component?

If we add idf_component.yml to support building with IDF 5.0, and use it to install nghttp, then there will be an issue when building the same project with 4.4 — installation will fail.

@kumekay do IDF versions requirements participate in component version resolution?

I.e. can we:
- upload nghttp v0.0.0, which would be a no-op component (no sources, no headers) declared compatible with IDF 4.x
- upload nghttp v1.41.0, which would be the real nghttp component compatible with IDF 5.x
- depending on the user's IDF version, install one or the other.

mahavirj: > One minor complication I haven't thought of before, how do we support projects which should be buildable both with 4.x and 5.x, and which depend on this component?

One thing we could do is to backport the change where we split out `http_parser` component to v4.4/v4.3. Then the current PR can have "idf: >=4.3", and the application should build against 4.x and 5.x.

kumekay:
> > Is there any way to upload these components (to say test registry) for test purpose?
> 
> We have a staging server https://staging.components.espressif.tools/ for testing. But your API token that you have for the production server will not work. you can ask @kumekay for access token

@mahavirj  I sent you token in the mattermost

nghttp/README.md
```diff
@@ -0,0 +1 @@
+nghttp2/README.rst
```
kumekay, line 1: We render Markdown from readme on our web client, not sure how good will this particular rst file will look. I would suggest uploading this component to the staging first and checking the look

mahavirj: I have sent staging server link for this component (in unicast). Rendering looked good to me, you may check as well.

nghttp/idf_component.yml
mahavirj: Sounds good. We can go ahead with v0.0.0 trick then. Please help me with the upload after this PR gets merged. @kumekay

sh2lib/sh2lib.h
```diff
@@ -0,0 +1,260 @@
...
+ * SPDX-FileCopyrightText: 2027-2021 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "esp_tls.h"
+#include <nghttp2/nghttp2.h>
+
```
igrr, line 11: Could you add C++ guards to this header file?

sh2lib/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+version: "0.1.0"
+description: HTTP2 TLS Abstraction Layer
+url: https://github.com/espressif/idf-extra-components/tree/master/sh2lib
+dependencies:
+  idf: ">=5.0"
+  espressif/nghttp: ">=1.41.0"
```
igrr, line 6: I think if this component has a dependency on nghttp >= 1.41.0, then the trick with nghttp 0.0.0 won't work, will it?

It might indeed be simpler to publish http_parser component in the component manager just for 4.x versions to use...

nghttp/port/private_include/config.h
```diff
@@ -0,0 +1,36 @@
...
+#if (!defined(nghttp_unlikely))
+#define nghttp_unlikely(Expression)	!!(Expression)
+#endif
+
+#define nghttp_ASSERT(Expression) do{if (!(Expression)) printf("%d\n", __LINE__);}while(0)
+
+#define CU_ASSERT(a) nghttp_ASSERT(a)
+#define CU_ASSERT_FATAL(a) nghttp_ASSERT(a)
+
+#if 1
```
igrr, line 22: Is this a debugging leftover?

nghttp/idf_component.yml
kumekay: @mahavirj  I tried this approach and found an issue with the component manager that blocks the 0.0.0 trick right now. We will work on the fix

sh2lib/idf_component.yml
kumekay: We have a planned feature for the component manager to support IDF-linked dependencies and target linked, i.e. it will be possible to define a dependency like:

```
expressif/nghttp:
  version: ">=1.41.0"
  rules:
    idf: ">=5.0"
```
So it will only try to install this dependency only if the rule is satisfied. 

This feature is planned for this quarter, but I cannot say more precisely.

nghttp/idf_component.yml
kumekay: At the same time, I don't mind merging this as it is, with support for IDF 5.0+ only. 

> how do we support projects which should be buildable both with 4.x and 5.x,

@igrr  Which projects have this problem?

sh2lib/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+version: "0.1.0"
```
kumekay, line 1: Do we really want a version with a leading zero?  According to semver 2.0 https://semver.org/#spec-item-4 
Public API should be considered unstable at this point. 
If we are going to include it to IDF, I would suggest to release this with 
`version: 1.0.0` 

and add dependency to the IDF example as 
`espressif/sh2lib: "~1.0.0"` (>=1.0.0,<2.0.0)

nghttp/idf_component.yml
igrr: This was a hypothetical issue — sh2lib wasn't use much (it seems) so I don't know if there will be a real project which has this issue. However we should find a general solution for this issue and apply it for all the components we move out in v5.0. Any customer project which should be build-able with both 4.4 and 5.0 might be affected, and currently there are no clear instructions about making the projects compatible with both versions.

kumekay: In this case solution proposed in https://github.com/espressif/idf-extra-components/pull/4#discussion_r778121315 will work the best, however it's not implemented yet on the component manager side

sh2lib/idf_component.yml
mahavirj: Good point. Agree that we should start with version `1.0.0.` here. I have fixed this.

nghttp/port/private_include/config.h
mahavirj: Fixed, thanks.

sh2lib/sh2lib.h
mahavirj: Done.

nghttp/idf_component.yml
mahavirj: Alternatively, if we consider option from https://github.com/espressif/idf-extra-components/pull/4#discussion_r777405139 then that can also work. I have tried building application against v4.4 using same approach, it works. We can consider either of the approaches if requirement comes up.

PR #5, opened 2022-01-12
Feature: Add expat component

mahavirj:
This component is pretty independent and good candidate for component manager.


tore-espressif (changes requested):
Could you please update file https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt ? Something like this

```
set(EXTRA_COMPONENT_DIRS "../libsodium ../expat")

# Set the components to include the tests for.
set(TEST_COMPONENTS libsodium expat CACHE STRING "List of components to test")
```

We don't run these tests anywhere yet, they are only built.

I guess that the test build will fail because it is set for idf v4.3, so I'll fix it later

mahavirj:
@tore-espressif Added entry for this component to `test_app`. Also added partition table for increased firmware size and verified `expat` tests locally.

tore-espressif:
@mahavirj LGTM, thank you.
Do you want anybody else to review, or can I merge it now?

mahavirj:
> LGTM, thank you.

Thanks for approving, I will go ahead and merge it.

PR #7, opened 2022-01-24
ci: Run test_app in CI

tore-espressif:


tore-espressif:
Will be handled by #13

PR #13, opened 2022-02-17
Run Test Application on target

tore-espressif:
~~The artifacts will be used you 'run_test' action~~


tore-espressif:
FYI @igrr

tore-espressif:
@mahavirj @igrr @hfudev @tomassebestik @kumekay @Ouss4 PTAL

In this PR:
1. All components are build in CI (even if they don't have unit tests)
2. Unit test_app is run on our self-hosted runners. It uses pytest-embedded test framework.
3. Test results are parsed by github action bot.

.github/workflows/build_and_run_test_app.yml
```diff
@@ -0,0 +1,111 @@
...
+        with:
+          fetch-depth: 0
+      - uses: actions/download-artifact@v2
+        with:
+          name: test_app_bin_${{ matrix.idf_target }}_${{ matrix.idf_ver }}
+          path: test_app/build
+      - name: Install Python packages
+        env:
+          PIP_EXTRA_INDEX_URL: "https://www.piwheels.org/simple"
+        run: pip install esptool pytest-embedded==0.6.0rc0 pytest-embedded-serial-esp==0.6.0rc0
```
hfudev, line 87: nitpick, you can simplify it to `pip install esptool pytest-embedded-serial-esp==0.6.0rc0`

hfudev (commented):
nice job. LGTM!

.github/workflows/build_and_run_test_app.yml
```diff
@@ -0,0 +1,111 @@
...
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  build:
+    name: Build Test App
+    strategy:
+      matrix:
+        idf_ver: ["release-v4.3", "release-v4.4", "latest"]
+        idf_target: ["esp32", "esp32c3", "esp32s3"] # @todo ESP32S2 has less RAM and the test_app will not fit
```
igrr, line 13: One note about this — I think it's not very convenient to specify the target list for test apps here in CI configuration. It will probably get quite messy if the repository gets more test applications, and each test application will have its own set of targets.

We could consider using find_apps.py and build_apps.py scripts from IDF to perform builds for all targets, but in that case we would need to support the `merge_bin` step as part of build_apps.py.

(No action needed for now, just bringing this up. I'm afraid the missing tests for S2 will be easy to forget about.)

.github/workflows/build_and_run_test_app.yml
```diff
@@ -0,0 +1,111 @@
...
+          IDF_TARGET: ${{ matrix.idf_target }}
+        shell: bash
+        working-directory: test_app
+        run: |
+          . ${IDF_PATH}/export.sh
+          idf.py build
+      - name: Merge binaries
+        working-directory: test_app/build
+        run: |
+          pip install esptool
```
igrr, line 34: Don't we already have esptool.py in the container? Or is the issue that esptool.py in release/v4.3 doesn't support `merge_bin`?

test_app/main/test_app_main.c
```diff
@@ -11,6 +11,8 @@
 void app_main(void)
 {
     UNITY_BEGIN();
-    unity_run_menu();
+    //unity_run_tests_by_tag("[libsodium]", false);
+    unity_run_all_tests();
+    //unity_run_menu();
```
igrr, line 16: Are the commented out lines needed? If no, can they be removed?

(I wonder if the reason for keeping them is that you'd like to have different behavior when running tests in CI and when running them manually. In that case we could add some Kconfig option to the test app ("Run all tests automatically") and enable it only in CI builds.)

test_app/CMakeLists.txt
```diff
@@ -2,7 +2,16 @@
...
+    list(APPEND EXTRA_COMPONENT_DIRS ../pid_ctrl)
+endif()
+
+if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.0")
+    list(APPEND EXTRA_COMPONENT_DIRS ../sh2lib ../nghttp)
+endif()
+
 
 # Set the components to include the tests for.
 set(TEST_COMPONENTS libsodium expat esp_encrypted_img CACHE STRING "List of components to test")
```
igrr, line 17: I wonder what would happen if we set TEST_COMPONENTS equal to EXTRA_COMPONENT_DIRS (stripping the `../` prefix from each).
Would the build system complain that some components don't actually have a test subdirectory?

If it wouldn't this would be a convenient way to not forget to add new components to TEST_COMPONENTS later.

igrr (approved):
Looks very neat, I really like the idea to publish the test result in the PR comment.
All comments are pretty minor. You can note them down and fix in another PR if you like.

Just one stupid question: have you tested that CI job actually fails when one of the unit tests fails? Looking at the CI yml file, i think it should, but I'm not 100% sure.

.github/workflows/build_and_run_test_app.yml
```diff
@@ -0,0 +1,111 @@
...
+      matrix:
+        idf_ver: ["release-v4.3", "release-v4.4", "latest"]
+        idf_target: ["esp32", "esp32c3", "esp32s3"] # @todo ESP32S2 has less RAM and the test_app will not fit
+        exclude:
+          - idf_ver: "release-v4.3"
+            idf_target: esp32s3 # ESP32S3 support started with version 4.4
+    runs-on: ubuntu-20.04
+    container: espressif/idf:${{ matrix.idf_ver }}
+    steps:
+      - uses: actions/checkout@v1
```
Ouss4, line 20:
```suggestion
      - uses: actions/checkout@v2
```

.github/workflows/build_and_run_test_app.yml
```diff
@@ -0,0 +1,111 @@
...
+          idf.py build
+      - name: Merge binaries
+        working-directory: test_app/build
+        run: |
+          pip install esptool
+          python -m esptool --chip esp32 merge_bin --fill-flash-size 4MB -o flash_image.bin @flash_args
+      - uses: actions/upload-artifact@v2
+        with:
+          name: test_app_bin_${{ matrix.idf_target }}_${{ matrix.idf_ver }}
+          path: test_app/build/flash_image.bin
```
Ouss4, line 39: For my comment above of `pytest-embedded` is able to flash the binary, we gonna have to upload the `flasher_args.json` file too.

.github/workflows/build_and_run_test_app.yml
kumekay: In the case of outdated esptool it should be `pip install -U esptool`

.github/workflows/build_and_run_test_app.yml
```diff
@@ -0,0 +1,111 @@
...
+        shell: bash
+        working-directory: test_app
+        run: |
+          . ${IDF_PATH}/export.sh
+          idf.py build
+      - name: Merge binaries
+        working-directory: test_app/build
+        run: |
+          pip install esptool
+          python -m esptool --chip esp32 merge_bin --fill-flash-size 4MB -o flash_image.bin @flash_args
```
kumekay, line 35: Should not it be `--chip ${{ matrix.idf_target }}`?

.github/workflows/build_and_run_test_app.yml
tore-espressif: After this change, the latest stable `pytest-embedded` is installed (0.5.1) which does not contain some of the bugfixes in need.

tore-espressif:
> > Just one stupid question: have you tested that CI job actually fails when one of the unit tests fails? Looking at the CI yml file, i think it should, but I'm not 100% sure.
> 
> I think it should be that `pytest` returns a non-zero value for a failed test, that would make its step and workflow fail too.

Yes, the CI failed in case of failed tests. However, the test results were not propagated to next steps in case of failure. It is fiexd now.

See here for results with tampered expat test https://github.com/espressif/idf-extra-components/runs/5405924004

test_app/CMakeLists.txt
tore-espressif: This is a very good idea. The build system doesn't mind the non-existent `test` folder.

CMake updated.

.github/workflows/build_and_run_test_app.yml
tore-espressif: works fine with calling `esptool.py` directly.

I just didn't realize that I must run `. ${IDF_PATH}/export.sh` for each step explicitly...

tore-espressif:
@kumekay @igrr @Ouss4 @hfudev Thank you very much for your reviews!

I've updated the PR accordingly, please let me know whether you agree with the proposed resolutions. Thanks

tore-espressif:
I'll merge today, if there are no objections

PR #17, opened 2022-03-05
Some fixes/enhancements for esp_encrypted_img

MacDue:
Love the `esp_encrypted_img` component and I've been trying it out in a C++ ESP IDF project, 
this PR just addresses a few small issues I found.

The first is that the `esp_encrypted_img_<whateever>` functions currently accept `esp_decrypt_handle_t*` parameter rather than a `esp_decrypt_handle_t`.
This is not correct and produces compiler errors in C++ (`esp_decrypt_handle_t` is already the handle pointer) 

The next is that there seemed a bunch of unnecessary memory allocations with (`calloc()`), moving the `gcm_key` and `iv` directly into the `esp_encrypted_img_handle` only adds 48 bytes to it (and it is already dynamically allocated), 
and also simplifies the setup/error handling. Likewise, `got_auth` is only 16 bytes and can be allocated on the stack.

Finally, there was a possible memory leak of `rsa_pem`, if `esp_encrypted_img_decrypt_end()` was called without doing any decryption. This might happen if you wanted to cancel the decryption (e.g. a network error meant you could not download the firmware).

### DIFF
--- a/esp_encrypted_img/idf_component.yml
+++ b/esp_encrypted_img/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.0.3"
+version: "2.0.0"
 description: ESP Encrypted Image Abstraction Layer
 url: https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img
 dependencies:
--- a/esp_encrypted_img/idf_component.yml
+++ b/esp_encrypted_img/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.0.2"
+version: "1.0.3"
 description: ESP Encrypted Image Abstraction Layer
 url: https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img
 dependencies:
--- a/esp_encrypted_img/src/esp_encrypted_img.c
+++ b/esp_encrypted_img/src/esp_encrypted_img.c
@@ -29,20 +29,6 @@ typedef enum {
     ESP_PRE_ENC_DATA_DECODE_STATE,
 } esp_encrypted_img_state;
 
-struct esp_encrypted_img_handle {
-    const char *rsa_pem;
-    size_t rsa_len;
-    uint32_t binary_file_len;
-    uint32_t binary_file_read;
-    char *gcm_key;
-    char *iv;
-    char auth_tag[16];
-    esp_encrypted_img_state state;
-    mbedtls_gcm_context gcm_ctx;
-    size_t cache_buf_len;
-    char *cache_buf;
-};
-
 #define GCM_KEY_SIZE        32
 #define MAGIC_SIZE          4
 #define ENC_GCM_KEY_SIZE    384
@@ -50,6 +36,20 @@ struct esp_encrypted_img_handle {
 #define BIN_SIZE_DATA       4
 #define AUTH_SIZE           16
 #define RESERVED_HEADER     88
+
+struct esp_encrypted_img_handle {
+    const char *rsa_pem;
+    size_t rsa_len;
+    uint32_t binary_file_len;
+    uint32_t binary_file_read;
+    char gcm_key[GCM_KEY_SIZE];
+    char iv[IV_SIZE];
+    char auth_tag[AUTH_SIZE];
+    esp_encrypted_img_state state;
+    mbedtls_gcm_context gcm_ctx;
+    size_t cache_buf_len;
+    char *cache_buf;
+};
 
 typedef struct {
     char magic[MAGIC_SIZE];
@@ -69,10 +69,6 @@ typedef struct esp_encrypted_img_handle esp_encrypted_img_t;
 static int decipher_gcm_key(char *enc_gcm, esp_encrypted_img_t *handle)
 {
     int ret = 1;
-    handle->gcm_key = calloc(1, GCM_KEY_SIZE);
-    if (!handle->gcm_key) {
-        return ESP_ERR_NO_MEM;
-    }
     size_t olen = 0;
     mbedtls_pk_context pk;
     mbedtls_entropy_context entropy;
@@ -87,7 +83,6 @@ static int decipher_gcm_key(char *enc_gcm, esp_encrypted_img_t *handle)
                                       &entropy, (const unsigned char *) pers,
                                       strlen(pers))) != 0) {
         ESP_LOGE(TAG, "failed\n  ! mbedtls_ctr_drbg_seed returned -0x%04x\n", (unsigned int) - ret);
-        free(handle->gcm_key);
         goto exit;
     }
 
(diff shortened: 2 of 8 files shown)


mahavirj:
@MacDue 

Thanks for sending across this PR!

Changes generally looks good to me.

> Finally, there was a possible memory leak of rsa_pem, if esp_encrypted_img_decrypt_end() was called without doing any decryption. This might happen if you wanted to cancel the decryption (e.g. a network error meant you could not download the firmware).

Could you please check if we can add test case for this in https://github.com/espressif/idf-extra-components/blob/master/esp_encrypted_img/test/test.c? 

Test application can be found at https://github.com/espressif/idf-extra-components/tree/master/test_app, you can build it and execute tests on ESP target.

mahavirj:
@hmalpani Can you please review as well?

MacDue:
I've added a test for it with `xPortGetFreeHeapSize()` before/after `esp_encrypted_img_decrypt_start/end()`, this fails before these changes but passes now. I've allowed for some error in this test since I'm not sure how brittle testing this way is.

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -452,6 +450,7 @@ esp_err_t esp_encrypted_img_decrypt_end(esp_decrypt_handle_t ctx)
 exit:
     mbedtls_gcm_free(&handle->gcm_ctx);
     free(handle->cache_buf);
+    free(handle->rsa_pem);
```
hmalpani:
```suggestion
    if (handle->rsa_pem) {
        free(handle->rsa_pem);
    }

Comment (MacDue) [reply esp_encrypted_img/src/esp_encrypted_img.c]: Freeing a `NULL` pointer is a no-op, I make sure to set the pointer NULL when it's free'd early

**Comment by MacDue** [comment]
Pre-commit wanted to change all the line ending (sorry about that, it'd kind of ruined the diff)

File: esp_encrypted_img/src/esp_encrypted_img.c (hunk 1 above)
Comment (hmalpani) [reply esp_encrypted_img/src/esp_encrypted_img.c]: `handle->rsa_pem` will be freed after decrypting gcm key. Freeing it here again might cause an error.

**Comment by hmalpani** [comment]
@MacDue Changes looks good to me

**Review by tore-espressif — CHANGES_REQUESTED** [review]
@MacDue Could you please rebase on master? #13 was merged and now we run the tests in GH action, thank you!

**Comment by MacDue** [comment]
Rebased :+1:

**Comment by tore-espressif** [comment]
> Rebased 👍

The build is failing but it should be only on `latest` esp-idf. I'm checking it now

File: esp_encrypted_img/test/test.c [hunk 2]
```diff
@@ -304,3 +306,20 @@ TEST_CASE("Sending random size data at once", "[encrypted_img]")
...
     free(args);
 }
+
+TEST_CASE("Test canceling decryption frees memory", "[encrypted_img]")
+{
+    esp_decrypt_cfg_t cfg = {
+        .rsa_pub_key = (char *)rsa_private_pem_start,
+        .rsa_pub_key_len = rsa_private_pem_end - rsa_private_pem_start,
+    };
+    int free_bytes_start = xPortGetFreeHeapSize();
```
mahavirj, line 316: Just note: there is also `[leaks]` tag that we can in `TEST_CASE` declaration. We will improve this further, for now we can accept current solution.

MacDue:
> The build is failing but it should be only on `latest` esp-idf. I'm checking it now

Looks like `randombytes_esp32.c` needs to include `esp_random.h` (looks like that include was removed from `esp_system.h`)

I could try pushing a quick fix here?

esp_encrypted_img/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.0.2"
+version: "1.0.3"
```
mahavirj, line 1: Though this PR does not introduce any breaking change, but still API is being modified here.

I think that we should bump up `major` (that is to make it `2.0.0`) version here.

This shall allow for time window to:

 - Modify example in ESP-IDF at https://github.com/espressif/esp-idf/tree/master/examples/system/ota/pre_encrypted_ota
 - Update its version dependency at https://github.com/espressif/esp-idf/blob/0440aca4e3c0608f13a6980c339e32587d78bf91/examples/system/ota/pre_encrypted_ota/main/idf_component.yml#L3 to `^2.0.0.`

(Until these changes, ESP-IDF will safely continue to use existing version of this component)

@MacDue What do you think?

tore-espressif:
> Looks like randombytes_esp32.c needs to include esp_random.h (looks like that include was removed from esp_system.h)
>
> I could try pushing a quick fix here?

Yep, that must be it. If you could bump libsodium version too, that would be great :)

esp_encrypted_img/idf_component.yml
MacDue: I think that's fine, I'll change that now :+1:

MacDue:
I've switched to use `#if __has_include("esp_random.h")` switch should switch back to `esp_system.h` on older IDF versions

MacDue:
I've switched to use `#if __has_include("esp_random.h")` switch should switch back to `esp_system.h` on older IDF versions

esp_encrypted_img/.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
```
igrr, line 1: looks like this file was committed by accident?

tore-espressif:
@MacDue I'm sorry about the broken pipeline. [Python:3.7-slim-buster](https://hub.docker.com/layers/python/library/python/3.7-slim-buster/images/sha256-4747aa18694bd9c1fd1bc2e3c5873dc3d44d27c7788974466247928bf8f3a705?context=explore) docker image was updated this morning and it breaks our pipeline.

You can cherry-pick https://github.com/espressif/idf-extra-components/pull/18/commits/fbec8f5661592d6a09f58f68fd4a689a8aec3361 and it should work again.

MacDue:
Added the fix for the test runners

tore-espressif (approved):

mahavirj:
@MacDue 

- Could you please confirm requirement to change `libsodium` submodule update from this PR? I do not see any stable release after `1.0.18` from https://github.com/jedisct1/libsodium/tags

b1ackviking:
This PR also closes #19

MacDue:
@mahavirj I simply pulled the latest `stable` branch of libsodium as @tore-espressif asked to bump the version 

Also it seems like libsodium no regards the -stable branch as the latest release see comment: https://github.com/jedisct1/libsodium/issues/1041#issuecomment-799812417

mahavirj:
> @mahavirj I simply pulled the latest `stable` branch of libsodium as @tore-espressif asked to bump the version
> 
> Also it seems like libsodium now regards the -stable branch as the latest release see comment: [jedisct1/libsodium#1041 (comment)](https://github.com/jedisct1/libsodium/issues/1041#issuecomment-799812417)

Okay. Thanks for sharing details. In that case, we can go ahead with this change.

@tore-espressif Please take care of bumping up component version in followup PR for `libsodium`.

mahavirj (approved):

PR #27, opened 2022-03-18
ci: Run nightly tests

### LINKED ISSUE #26: Add a nightly build job with `master` branch of IDF
To detect regressions or incompatibilities in IDF `master` branch quicker, let's add `on: schedule` to https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/build_and_run_test_app.yml.

tore-espressif:
This cathces regressions on 4.3, 4.4 and latest release branches

Closes #26

### DIFF
--- a/.github/workflows/build_and_run_test_app.yml
+++ b/.github/workflows/build_and_run_test_app.yml
@@ -1,6 +1,8 @@
 name: Build and Run Test Application
 
 on:
+  schedule:
+    - cron: '0 0 * * *' # Once per day at midnight
   pull_request:
     types: [opened, reopened, synchronize]
 
@@ -8,6 +10,7 @@ jobs:
   build:
     name: Build Test App
     strategy:
+      fail-fast: false
       matrix:
         idf_ver: ["release-v4.3", "release-v4.4", "latest"]
         idf_target: ["esp32", "esp32c3", "esp32s3"] # @todo ESP32S2 has less RAM and the test_app will not fit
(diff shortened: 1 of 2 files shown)


igrr (approved):

.github/workflows/build_and_run_test_app.yml
```diff
@@ -1,13 +1,16 @@
...
+  schedule:
+    - cron: '0 0 * * *' # Once per day at midnight
   pull_request:
     types: [opened, reopened, synchronize]
 
 jobs:
   build:
     name: Build Test App
     strategy:
+      fail-fast: false
```
igrr, line 13: Not sure if this is possible or easy, but keeping fail-fast: true does make sense for PRs. For schedules indeed fail-fast: false is more useful. Perhaps `${{ github.event_name == 'pull_request' }}` would do the trick?

tore-espressif: I'll give it a try. Since we're talking about making it faster: Do we need nightly builds/tests again release 4.3 and 4.4 branches? Or is latest enough?

igrr: I think as for the nightly build, time doesn't matter that much, since there's no developer eagerly waiting for their pipeline to pass. It's just a comment related to pull requests.

tore-espressif: I'd prefer leaving `fail-fast: false` for all cases. It will help contributors see whether the compilation/test issues is global, or specific to certain IDF version.

igrr: Okay, let's keep it this way

PR #37, opened 2022-04-01
json: Add json_generator and json_parser components

shahpiyushv:


.gitmodules
```diff
@@ -13,3 +13,9 @@
 [submodule "coap/libcoap"]
 	path = coap/libcoap
 	url = https://github.com/obgm/libcoap.git
+[submodule "json_generator/json_generator"]
+	path = json_generator/upstream
+	url = https://github.com/espressif/json_generator.git
+[submodule "json_parser/upstream"]
+	path = json_parser/upstream
+	url = https://github.com/espressif/json_parser.git
```
tore-espressif, line 21: Could we upload the components directly from these repositories? instead of creating wrappers here?

json_generator/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2020 Piyush Shah <shahpiyushv@gmail.com>
```
tore-espressif, line 189: Could you please explain why isn't this licensed by Espressif?

tore-espressif (changes requested):

.gitmodules
shahpiyushv: The above repos are just mirrors of the original repos here, which are independent of esp-idf: https://github.com/shahpiyushv/json_parser, https://github.com/shahpiyushv/json_generator

json_generator/LICENSE
shahpiyushv: I just copied LICENSE from original sources, which were my personal projects outside Espressif. If our standard practice is to not retain original License, but instead add Espressif license, I can do so. However, I just referred the license files from other components and saw that it was not the case.

igrr: Keeping the original copyright notice looks fine to me!

.gitmodules
igrr: However the component URLs in idf_component.yml point to Espressif forks (e.g. https://github.com/espressif/json_generator).

Can we either redirect the submodule to Espressif fork or update the URLs?

igrr: My bad, the submodule does point to Espressif fork, so everything looks to be in order.

The relationship between three repos is indeed a bit complicated, but let's keep it as proposed by @shahpiyushv for now. If it gets difficult to manage we can revisit this.

tore-espressif:
I can see that https://github.com/espressif/json_parser uses `jsmn` which is already in our component registry https://components.espressif.com/component/espressif/jsmn

It would be nice if we could define dependency of this componet in `idf_component.yml`

But first I need to update https://github.com/espressif/idf-extra-components/tree/master/jsmn to version 1.1

shahpiyushv:
> I can see that https://github.com/espressif/json_parser uses `jsmn` which is already in our component registry https://components.espressif.com/component/espressif/jsmn
> 
> It would be nice if we could define dependency of this componet in `idf_component.yml` (user's project -> json_parser -> jsmn)
> 
> But first I need to update https://github.com/espressif/idf-extra-components/tree/master/jsmn to version 1.1

@tore-espressif , the original json_parser repo is independent of esp-idf and so, the jsmn submodule itself can't be removed from there. If we want to use the jsmn from idf-extra-components, we will have to include the individual source files of json_parser, instead of the complete json_parser repository. Is this what you mean? Or is there a different approach?

igrr:
> If we want to use the jsmn from idf-extra-components, we will have to include the individual source files of json_parser, instead of the complete json_parser repository. Is this what you mean? Or is there a different approach?

As far as I can tell we can use jsmn component without removing jsmn submodule from the upstream json_parser repository. I think it is sufficient to modify CMakeLists.txt of json_parser (in this PR) removing the component root directory from the list of include directories. Adding component root directory to the list of include directories seems to be the only thing that makes json_parser.c use the contents of jsmn submodule (due to `include <jsmn/jsmn.h>`).
If we remove this include directory from CMakeLista.txt, Jsmn submodule will still be there, but it won't be used. Then we can add a dependency on jsmn component to idf_componen.yml.

(Updating jsmn component to the required version as pointed out by @tore-espressif is also necessary, in this case)

tore-espressif:
I opened #42 in case you want to use JSMN v1.1.0.

But I'm fine with letting it as it is

tore-espressif (approved):

tore-espressif:
@shahpiyushv sorry, I forgot to follow this up. 
Do you have any news?

shahpiyushv:
@tore-espressif As discussed earlier, I have copied the sources of json_generator and json_parser into this and added a dependency for jsmn 1.1

tore-espressif:
@shahpiyushv the formatting check is failing, could you please have a look?

Apart from that, is this ready to merge?

shahpiyushv:
> @shahpiyushv the formatting check is failing, could you please have a look?
> 
> Apart from that, is this ready to merge?

Yes, it is ready. I have tried copy pasting these into RainMaker to validate that the code compilation and runtime working is fine.

PR #64, opened 2022-07-06
Ignore format warnings

Lapshin:
This is the first step to fix IDF's format warnings.

IDF has globally enabled compiler option `-Wno-format`. The fix will be in three steps:
1. add `-Wno-format` to affected components
2. disable the global option  `-Wno-format`
3. fix format warnings for each component

### DIFF
--- a/coap/CMakeLists.txt
+++ b/coap/CMakeLists.txt
@@ -42,3 +42,4 @@ idf_component_register(SRCS "${srcs}"
 #
 # TODO: find a way to move this to a port header
 target_compile_definitions(${COMPONENT_LIB} PUBLIC WITH_POSIX)
+target_compile_options(${COMPONENT_LIB} PRIVATE "-Wno-format")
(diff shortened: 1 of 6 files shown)


igrr:
@Lapshin thanks for the PR!

In theory we only need to add -Wno-format for IDF v5.0, but that's probably an unnecessary complication. LGTM!

Please open an issue in the "issues" section of this repository to remind us to do the last step (fix the warnings and remove this flag).

igrr:
By the way, to pass IDF CI you may need to bump the version numbers of some of these 3 components. 

When the PR is merged, this will cause the CI pipeline to release new versions to the component registry.

Lapshin:
> @Lapshin I'm not 100% sure, but the CI build job in this project should not fail on warnings. If that's the case, should we skip step 1 and 2, and move to step 3?

In IDF we have a line https://github.com/espressif/esp-idf/blob/master/tools/cmake/build.cmake#L99
Which will play after removing https://github.com/espressif/esp-idf/blob/master/tools/cmake/build.cmake#L108-L109

tore-espressif:
@Lapshin is this ready to be merged?

PR #66, opened 2022-07-08
usb_host: Add CDC-ACM driver

tore-espressif:
Move USB Host CDC driver to idf-extra-components

1. CDC driver copy pasted (with history) from esp-idf (including bugfix commit https://github.com/tore-espressif/esp-idf/commit/438234ad702fcd8ddc9781dcb2782f15a014d235)
2. Copy USB test_app to idf-extra-components

- [x] Change name of the uploaded package to usb_host_cdc_acm

* USB tests are not build/run in this PR yet
* Component uploaded to stagging server for testing https://staging.components.espressif.tools/component/espressif/usb_host_cdc_acm
* This component should then be used in https://github.com/espressif/esp-protocols/pull/25


tore-espressif:
@igrr @mavalki @dazza0 Could you share your ideas (location, time schedule...) about moving USB Class drivers out of esp-idf?

(tests will be migrated in follow-up MR)

tore-espressif:
@igrr

> Would it help if we preserved the commit history of the original component?

Good point, history added.

> what about using the same directory name in this repository?

Yep, I changed the component directory name to `usb_host_cdc_acm` and the base folder to `usb` so it can contain other usb components in the future (Trace, OTG...)

PR #75, opened 2022-07-27
coap: Update libcoap library to commit 8fbe440

mrdeep1:
# Checklist

- [ ] Component contains License
- [ ] Component contains README.md
- [ ] Component contains idf_component.yml file with `url` field defined
- [ ] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [ ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description

Fixes #74

Provides new functionality to build Client Only or Server Only code.

Provides new functionality to include TCP (and hence TLS) support.

Addresses coap component of #65.

PR #76, opened 2022-07-29
coap: ignore format warnings

Lapshin:
Added `-Wno-format` argument which was accidentally removed with https://github.com/espressif/idf-extra-components/pull/75

See https://github.com/espressif/idf-extra-components/pull/64 for more details

### DIFF
--- a/coap/CMakeLists.txt
+++ b/coap/CMakeLists.txt
@@ -37,4 +37,5 @@ set(srcs
 idf_component_register(SRCS "${srcs}"
                     INCLUDE_DIRS "${include_dirs}"
                     REQUIRES lwip mbedtls)
+target_compile_options(${COMPONENT_LIB} PRIVATE "-Wno-format")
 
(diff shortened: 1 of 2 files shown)


mrdeep1:
Apologies for the confusion.  I will look at libcoap code and fix why it needs this option in place.  It all worked when I tested - presumably because the global option was still in place.

igrr:
No problem at all! If you want to test it, you can explicitly set `-Wformat -Werror=format` flags for libcoap to make sure the warnings are reported.

The source of the warnings is the definition of `uint32_t`:
- In IDF 4.x, for Xtensa architecture, `uint32_t` was defined as `unsigned int`. So format string `%u` or `%d` worked.
- In IDF 4.x, for RISC-V architecture, `uint32_t` was defined as `unsigned long`. The correct format string is `%lu` or `%ld` in this case. We had format warnings disabled for RISC-V for this reason.
- In IDF 5.x, we are switching the definition of `uint32_t` for Xtensa to `unsigned long`. This matches RISC-V, ARM, and also the upstream Xtensa GCC. So now all architectures require `%lu` or `%ld`.

To support different architectures and compiler/IDF versions (and probably other frameworks/chips/architectures you need to support), it is recommended to use inttypes.h macros, like `PRIu32`.

PR #83, opened 2022-08-08
ci: build with extra warnings and -Werror

igrr:
Build with -Werror to prevent new warnings from being introduced. The set of extra warning flags matches ESP-IDF internal CI builds.

- [x] build passes with extra flags
- [x] build fails when a warning is added (https://github.com/espressif/idf-extra-components/runs/7720984491?check_suite_focus=true#step:4:1252)

Closes https://github.com/espressif/idf-extra-components/issues/79


tore-espressif:
LGTM!

PR #84, opened 2022-08-10
bdc_motor: added BDC motor control library

suda-morris:
Copied from [the BDC example](https://github.com/espressif/esp-idf/tree/master/examples/peripherals/mcpwm/mcpwm_bdc_speed_control/components/bdc_motor)

The interface there should be much close to a "stable" one, so it's time to publish it to the component registry.

No test_app is added for now, but we have one test in the esp-idf example.

### DIFF
--- a/bdc_motor/idf_component.yml
+++ b/bdc_motor/idf_component.yml
@@ -1,0 +1,5 @@
+version: "0.1.0"
+description: Brushed DC Motor Control Driver
+url: https://github.com/espressif/idf-extra-components/tree/master/bdc_motor
+dependencies:
+  idf: ">=5.0"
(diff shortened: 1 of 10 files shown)


bdc_motor/idf_component.yml
```diff
@@ -0,0 +1,5 @@
+version: "0.1.0"
```
tore-espressif, line 1: If you don't expect any changes in foreseeable future, I'd suggest to start with 1.0.0. But it is up to you

tore-espressif (approved):
@suda-morris Thanks for the PR, LGTM!

bdc_motor/idf_component.yml
suda-morris: IMO, the APIs there are close to stable, but at the moment, this is still a very new library, better to give the users some time to test it and feedback.

PR #90, opened 2022-09-08
usb_host_msc: Fix compatibility with IDF v4.4

MaValki:
Since API of `fatfs` has changed between IDF versions v4.4 and 5.0, compile time condition was added in order to cover both.

### DIFF
--- a/usb/usb_host_msc/src/msc_host_vfs.c
+++ b/usb/usb_host_msc/src/msc_host_vfs.c
@@ -13,6 +13,7 @@
 #include "diskio_impl.h"
 #include "ffconf.h"
 #include "ff.h"
+#include "esp_idf_version.h"
 
 #define DRIVE_STR_LEN 3
 
@@ -33,8 +34,14 @@ static esp_err_t msc_format_storage(size_t block_size, size_t allocation_size, c
 
     // Valid value of cluster size is between sector_size and 128 * sector_size.
     size_t cluster_size = MIN(MAX(allocation_size, block_size), 128 * block_size);
+
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
+    FRESULT err = f_mkfs(drv, FM_ANY | FM_SFD, cluster_size, workbuf, workbuf_size);
+#else
     const MKFS_PARM opt = {(BYTE)(FM_ANY | FM_SFD), 0, 0, 0, cluster_size};
     FRESULT err = f_mkfs(drv, &opt, workbuf, workbuf_size);
+#endif
+
     if (err) {
         ESP_LOGE(TAG, "Formatting failed with error: %d", err);
         free(workbuf);
--- a/.github/workflows/build_and_run_test_app_usb.yml
+++ b/.github/workflows/build_and_run_test_app_usb.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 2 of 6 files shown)


MaValki:
@tore-espressif PTAL.

usb/usb_host_msc/src/msc_host_vfs.c
```diff
@@ -33,8 +33,14 @@ static esp_err_t msc_format_storage(size_t block_size, size_t allocation_size, c
 
     // Valid value of cluster size is between sector_size and 128 * sector_size.
     size_t cluster_size = MIN(MAX(allocation_size, block_size), 128 * block_size);
+
+#if ESP_IDF_VERSION <= ESP_IDF_VERSION_VAL(4, 4, 2)
```
igrr, line 37: Since there will probably be a v4.4.3 with the same API as v4.4.2, suggest rephrasing this as:
```suggestion
#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
```

igrr:
Test failures seem to be unrelated. I've opened https://github.com/espressif/idf-extra-components/pull/91 with a probable fix.

usb/usb_host_msc/src/msc_host_vfs.c
MaValki: Good point

usb/usb_host_msc/src/msc_host_vfs.c
```diff
@@ -33,8 +33,14 @@ static esp_err_t msc_format_storage(size_t block_size, size_t allocation_size, c
 
     // Valid value of cluster size is between sector_size and 128 * sector_size.
     size_t cluster_size = MIN(MAX(allocation_size, block_size), 128 * block_size);
+
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
```
tore-espressif, line 37: Could you please add `#include "esp_idf_version.h"` to the begginig of the file?

The `ESP_IDF_VERSION_VAL` definition has somehow found its way to this file, but it'd be better to have the include dfined.

tore-espressif (approved):
On minor comment, otherwise LGTM!

tore-espressif:
@MaValki I'm sorry, I forgot about CI. Could you please add `release-v4.4` to this list?

https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/build_and_run_test_app_usb.yml#L17

igrr:
@tore-espressif that was well noticed! It looks like some format strings were adjusted to work with IDF v5.0 (where uint32_t == unsigned long), and now fail on v4.4 (where uint32_t == unsigned int), e.g.:
```
printf("\tdwClockFrequency %lu\n", desc->dwClockFrequency);
```
The recommended way to fix these is to use format specifier macros from inttypes.h:
```
#include <inttypes.h>
....
printf("\tdwClockFrequency %" PRIu32 "\n", desc->dwClockFrequency);
```

MaValki:
@igrr Thank you clarifying format macros. For the sake of consistency, I would replace them across the whole source file. However , I am not sure whether this is now preferred way of formatting strings (for v4.4 and v5.0).

igrr:
@MaValki Yes, you can replace all the cases when uint32_t or int32_t is formatted with the macros from inttypes.h. This will work with both 4.x and 5.x. Other types (plain `int` or `unsigned`) can be left as is.

MaValki:
@tore PTAL on changes in tests.

PR #93, opened 2022-09-12
USB: Virtual COM Ports drivers

tore-espressif:
# Change description
This PR adds Virtual COM Ports drivers for usually used USB devices. The drivers are based on examples https://github.com/espressif/esp-idf/tree/master/examples/peripherals/usb/host/cdc/cdc_acm_vcp

Notable changes (compared to the example in esp-idf):
* Lifetime of the drivers is now managed by smart pointer. _EDIT:_ After a review, it was decided that we won't force users to use smart pointers. But the drivers are intended for them.
* Added new class `VCP` that manages all types of VCP devices; so called VCP service

# Information for reviewers:
* **There is an example app `vcp_example` that shows intended usage** with some useful comments
* VCP drivers have layered architecture, here is a quick summary for your convenience:

| Layer no. |       Name       |        Location        |        Language       | Remarks                                              |
|:---------:|:----------------:|:----------------------:|:---------------------:|------------------------------------------------------|
|     1     | usb_host_vcp     | IDF component registry | C++                   | VCP service. Optional, does not have to be used                   |
|     2     | usb_host_xxx_vcp | IDF component registry | C++                   | Several independent packages for specific VCP device |
|     3     | usb_host_cdc_acm | IDF component registry | C, offers C++ wrapper | Base class implementing USB-CDC class                |
|     4     | usb_host_lib     | esp-idf                | C                     | Library for USB Host on ESP32-Sx chips               |

* I don't expect you to go through the whole code, but please focus on C++ exceptions and smart pointers, as this is the first time I'm using them
* The goal is to offer unified interface to users. They don't have to worry about which specific USB<->UART converter was plugged into the USB host port. The VCP always picks the correct driver.

# Checklist

- [x] CI passing


tore-espressif:
Latest changes:
1. removed strict smart_pointer. Now it is up to the user whether he wants to use smart pointers or not
2. Removed ugly function pointer typedef for factory methods
3. Unnecessary std::vectors turned to constexpr std::arrays

tore-espressif:
@euripedesrocha PTAL

PR #95, opened 2022-09-16
Add Eigen port to ESP

dmitry1945:
# Checklist

- [x] CI passing

# Change description
Add Eigen library to the IDF.
Component working with Eigen library below version 3.4.

### DIFF
--- a/eigen/CMakeLists.txt
+++ b/eigen/CMakeLists.txt
@@ -1,0 +1,54 @@
+idf_component_register(
+                       # We need the dummy source file so that the component
+                       # library is not an interface library. This allows to
+                       # get the list of include directories from other components
+                       # via INCLUDE_DIRECTORIES property later on.
+                       SRCS dummy.c)
+
+# Determine compilation flags used for building Eigen
+# Flags inherited from IDF build system and other IDF components:
+set(idf_include_directories $<TARGET_PROPERTY:idf::eigen,INCLUDE_DIRECTORIES>)
+set(includes "-I$<JOIN:${idf_include_directories}, -I>")
+
+if(CONFIG_COMPILER_OPTIMIZATION_DEFAULT)
+    set(opt_args -DCMAKE_BUILD_TYPE=Debug)
+elseif(CONFIG_COMPILER_OPTIMIZATION_SIZE)
+    set(opt_args -DCMAKE_BUILD_TYPE=Relesee)
+elseif(COMPILER_OPTIMIZATION_PERF)
+    set(opt_args -DCMAKE_BUILD_TYPE=Relesee)
+elseif(COMPILER_OPTIMIZATION_NONE)
+    set(opt_args -DCMAKE_BUILD_TYPE=Debug)
+else()
+    message(FATAL_ERROR "Unsupported optimization level")
+endif()
+
+include(ExternalProject)
+
+# Build Eigen in this directory:
+set(BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/eigen-build)
+
+# Add Eigen as a subproject.
+ExternalProject_Add(eigen_proj
+    SOURCE_DIR ${COMPONENT_DIR}/eigen
+    BINARY_DIR ${BINARY_DIR}
+    # These two options are set so that Ninja immediately outputs
+    # the subproject build to the terminal. Otherwise it looks like the
+    # build process "hangs" for too long until Eigen build is complete.
+    USES_TERMINAL_CONFIGURE TRUE
+    USES_TERMINAL_BUILD TRUE
+    # Arguments to pass to Eigen CMake invocation:
+    CMAKE_ARGS
+        ${opt_args}
+        -DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/install
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+)
+
+add_dependencies(${COMPONENT_LIB} eigen_proj)
+
+# Attach generated libraries and header files to an interface library:
+add_library(eigen_interface_lib INTERFACE)
+set_target_properties(eigen_interface_lib PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${BINARY_DIR}/install/include)
+add_dependencies(eigen_interface_lib eigen_proj)
+
+# Finally, link the interface library to the component library:
+target_link_libraries(${COMPONENT_LIB} INTERFACE eigen_interface_lib)
--- a/eigen/LICENSE
+++ b/eigen/LICENSE
@@ -1,0 +1,373 @@
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
+1.1. "Contributor"
+    means each individual or legal entity that creates, contributes to
+    the creation of, or owns Covered Software.
+
+1.2. "Contributor Version"
+    means the combination of the Contributions of others (if any) used
+    by a Contributor and that particular Contributor's Contribution.
+
+1.3. "Contribution"
+    means Covered Software of a particular Contributor.
+
+1.4. "Covered Software"
+    means Source Code Form to which the initial Contributor has attached
+    the notice in Exhibit A, the Executable Form of such Source Code
+    Form, and Modifications of such Source Code Form, in each case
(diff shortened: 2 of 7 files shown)


eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+                       PRIV_REQUIRES vfs lwip pthread newlib)
+
+# Determine compilation flags used for building Eigen
+# Flags inherited from IDF build system and other IDF components:
+set(idf_include_directories $<TARGET_PROPERTY:idf::eigen,INCLUDE_DIRECTORIES>)
+set(includes "-I$<JOIN:${idf_include_directories}, -I>")
+# IDF is built with -D_GNU_SOURCE, but we don't actually implement all _GNU_SOURCE
+# features. Defining _GNU_SOURCE causes Eigen to try to use them, which fails.
+# Use _DEFAULT_SOURCE instead, it provides enough features to build Eigen.
+set(extra_defines -D_DEFAULT_SOURCE)
```
igrr, line 18: This seems to be copy-pasted from OpenCV CMakeLists.txt file. Is this actually necessary for Eigen? If no, please remove this part, and -D_DEFAULT_SOURCE=1 on the lines below.

eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+# Final flags for C and C++:
+set(c_flags "${includes} ${extra_defines} -D_DEFAULT_SOURCE=1")
+set(cxx_flags "${includes} ${extra_defines} -D_DEFAULT_SOURCE=1")
+
+set(assert_flags "${assert_flags} -DDEBUG -D_DEBUG -ffunction-sections -fdata-sections ")
```
igrr, line 23: It's a bit odd that `-ffunction-sections -fdata-sections` are part of `assert_flags`.

Can you define something like `common_flags` and put `-ffunction-sections -fdata-sections -ggdb` there?

eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+
+if(CONFIG_COMPILER_OPTIMIZATION_DEFAULT)
+    set(opt_flags "-Og -ggdb -mlongcalls ${assert_flags}")
```
igrr, line 26: Adding `-mlongcalls` here won't work for the RISC-V targets. Please do this via
```
if(CONFIG_IDF_TARGET_ARCH_XTENSA)
    list(APPEND common_flags -mlongcalls)
endif()
```

(`common_flags` is part of the suggestion above)

eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+    set(opt_args -DCMAKE_BUILD_TYPE=Debug
+                 -DCMAKE_C_FLAGS_DEBUG=${opt_flags}
+                 -DCMAKE_CXX_FLAGS_DEBUG=${opt_flags})
+elseif(CONFIG_COMPILER_OPTIMIZATION_SIZE)
+    # Normally we should use MinSizeRel build type for this,
+    # however Eigen CMake files only handle Debug and Release
+    # (see e.g. Eigen/cmake/EigenCompilerOptions.cmake).
```
igrr, line 33: There is no EigenCompilerOptions.cmake in https://gitlab.com/libeigen/eigen/-/tree/master/cmake...

It looks like this part has been copy-pasted from OpenCV CMakeLists.txt file.

eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+    # So we redefine the flags for Release instead.
+    set(opt_flags "-Os -ggdb -mlongcalls ${assert_flags}")
+    set(opt_args -DCMAKE_BUILD_TYPE=Release
+                 -DCMAKE_C_FLAGS_RELEASE=${opt_flags}
+                 -DCMAKE_CXX_FLAGS_RELEASE=${opt_flags})
+elseif(COMPILER_OPTIMIZATION_PERF)
+    # Currently Eigen fails to compile at -O2 level on Xtensa due to a compiler issue,
```
igrr, line 40: Does it indeed fail, or is this also only copy-pasted from OpenCV?

eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+    set(opt_flags "-O0 -ggdb -mlongcalls ${assert_flags}")
+    set(opt_args -DCMAKE_BUILD_TYPE=Debug
+                 -DCMAKE_C_FLAGS_DEBUG=${opt_flags}
+                 -DCMAKE_CXX_FLAGS_DEBUG=${opt_flags})
+else()
+    message(FATAL_ERROR "Unsupported optimization level")
+endif()
+
+# Python to use for Eigen build
+idf_build_get_property(python PYTHON)
```
igrr, line 57: This seems to not be used anywhere, copy-pasted from OpenCV?

eigen/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+    USES_TERMINAL_CONFIGURE TRUE
+    USES_TERMINAL_BUILD TRUE
+    # Arguments to pass to Eigen CMake invocation:
+    CMAKE_ARGS
+        -DCMAKE_C_FLAGS=${c_flags}
+        -DCMAKE_CXX_FLAGS=${cxx_flags}
+        ${opt_args}
+        -DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/install
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+        -DCMAKE_SYSTEM_PROCESSOR=${CONFIG_IDF_TARGET_ARCH}
```
igrr, line 81: I haven't found any references in `CMAKE_SYSTEM_PROCESSOR` in libeigen. Is this line necessary?

eigen/LICENSE
```diff
@@ -0,0 +1 @@
+eigen/COPYING.README
```
igrr, line 1: I think the component manager will not be able to determine the correct license in this case. It's probably better to copy the MPL license from eigen submodule here.

(cc @kumekay in case there are better options)

eigen/README.md
```diff
@@ -0,0 +1,5 @@
+The component include Eigen library into the esp-idf as a component.
+The Eigen is a C++ template library for linear algebra: matrices, vectors, numerical solvers, and related algorithms.
+The source code for Eigne is here: https://gitlab.com/libeigen/eigen
+
+The example SVD shows how to include Eigen library into the esp-idf project.
```
igrr, line 5: Since there is no example (yet), suggest removing this line.

eigen/README.md
```diff
@@ -0,0 +1,5 @@
+The component include Eigen library into the esp-idf as a component.
```
igrr, line 1: "The component ... as a component"

igrr (changes requested):
As eigen seems to be the slightly simpler case than OpenCV, I have reviewed this first. Some of the comments will apply to OpenCV as well.

eigen/CMakeLists.txt
igrr: Actually this shouldn't be necessary at all, since we are passing CMAKE_TOOLCHAIN_FILE to the subproject, and the toolchain file already includes -mlongcalls.

What kind of errors are you getting if you don't add -mlongcalls here?

eigen/README.md
```diff
@@ -0,0 +1,3 @@
+The component port Eigen library into the esp-idf.
+The Eigen is a C++ template library for linear algebra: matrices, vectors, numerical solvers, and related algorithms.
+The source code for Eigne is here: https://gitlab.com/libeigen/eigen
```
tore-espressif, line 3:
```suggestion
# Eigen for ESP-IDF

This component ports Eigen library into the esp-idf.

**Eigen is a C++ template library for linear algebra: matrices, vectors, numerical solvers, and related algorithms.**

For more information go to http://eigen.tuxfamily.org/.

For ***pull request***, ***bug reports***, and ***feature requests***, go to https://gitlab.com/libeigen/eigen.

```

tore-espressif (approved):
@dmitry1945 thank you very much for the PR.

I left only one optional nitpick about the readme. It will be rendered in IDF component registry, e.g. here https://components.espressif.com/components/espressif/esp_encrypted_img

PR #96, opened 2022-09-23
jpeg: Move component esp_jpeg from BSP and add test.

espzav:
# Checklist

- [ ] CI passing


espzav:
@suda-morris PTAL

esp_jpeg/test/tjpgd_test.c
```diff
@@ -0,0 +1,73 @@
+#include <stdio.h>
+#include <string.h>
```
tore-espressif, line 2: Please add license + author header

tore-espressif (approved):
@espzav thank you for the follow-up. 

Are there any notable changes from the version in esp=bsp? (v1.0.2)

esp_jpeg/test/tjpgd_test.c
espzav: No change in esp_jpeg component. Only added test and moved from BSP. I increased the version for upload changed URL to component manager.

PR #97, opened 2022-09-30
USB MSC Transfer timeout increased to 5 seconds.

MaValki:
Transfer timeout increased from 1 second to 5 seconds (as recommended in USB specs).
Shorter timeout caused undesired behavior on some USB flash drives, such as re-connection issues: 
https://github.com/espressif/esp-idf/issues/9619

### DIFF
--- a/usb/usb_host_msc/src/msc_host.c
+++ b/usb/usb_host_msc/src/msc_host.c
@@ -532,7 +532,7 @@ esp_err_t msc_bulk_transfer(msc_device_t *device, uint8_t *data, size_t size, ms
     xfer->device_handle = device->handle;
     xfer->bEndpointAddress = endpoint;
     xfer->callback = transfer_callback;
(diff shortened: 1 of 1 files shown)


MaValki:
@tore-espressif  PTAL

tore-espressif:
@MaValki thanks a lot! any chance you could have a look on the memory leaks too https://github.com/espressif/esp-idf/issues/9859? If no, we can merge it as is and I'll follow up later.

MaValki:
@tore-espressif sure, I'll take a look. However, I'd still merge this as is, as it's unrelated.

tore-espressif:
> However, I'd still merge this as is, as it's unrelated.

Fair enough, I'll rebase the branch and re-run the tests (seems like unrelated failures)

tore-espressif (approved):

PR #100, opened 2022-10-26
coap: Add mbedTLS and debug config for CoAP build option (IEC-19)

ESP-YJM:
# Checklist

- [ ] CI passing

# Change description
1. Add `COAP_MBEDTLS_SUPPORT` Kconfig to make CoAP support DTLS or not. This can help save about 80 KB firmware size.
2. Depend on `COAP_MBEDTLS_DEBUG` Kconfig to make CoAP support debug or not. This can help save about 20 KB firmware size.

### DIFF
--- a/coap/Kconfig
+++ b/coap/Kconfig
@@ -1,8 +1,18 @@
 menu "CoAP Configuration"
     visible if LWIP_IPV6
 
+    config COAP_MBEDTLS_SUPPORT
+        bool "Enable mbedTLS within CoAP"
+        depends on MBEDTLS_TLS_ENABLED
+        default y
+        help
+            Enable mbedTLS functionality for CoAP.
+
+            If this option is disabled, redundent CoAP mbedTLS code is removed.
+
     choice COAP_MBEDTLS_ENCRYPTION_MODE
         prompt "CoAP Encryption method"
+        depends on COAP_MBEDTLS_SUPPORT
         default COAP_MBEDTLS_PSK
         help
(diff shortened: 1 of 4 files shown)


coap/Kconfig
```diff
@@ -1,8 +1,18 @@
...
     visible if LWIP_IPV6
 
+    config COAP_MBEDTLS_SUPPORT
+        bool "Enable mbedTLS within CoAP"
+        depends on MBEDTLS_TLS_ENABLED
+        default y
+        help
+            Enable mbedTLS functionality for CoAP.
+
+            If this option is disabled, redundent CoAP mbedTLS code is removed.
```
mahavirj, line 11: This will also make the underlying transport insecure. I am not sure how this config would be really helpful, can you please elaborate?

ESP-YJM: When only need CoAP and no DTLS, we can disable this config and it will save a lot of firmware size. I agree that CoAP transported without DTLS is not secure, but users can encrypt the CoAP payload themselves and then transport it using CoAP, which is also a way to use it. Just as RainMaker uses HTTP for local control instead of HTTPS, the data content is encrypted by negotiation. And I am going to add CoAP transmission to the protocomm layer. At present, I am comparing the memory, time and firmware size of HTTP and CoAP. Later, I will provide MR internally, and I need you to share your suggestions.

mahavirj:
Suggest creating new PR if this feature is still required.

PR #102, opened 2022-11-01
Draft: add touch_element component

L-KAYA:
# Checklist

- [ ] CI passing

# Change description


L-KAYA:
I'm planning to do the touch driver NG and the `touch_element` seems more like a library of touch driver. Therefore, I think move it into component manager makes more sense, and it will be mark as v1.0.0. But I'd like to add  `touch_element` into `idf-extra-component` first meanwhile keep it in IDF as well before the touch driver NG is merged. When the touch driver NG is ready, the updated `touch-element` in IDF will also be synchronized to `idf-extra-component` marked as version v2.0.0.

L-KAYA:
I checked with the AE colleague, he said currently no important users using `touch_element` since our touch can't pass the authentication.

L-KAYA:
> Thanks, understood! We'll see if we can also move some components out of IDF in v5.1 without making that a breaking change, through some new component manager features.

Glad to know this! Then we can remove the components with no pain!

touch_element/test_apps/main/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+idf_component_register(SRCS "test_app_main.c"
+                            "test_touch_element.c"
+                            "test_touch_button.c"
+                            "test_touch_slider.c"
+                            "test_touch_matrix.c"
+                       INCLUDE_DIRS "../../include_priv"
+                       REQUIRES "touch_element" "unity"
```
tore-espressif, line 7: Are you sure that this actually compile and links the `touch_element` from this repository?
It might link to `touch_element` from esp-idf. (But I need to double-check)

touch_element/test_apps/sdkconfig.ci.opt_o0
```diff
@@ -0,0 +1 @@
+CONFIG_COMPILER_OPTIMIZATION_NONE=y
```
tore-espressif, line 1: At the moment the `.ci` files are not taken into consideration in our CI. (point for future enhacement)

tore-espressif (commented):
I have just few questions for the beginning:

touch_element/test_apps/main/CMakeLists.txt
L-KAYA: Nice catch! I missed this point, I'll test it. BTW, do we have any way to run the CI locally? So that I can reduce the retry times on CI.

touch_element/test_apps/sdkconfig.ci.opt_o0
L-KAYA: OK, got it!

touch_element/test_apps/main/CMakeLists.txt
tore-espressif: Since you need only a generic runner, you have to do only the following:
1. cd to your touch_element test_app
2. Build for whichever target you want (eg. esp32s2)
3. Run `pytest --target=esp32s2`

(You need to have pytest installed on your local machine, 
```
pip install pytest-embedded pytest-embedded-serial-esp pytest-embedded-idf
```
)

L-KAYA:
Mark this PR as draft, we are discussing the new plan about how to deal with `touch element` in `touch driver NG`

L-KAYA:
Hi @tore-espressif , we discussed about whether to move the current `touch_element` into this repo, now we agree that better to keep the current `touch_element`(based on the legacy touch driver) in IDF but add the `touch_element` (based on the new touch driver)  into this repo in future. So I would close this PR for now, because the new `touch_element` won't be ready in a short term.

PR #103, opened 2022-11-04
Feature/upgrade components

hmalpani:
Upgraded components:

- cbor: v0.6.0
- expat: v2.5.0
- nghttp: v1.50.0


cbor/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "0.5.4"
+version: "0.6.0"
```
mahavirj, line 1: Did you try running IDF example with this component version? Any impact on the memory utilization?

mahavirj: Also I see float encoding related new source files here, should we add them to our component build list?

nghttp/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.41.0"
+version: "1.50.0"
```
mahavirj, line 1: Did you check http2_request example from IDF with this version?

cbor/idf_component.yml
hmalpani: Yes I tested the cbor example in idf with this version. It worked as expected. However the binary size increased by ~2KB

hmalpani: I have also added the float encoding related source files in CMakeList.txt

nghttp/idf_component.yml
hmalpani: http2_request example worked fine with this version of nghttp2. However we have to add -Wno-format compile option in CMakeList of nghttp as done in this PR otherwise we get build error.

Also this version have an heavy impact on binary size. Binary size increases by about 11KB

mahavirj (approved):

PR #108, opened 2022-11-17
essl: fixed build failure issue on chip which doesn't support SDMMC host

ginkgm:
# Fixed build issue

On all chips except ESP32, essl_sdio.c should be able to be built. Theoratically ESSL should be able to make use of the SPI interface to  communicate with ESP32 SDIO slave, even if the chip have no SDIO Slave interface.

- On ESP32-S3, there is no SDIO slave, but it can use SDMMC host to communicate with slave;
- On all chips, there is no SDIO slave, but ideally it can make use of SPI interface, though currently there is a bug with its timing.

Fix this issue.

# Remove dependency to soc/host_reg.h

The host should work without depending on the private information of ESP32. Only the public part is necessary. Remove the dependency to (esp32)/soc/host_reg.h so that ESSL can be built, and be more portable.

# Upgrade to v1.0.1

I think these can be treated as bugs.


ginkgm:
@Icarus113 PTAL

Icarus113:
LGTM @ginkgm

mahavirj:
@ginkgm 

- Please update the PR title, it can talk about the the problem that this PR addresses
- Please also update the PR description, e.g., on which chip the build failure was observed etc.

ginkgm:
@mahavirj updated, PTAL again

PR #114, opened 2022-12-14
Example for Eigen idf component

dmitry1945:
# Checklist

- [x ] Component contains License
- [ x] Component contains README.md
- [ x] Component contains idf_component.yml file with `url` field defined
- [ x] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [x ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] CI passing

# Change description
This PR add example to the Eigen component directory that shows 
how to use the eigen library and IDF Eigen component.

Regards,
Dmitry

### DIFF
--- a/eigen/examples/svd/README.md
+++ b/eigen/examples/svd/README.md
@@ -1,0 +1,14 @@
+# Eigen example for IDF Eigen component
+
+This example shows how to use Eigen library in ESP-IDF projects.
+
+The example does matrix multiplication and single value decomposition (SVD)
+transformation.
+
+This example does not require any special hardware, and can be run on any common development board.
+
+To run the example on target please run:
+
+```
+idf.py -p PORT flash monitor
+```
--- a/eigen/examples/svd/main/idf_component.yml
+++ b/eigen/examples/svd/main/idf_component.yml
@@ -1,0 +1,10 @@
+## IDF Component Manager Manifest File
+dependencies:
+  espressif/eigen: 
+    version: "^3.4.0"
+    # This line define the local path of the eigen component because this 
+    # example is part of the eigen component. This line is optional.
+    override_path: "../../.."
+  ## Required IDF version
+  idf:
+    version: ">=4.3.0"
--- a/eigen/examples/svd/main/main.cpp
+++ b/eigen/examples/svd/main/main.cpp
@@ -1,0 +1,46 @@
+/*
+ * SPDX-FileCopyrightText: 2010-2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+*/
+
+#include <iostream>
+
+#include <eigen3/Eigen/Eigen>
+
+extern "C" void app_main(void);
+
+static void multiply2Matrices()
+{
+    Eigen::MatrixXf M(2, 2);
+    Eigen::MatrixXf V(2, 2);
+    for (int i = 0; i <= 1; i++) {
+        for (int j = 0; j <= 1; j++) {
+            M(i, j) = 1;
+            V(i, j) = 2;
+        }
+    }
+    Eigen::MatrixXf Result = M * V;
+    std::cout << "MatrixXf Result = " << std::endl << Result << std::endl;
+}
+
+static void runSVD()
+{
+    Eigen::MatrixXf C;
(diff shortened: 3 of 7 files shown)


eigen/examples/svd/main/main.cpp
```diff
@@ -0,0 +1,57 @@
+/* Import Prebuilt Library Example
```
igrr, line 1: This line probably needs to be updated

eigen/examples/svd/main/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <esp_log.h>
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
```
igrr, line 17: Looks like this is unused? Please keep only the includes which the program actually needs.

eigen/examples/svd/main/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <iostream>
```
igrr, line 18: Since you also `printf` below, do you need iostream? Note that usage of iostream increases the binary size quite significantly, not sure if we should use that in the examples.

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/cplusplus.html#iostream

eigen/examples/svd/main/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <eigen3/Eigen/Eigen>
+
+extern "C" void app_main(void);
+
+
+void multiply2Matrices()
```
igrr, line 25:
```suggestion
static void multiply2Matrices()
```

eigen/examples/svd/main/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+        for (int j = 0; j <= 1; j++) {
+            M(i, j) = 1;
+            V(i, j) = 2;
+        }
+    }
+    Eigen::MatrixXf Result = M * V;
+    std::cout << "MatrixXf Result = " << std::endl << Result << std::endl;
+}
+
+void runSVD()
```
igrr, line 39:
```suggestion
static void runSVD()
```

eigen/examples/svd/main/idf_component.yml
```diff
@@ -0,0 +1,10 @@
+## IDF Component Manager Manifest File
+dependencies:
+  espressif/eigen: 
+    version: "^3.4.0"
+    # This line define the local path of the eigen component because this 
+    # example is part of the eigen component. This line is optional.
+    override_path: "../../.."
+  ## Required IDF version
+  idf:
+    version: ">=4.1.0"
```
igrr, line 10: Eigen itself depends on IDF v4.3 or later here:
https://github.com/espressif/idf-extra-components/blob/a4049b54515f14e5066611f8087f23bd9a47d11b/eigen/idf_component.yml#L6
so I suggest making this consistent
```suggestion
    version: ">=4.3.0"
```

eigen/examples/svd/README.md
```diff
@@ -0,0 +1,14 @@
+# Eigen example for IDF Eigen component
+
+This example shows how to use Eigen library in ESP-IDF projects.
+
+The example make matrix multiplication and single value decomposition (SVD)
```
igrr, line 5:
```suggestion
The example does matrix multiplication and single value decomposition (SVD) 
```

igrr (changes requested):
@dmitry1945 Looks good to me aside from a few minor comments.

Please rebase this branch on top of the latest `master` as well.

eigen/examples/svd/main/main.cpp
dmitry1945: iostream used for print formated matrix values. I can't remove it.

igrr: Alright, that makes sense. In that case, could you modify `printf("Eigen example.\n");` to also use std::cout?

igrr (approved):

PR #119, opened 2022-12-22
led_strip: support dma feature and various clock source

suda-morris:
This PR enables the user to select more clock sources for the led strip (RMT backend). 

DMA support can also be enabled optionally.

Restructured the component source file and updated the README file.

### DIFF
--- a/led_strip/include/led_strip_rmt.h
+++ b/led_strip/include/led_strip_rmt.h
@@ -1,0 +1,45 @@
+/*
+ * SPDX-FileCopyrightText: 2022 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <stdint.h>
+#include "esp_err.h"
+#include "driver/rmt_types.h"
+#include "led_strip_types.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+/**
+ * @brief LED Strip RMT specific configuration
+ */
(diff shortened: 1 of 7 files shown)


suda-morris:
Hi @tore-espressif  would you please help review the PR?

led_strip/include/led_strip_rmt.h
```diff
@@ -0,0 +1,45 @@
...
+
+
+/**
+ * @brief LED Strip RMT specific configuration
+ */
+typedef struct {
+    rmt_clock_source_t clk_src; /*!< RMT clock source */
+    uint32_t resolution_hz;     /*!< RMT tick resolution, if set to zero, a default resolution (10MHz) will be applied */
+    struct {
+        uint32_t with_dma: 1;   /*!< Use DMA to transmit data */
```
tore-espressif, line 25: Would it be worth adding more RMT flags? Such as `invert_out` or `io_od_mode`?

(https://github.com/espressif/esp-idf/blob/v5.0/components/driver/include/driver/rmt_tx.h#L44)

tore-espressif (approved):
@suda-morris Just one question. Otherwise LGTM!

led_strip/include/led_strip_rmt.h
suda-morris: Thanks for the reminder, I added the `nver_out` flag now. But the "io_od_mode" is not necessary for the led_strip application AFAIK.

tore-espressif (approved):

PR #121, opened 2023-01-04
Do not pass the private key for encryption, the public one will be en…

RaD:
…ough and safe

# Checklist

- [ ] CI passing

# Change description
It is unsafe to require the private key on encryption stage, just use public key.

### DIFF
--- a/esp_encrypted_img/tools/esp_enc_img_gen.py
+++ b/esp_encrypted_img/tools/esp_enc_img_gen.py
@@ -46,9 +46,7 @@ def encrypt(input_file: str, rsa_key_file_name: str, output_file: str) -> None:
         data = image.read()
 
     with open(rsa_key_file_name, 'rb') as key_file:
-        private_key = serialization.load_pem_private_key(key_file.read(), password=None)
-
-    public_key = private_key.public_key()
+        public_key = serialization.load_pem_public_key(key_file.read())
 
     gcm_key = generate_key_GCM(GCM_KEY_SIZE)
     iv = generate_IV_GCM()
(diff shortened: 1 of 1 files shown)


esp_encrypted_img/tools/esp_enc_img_gen.py
```diff
@@ -46,9 +46,7 @@ def encrypt(input_file: str, rsa_key_file_name: str, output_file: str) -> None:
         data = image.read()
 
     with open(rsa_key_file_name, 'rb') as key_file:
-        private_key = serialization.load_pem_private_key(key_file.read(), password=None)
-
-    public_key = private_key.public_key()
+        public_key = serialization.load_pem_public_key(key_file.read())
```
mahavirj, line 49: This will break the example code from https://github.com/espressif/esp-idf/blob/4c98bee8a4417d92a387ef2416be14ff1fdfa92a/examples/system/ota/pre_encrypted_ota/main/CMakeLists.txt#L7

I think we can dynamically detect the key type and then use appropriate method to load it. Something along the lines of:

```
if b"-BEGIN RSA PRIVATE KEY" in key_data or b"-BEGIN PRIVATE KEY" in key_data:
   load_as_priv_key
elif b"-BEGIN PUBLIC KEY" in key_data:
   load_as_pub_key
```

Could you please check this once?

mahavirj (changes requested):

esp_encrypted_img/tools/esp_enc_img_gen.py
mahavirj: Added the change in https://github.com/espressif/idf-extra-components/pull/124

mahavirj:
Fixed in https://github.com/espressif/idf-extra-components/pull/124. Closing this PR.

PR #122, opened 2023-01-04
Change minimum IDF version for `cbor` component to IDF v5.0

dhavalgujar:
# Checklist

- [] _Optional:_ Component contains unit tests
- [x] CI passing

# Change description
- IDF versions below v5.0 already include the `cbor` component in source format.
- Due to this, there is a conflict if the `cbor` component is used on lower versions of IDF and the only way to fix it is by adding a rule to the `idf_component.yml`
- This commit sets the minimum version to IDF v5.0 to fix this behaviour.

### DIFF
--- a/cbor/idf_component.yml
+++ b/cbor/idf_component.yml
@@ -1,5 +1,5 @@
-version: "0.6.0"
+version: "0.6.1"
 description: "CBOR: Concise Binary Object Representation Library"
 url: https://github.com/espressif/idf-extra-components/tree/master/cbor
 dependencies:
-  idf: ">=4.3"
+  idf: ">=5.0"


dhavalgujar:
> I think we still want to have the option of installing the latest version of cbor component in IDF v4.x.

Ah, I see. I thought the version of `cbor` in IDF v4.4 and v0.6.0 from the IDF Component Registry were the same.

> Could you explain why this is problematic?

It isn't problematic, it is just a different pattern than what has been followed by other components which were also removed from IDF in v5.0 and added to the IDF Component Registry.
For example: The `led_strip` component exists as [a part of](https://github.com/espressif/esp-idf/tree/release/v4.4/examples/common_components/led_strip) IDF v4.4 but the latest one from the [registry](https://components.espressif.com/components/espressif/led_strip) has IDF v5.0 as the minimum. 
Again, there might be other reasons why that component lists IDF v5.0 as the minimum which I am unaware of.

> I am not sure whether your proposed change will be effective.

Right, this might not work unless we can change the minimum IDF compatibility of an older version which I'm assuming isn't possible and probably a bad idea?

mahavirj:
Lets keep current IDF 4.3 version approach for `cbor` component. Provides an advantage of easy upgrade across previous releases..

PR #123, opened 2023-01-09
USB: esp_tinyusb - fix memory leak in case of error, refactoring

esp-saurabhbansal:
# Checklist

- [ ] CI passing

# Change description
1. Fix memory leak while returning in case of error. While returning in case of error, these should be called for freeing up the resources:
    a. tinyusb_cdc_deinit()
    b. free(acm->rx_tfbuf);
2. Remove unused 'initialized'
3. Remove hard coding of "/dev/tusb_cdc"

### DIFF
--- a/usb/esp_tinyusb/tusb_cdc_acm.c
+++ b/usb/esp_tinyusb/tusb_cdc_acm.c
@@ -21,7 +21,6 @@
 
 
 typedef struct {
-    bool initialized;
     size_t rx_unread_buf_sz;
     RingbufHandle_t rx_unread_buf;
     SemaphoreHandle_t ringbuf_read_mux;
@@ -352,6 +351,7 @@ static void free_obj(tinyusb_cdcacm_itf_t itf)
 
 esp_err_t tusb_cdc_acm_init(const tinyusb_config_cdcacm_t *cfg)
 {
+    esp_err_t ret = ESP_OK;
     int itf = (int)cfg->cdc_port;
     /* Creating a CDC object */
     const tinyusb_config_cdc_t cdc_cfg = {
@@ -359,10 +359,10 @@ esp_err_t tusb_cdc_acm_init(const tinyusb_config_cdcacm_t *cfg)
         .cdc_class = TUSB_CLASS_CDC,
         .cdc_subclass.comm_subclass = CDC_COMM_SUBCLASS_ABSTRACT_CONTROL_MODEL
     };
-    ESP_RETURN_ON_ERROR(tinyusb_cdc_init(itf, &cdc_cfg), TAG, "tinyusb_cdc_init failed");
-    ESP_RETURN_ON_ERROR(alloc_obj(itf), TAG, "alloc_obj failed");
 
-    esp_tusb_cdcacm_t *acm = get_acm(itf);
+    ESP_RETURN_ON_ERROR(tinyusb_cdc_init(itf, &cdc_cfg), TAG, "tinyusb_cdc_init failed");
+    ESP_GOTO_ON_FALSE(!alloc_obj(itf), ESP_FAIL, out4, TAG, "alloc_obj failed");
+
     /* Callbacks setting up*/
     if (cfg->callback_rx) {
         tinyusb_cdcacm_register_callback(itf, CDC_EVENT_RX, cfg->callback_rx);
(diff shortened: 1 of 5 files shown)


esp-saurabhbansal:
@tore-espressif @roma-jam  @igrr 
PTAL

usb/esp_tinyusb/tusb_cdc_acm.c
```diff
@@ -382,26 +382,34 @@ esp_err_t tusb_cdc_acm_init(const tinyusb_config_cdcacm_t *cfg)
     acm->ringbuf_read_mux = xSemaphoreCreateMutex();
     if (acm->ringbuf_read_mux == NULL) {
         ESP_LOGE(TAG, "Creation of a ringbuf mutex failed");
-        free_obj(itf);
-        return ESP_ERR_NO_MEM;
+        ret = ESP_ERR_NO_MEM;
+        goto clean;
     }
```
tore-espressif, line 387: Nitpick: Could we use the `ESP_GOTO_ON_FALSE` macros here? to have the same style as in IDF code?

usb/esp_tinyusb/tusb_cdc_acm.c
```diff
@@ -382,26 +382,34 @@ esp_err_t tusb_cdc_acm_init(const tinyusb_config_cdcacm_t *cfg)
...
         return ESP_OK;
     }
+clean:
+    if (acm->rx_tfbuf) {
+        free(acm->rx_tfbuf);
+        acm->rx_tfbuf = NULL;
+    }
+    free_obj(itf);
+    tinyusb_cdc_deinit(itf);
+    return ret;
```
tore-espressif, line 412: Should we unregister the callbacks too?

usb/esp_tinyusb/tusb_cdc_acm.c
```diff
@@ -382,26 +382,34 @@ esp_err_t tusb_cdc_acm_init(const tinyusb_config_cdcacm_t *cfg)
...
         ESP_LOGD(TAG, "Comm Initialized buff:%d bytes", cfg->rx_unread_buf_sz);
         return ESP_OK;
     }
+clean:
+    if (acm->rx_tfbuf) {
+        free(acm->rx_tfbuf);
+        acm->rx_tfbuf = NULL;
+    }
+    free_obj(itf);
+    tinyusb_cdc_deinit(itf);
```
tore-espressif, line 411: Could you please review the `free_obj` and `deinit` functions? I find it quite confusing. Could we merge the allocation and initialization (and analogously freeing and deinitialization) to just one function?

tore-espressif (commented):
Could you also please bump version of `esp_tinyusb/idf_component.yml` to `1.0.3`?

This way, the component will be uploaded to IDF component registry, after merging

usb/esp_tinyusb/tusb_cdc_acm.c
esp-saurabhbansal: done

usb/esp_tinyusb/tusb_cdc_acm.c
esp-saurabhbansal: No need.
On registering callbacks, these are assigned in CDC_ACM object esp_tusb_cdcacm_t. 
free_obj(itf); actually frees up the memory allocated by esp_tusb_cdcacm_t struct. So, unregistering of callbacks are not needed because memory is getting freed up.

usb/esp_tinyusb/tusb_cdc_acm.c
esp-saurabhbansal: esp_tusb_cdcacm_t is a subclass specific object (i.e. ACM) which is available only in tusb_sdc_acm.c. It is hidden from generic cdc class source file cdc.c.
alloc_obj()/free_obj() is for allocation/freeing the subclass specific object esp_tusb_cdcacm_t (Here, only ACM type) and thus used only in tusb_sdc_acm.c. By this design and segregation, we should not move the subclass specific data to generic cdc class file.

esp-saurabhbansal: @tore-espressif Please check and approve if no more comments.

tore-espressif: OK, agreed

tore-espressif (approved):

usb/esp_tinyusb/tusb_cdc_acm.c
```diff
@@ -380,28 +380,25 @@ esp_err_t tusb_cdc_acm_init(const tinyusb_config_cdcacm_t *cfg)
...
-        return ESP_ERR_NO_MEM;
-    } else {
-        ESP_LOGD(TAG, "Comm Initialized buff:%d bytes", cfg->rx_unread_buf_sz);
-        return ESP_OK;
+    ESP_GOTO_ON_FALSE(acm->rx_unread_buf, ESP_ERR_NO_MEM, clean, TAG, "Creation buffer error");
+
+    ESP_LOGD(TAG, "Comm Initialized buff:%d bytes", cfg->rx_unread_buf_sz);
+    return ESP_OK;
+clean:
+    if (acm->rx_tfbuf) {
```
roma-jam, line 395: Hi Saurabh, 
Great job!

One thing that I want to clarify for me is the next. During the tusb_cdc_acm_init function we:
- init tinyusb class device (with tinyusb_cdc_init())
- allocate cdc acm object
- create mutex (ringbuf_read_mux)
- allocate rx_tfbuf 
- allocate ring buffer (rx_unread_buf). 

During my attempts to understand the logic I see, that two main things should be achieved at the end:
1. We can use label goto when we need to separate some different steps from each other.
2. We should free all object, that was allocated at the step/steps before. 

In this case I am concern about the logic, when we need to use only one label (As much as much often at this time we can avoid using that at all). 
The second one tells us that we need to get freed all object had been allocated. That means that we need to do it in a backwards order: 
- free ring buffer
- free rx buffer
- delete mutex
- free cdc acm object
- deinit class usb device

From that point of view, it looks that with the goto labels we can do the following: 
`

    init_tinyusb_class_device();
        if not goto init_fail;
    allocate_cdc_acm_object();
        if not goto allocate_object_fail;
    create_mutex();
        if not goto create_mutex_failure;
    allocate_rx_tfbuf();
        if not goto alloc_buf_err;
    allocate_ring_buffer();
         if not goto ring_buffer_fail;
    return OK;

    ring_buffer_fail:
         free_rx_buffer();
    alloc_buf_error:
         delete_mutex();
    create_mutex_failure:
        free_cdc_acm_object();
    allocate_object_fail:
        deinit_class_usb_device();
    init_fail: 
        // nothing else to free
        return ERR;
`

Honestly, I am not a fan of the goto labels, but in such case it seems to be at least justified. 
Maybe I not understand the logic or I do not have the whole image...But it looks that in the suggested logic we: 
- use goto labels for their purpose
- freed all the object had been allocated... 

That is the only thing I would like to highlight.

tore-espressif: C standard says that passing NULL to `free` does nothing. So in order to have the clean-up procedure as trivial as possible, we usually use only one label where we free everything (even if the allocation did not occur). Does that make sense to you @roma-jam ?

Very good answer for this question, for you reference: https://stackoverflow.com/questions/6084218/is-it-good-practice-to-free-a-null-pointer-in-c

roma-jam: @tore-espressif If the clean-up procedure is only for freeing allocate memory - yes, that does make perfect sense.

tore-espressif: Are you OK with merging this?

roma-jam: @tore-espressif I am not sure. 
Probably we need to delete mutex acm->ringbuf_read_mux in case acm->rx_tfbuf allocation error. 
I do not see it in code.

esp-saurabhbansal: @roma-jam @tore-espressif Please check again the new code with additional goto labels.

roma-jam (approved):
LGTM

PR #126, opened 2023-01-13
led_strip: support for 4 components RGBW leds (SK6812)

Pac72:
# Checklist

- [ ] CI passing

# Change description
Implementing support for RGBW leds, specifically for SK6812.
These devices have 4 leds, the standard three RGB ones and a fourth white led.

### DIFF
--- a/led_strip/include/led_strip_types.h
+++ b/led_strip/include/led_strip_types.h
@@ -11,7 +11,15 @@
 extern "C" {
 #endif
 
-enum led_type_t { LED_TYPE_WS2812 = 0, LED_TYPE_SK6812 };
+typedef enum {
+    LED_PIXEL_FORMAT_GRB,
+    LED_PIXEL_FORMAT_GRBW
+} led_pixel_format_t;
+
+typedef enum {
+    LED_MODEL_WS2812,
+    LED_MODEL_SK6812
+} led_model_t;
 
 /**
  * @brief LED strip handle
@@ -24,7 +32,8 @@ typedef struct led_strip_t *led_strip_handle_t;
 typedef struct {
     uint32_t strip_gpio_num; /*!< GPIO number that used by LED strip */
     uint32_t max_leds;       /*!< Maximum LEDs in a single strip */
-    enum led_type_t led_type;
+    led_pixel_format_t led_pixel_format;
+    led_model_t led_model;
     struct {
         uint32_t invert_out: 1; /*!< Invert output signal */
     } flags;
--- a/led_strip/include/led_strip_types.h
+++ b/led_strip/include/led_strip_types.h
@@ -11,6 +11,8 @@
 extern "C" {
 #endif
 
+enum led_type_t { LED_TYPE_WS2812 = 0, LED_TYPE_SK6812 };
+
 /**
  * @brief LED strip handle
  */
@@ -22,6 +24,7 @@ typedef struct led_strip_t *led_strip_handle_t;
 typedef struct {
     uint32_t strip_gpio_num; /*!< GPIO number that used by LED strip */
     uint32_t max_leds;       /*!< Maximum LEDs in a single strip */
+    enum led_type_t led_type;
     struct {
         uint32_t invert_out: 1; /*!< Invert output signal */
     } flags;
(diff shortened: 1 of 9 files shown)


led_strip/src/led_strip_rmt_dev.c
```diff
@@ -22,18 +22,36 @@ typedef struct {
...
+        rmt_strip->pixel_buf[start + 3] = 0;
+    }
+    return ESP_OK;
+}
+
+static esp_err_t led_strip_rmt_set_pixel_rgbw(led_strip_t *strip, uint32_t index, uint32_t red, uint32_t green, uint32_t blue, uint32_t white)
+{
+    led_strip_rmt_obj *rmt_strip = __containerof(strip, led_strip_rmt_obj, base);
+    ESP_RETURN_ON_FALSE(index < rmt_strip->strip_len, ESP_ERR_INVALID_ARG, TAG, "index out of maximum number of LEDs");
+    ESP_RETURN_ON_FALSE(rmt_strip->bytes_per_pixel == 4, ESP_ERR_INVALID_ARG, TAG, "wrong LED type, expected 4 bytes per pixel");
```
suda-morris, line 48:
```suggestion
    ESP_RETURN_ON_FALSE(rmt_strip->bytes_per_pixel == 4, ESP_ERR_INVALID_ARG, TAG, "wrong LED pixel format, expected 4 bytes per pixel");
```

suda-morris (changes requested):
@Pac72 Thank you very much for adding support for more LED strip models!

led_strip/include/led_strip_types.h
```diff
@@ -11,6 +11,8 @@
 extern "C" {
 #endif
 
+enum led_type_t { LED_TYPE_WS2812 = 0, LED_TYPE_SK6812 };
```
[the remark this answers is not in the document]
Pac72: @suda-morris, it sounds pretty sensible.
Indeed, the bunch of datasheets for these things is a real mes<DEL><DEL><DEL> complex matter.
I analyzed a few dozens here: https://docs.google.com/spreadsheets/d/1-AGjwhz6Os8QdFERwoYQDjFJy2YC0yVbzDxkl0ICr7E/edit?usp=sharing
There's not an easy way out, better stick on the simple concept, like the pixel format.
If you don't mind, I'd rather prefer using LED_PIXEL_FORMAT_GRB, LED_PIXEL_FORMAT_GRBW, since I found some datasheets showing an R, G, B byte sequence.
About the bit timing management, I'll leave the two configurations, but after the datasheets analysis I think it would be better to provide a single default sound timing (or two, a fast one and a more conservative one) and give the user the opportunity to customize it in case of problems.

led_strip/include/led_strip_types.h
```diff
@@ -22,6 +32,8 @@ typedef struct led_strip_t *led_strip_handle_t;
 typedef struct {
     uint32_t strip_gpio_num; /*!< GPIO number that used by LED strip */
     uint32_t max_leds;       /*!< Maximum LEDs in a single strip */
+    led_pixel_format_t led_pixel_format;
```
suda-morris, line 35: would you please add some docs beside? We will add API documentation for this library, and the Doxygen tool will warn here (uncommented struct field).

led_strip/include/led_strip_types.h
```diff
@@ -11,6 +11,16 @@
 extern "C" {
 #endif
 
+typedef enum {
```
suda-morris, line 14: likewise, missing comments on the struct and its fields

led_strip/src/led_strip_rmt_encoder.h
```diff
@@ -17,6 +18,7 @@ extern "C" {
  */
 typedef struct {
     uint32_t resolution; /*!< Encoder resolution, in Hz */
+    led_model_t led_model;
```
suda-morris, line 21: missing comments

suda-morris (approved):
Thank you @Pac72 for the updating! Looks nice.

led_strip/include/led_strip_types.h
Pac72: It seems I arrived late to the party, master already contains the documentation...

PR #130, opened 2023-01-24
cbor: Add example for cbor library

hmalpani:
* Add example in `CBOR` component.
* Add workflow to run `CBOR` example in CI.
* Remove test_app from `CBOR` example.


cbor/examples/cbor/README.md
```diff
@@ -0,0 +1,59 @@
+| Supported Targets | ESP32 | ESP32-C2 | ESP32-C3 | ESP32-C6 | ESP32-H2 | ESP32-S2 | ESP32-S3 |
```
mahavirj, line 1: This is pretty high level example and it should be supported for all targets. Shall we remove this line?

cbor/examples/cbor/main/idf_component.yml
```diff
@@ -0,0 +1,3 @@
+## IDF Component Manager Manifest File
```
mahavirj, line 1: Probably we can add following lines for this example here:

```
version: "1.0.0"
description: CBOR Example
```

cbor/examples/cbor/main/idf_component.yml
```diff
@@ -0,0 +1,3 @@
...
+dependencies:
+  espressif/cbor: ">=0.5.4,<1.0.0"
```
mahavirj, line 3: We should consider adding `override_path` here, please refer to https://github.com/espressif/idf-component-manager/wiki#adding-dependency-on-the-component-for-examples

cbor/examples/cbor/README.md
hmalpani: Removed

cbor/examples/cbor/main/idf_component.yml
hmalpani: Added `override_path` in idf_component.yml file.

cbor/examples/cbor/main/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+## IDF Component Manager Manifest File
+version: "1.0.0"
+description: CBOR Example
+dependencies:
+  espressif/cbor:
+    override_path: '../../../'
```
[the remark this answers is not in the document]
hmalpani: `override_path` states that the example should use the local component instead of getting it from the component manager. I don't think not giving the version would make any difference.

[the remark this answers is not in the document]
mahavirj: > override_path states that the example should use the local component instead of getting it from the component manager

Only when you cloned the example directly from the component repository. When you set up the example using component manager cli then it shall remove the `override_path` and hence version becomes important.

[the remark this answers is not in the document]
hmalpani: Oh okay. Got it.

hmalpani:
@mahavirj I have also added workflow for running cbor example. It passes in the CI. PTAL. If it looks good I think we can remove the test_app from cbor component.

mahavirj:
> @mahavirj I have also added workflow for running cbor example. It passes in the CI. PTAL. If it looks good I think we can remove the test_app from cbor component.

Looks good to me. Lets remove the test_app.

PR #131, opened 2023-01-29
tjpgd: remove ROM support for esp32h2

suda-morris:
ESP32H2's ROM doesn't contain the tjpgd library.


espzav:
@suda-morris Thank you for add esp32h2 support!

PR #133, opened 2023-02-03
usb: add ECM device class

Jacques-zhao:
Added support for USB NCM class.
This MR relies on a tinyusb [MR](https://github.com/espressif/tinyusb/pull/9). The USB NCM example will be placed in the examples/peripherals/usb/device/tusb_ncm directory under esp-idf repository.

### DIFF
--- a/usb/esp_tinyusb/Kconfig
+++ b/usb/esp_tinyusb/Kconfig
@@ -112,6 +112,14 @@ menu "TinyUSB Stack"
                 MSC FIFO size, in bytes.
     endmenu # "Massive Storage Class"
 
+    menu "USB Network Class (NCM)"
+        config TINYUSB_NCM_ENABLE
+            bool "Enable TinyUSB NCM driver"
+            default n
+            help
+                Enable USB NET TinyUSB driver.
+    endmenu # "usb network"
+
     menu "Communication Device Class (CDC)"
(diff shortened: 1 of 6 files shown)


roma-jam:
@Jacques-zhao Hi, well done and you have done a great job! 

The one thing I would like to ask. 
If there is an additional string descriptor logic has been added, probably it is better to do that with a specific function. 
For example, "tusb_set_extra_str_descritpor" or "tusb_set_str_descriptor", which should have been added as an independent function from "tusb_set_descriptor". 

That can help us to achieve following things:
- Not to change tusb_set_decriptor parameters for those application, which has already used the previous list of args (dev_desc, str_desc, cfg_desc)
- Distinguish the logic of adding additional extra string descriptors to the device, which can or cannot be used depending on the build config (I mean CONFIG_TINYUSB_NCM_ENABLE).

What do you think? 

Anyway, thanks for the changes. 
Probably, @tore-espressif can suggest more things here. 

Always ready to a discuss, so feel free to do that. 
RJ

tore-espressif:
@roma-jam @Jacques-zhao I started refactoring the string descriptors handling too :) 

I'll review this (or push my code for comparison) soon

Jacques-zhao:
I will resubmit a new version based on Tomas PR(https://github.com/espressif/idf-extra-components/pull/134)

PR #140, opened 2023-02-10
json_parser: add idf_version rules for jsmn

tswen:
# Checklist

- [ ] CI passing

# Change description
Fix a compilation error of `json_parser` component on idf v4.x.

Because the jsmn component was part of idf v4.4 and earlier.

The compile errors were:
```c
  ERROR: Cannot process component requirements.  Multiple candidates to
  satisfy project requirements:

    requirement: "jsmn" candidates: "espressif__jsmn, jsmn"


```

### DIFF
--- a/json_parser/idf_component.yml
+++ b/json_parser/idf_component.yml
@@ -1,5 +1,8 @@
-version: "1.0.2"
+version: "1.0.3"
 description: This is a simple, light weight JSON parser built on top of jsmn
 url: https://github.com/espressif/json_parser
 dependencies:
-  jsmn: "~1.1"
+  jsmn:
+    version: "~1.1"
+    rules:
+      - if: "idf_version >=5.0"


igrr:
@shahpiyushv PTAL

shahpiyushv:
@tswen , can you add some explanation as to why this is being done. I myself know that jsmn was part of idf earlier, but is not there anymore and so pulling this component must be causing component duplication. But, it would be better to have this info in commit message too.

Meanwhile, have you tested if this works well with the jsmn from idf v4.x?

json_parser/idf_component.yml
```diff
@@ -1,5 +1,8 @@
-version: "1.0.2"
+version: "1.0.3"
```
shahpiyushv, line 1: @igrr , do we need to change component version just for updating dependency?

igrr:
> Meanwhile, have you tested if this works well with the jsmn from idf v4.x?

@shahpiyushv could you suggest a simple smoke test case for json_parser component? Then we could add it to the repo and run the test automatically, ensuring that the component works with all the supported versions of IDF.

json_parser/idf_component.yml
igrr: Yes, we do. A component will get published to the registry only when its version is changed. It's not possible to "overwrite" an already published version of the component.

igrr:
@tswen please click through the CLA as well. (This is not strictly necessary since you are an employee, but it's just easier to get all the status checks green this way.)

shahpiyushv:
> > Meanwhile, have you tested if this works well with the jsmn from idf v4.x?
> 
> @shahpiyushv could you suggest a simple smoke test case for json_parser component? Then we could add it to the repo and run the test automatically, ensuring that the component works with all the supported versions of IDF.

There is actually a generic test code in the original repo if someone wants to adapt it for esp-idf: https://github.com/espressif/json_parser/blob/master/tests/main.c. If there are some pointers to help add the tests here, I can do it myself too.

igrr:
@shahpiyushv I guess it could look [like this](https://github.com/espressif/idf-extra-components/blob/66359e25424199907cf6fa0ee327c9019cee8f34/expat/test/test_expat.c#L75) — very similar to the version in your repo, just wrapped into a TEST_CASE macro and with a bunch of TEST_ASSERTs instead of printtfs, so that the results can be verified automatically.

tswen:
@shahpiyushv I tested the change on idf 4.4 and it works well.

igrr:
@tswen This change looks good to me! As @shahpiyushv said, please update the Pull Request description to explain the reason behind this change. Then it should be ready it merge.

tswen:
@igrr @shahpiyushv Thanks to your review suggestions, I've updated the change description.

igrr (approved):

igrr:
@tswen Have merged, thank you for the contribution!

PR #147, opened 2023-02-20
Bugfix/json parser and others

igrr:
## Change description

* Include json_parser and json_generator components into the test_app build.
* Add a test for json_parser, based on https://github.com/espressif/json_parser/blob/master/tests/main.c
* Specify override_path for jsmn in json_parser. Since jsmn is in the same repository, we want to use the copy of the component from the repository when building the test app.
* Likewise, add override_path for nghttp in sh2lib.
* Revert the "rules" entry for json_parser added in #140, since it is not currently supported by idf-component-manager when uploading components to the registry. This requires a bugfix release on the idf-component-manager side, which might take several days. Reverting this change helps unblock CI.

There are no functional changes for components themselves, so I'm not increasing the versions here.

### DIFF
--- a/json_parser/idf_component.yml
+++ b/json_parser/idf_component.yml
@@ -1,9 +1,14 @@
-version: "1.0.3"
+version: "1.0.2"
 description: This is a simple, light weight JSON parser built on top of jsmn
 url: https://github.com/espressif/json_parser
 dependencies:
   jsmn:
     version: "~1.1"
-    rules:
-      - if: "idf_version >=5.0"
+    # The "rules" entry is temporarily disabled, as idf-component-manager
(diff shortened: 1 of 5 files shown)


igrr:
@tswen FYI, i had to revert your PR #140 since it happens to break idf-component-manager (https://github.com/espressif/idf-extra-components/actions/runs/4224071639/jobs/7334567785). I will add this rule back once the bugfix release of idf-component-manager is done.

igrr:
@shahpiyushv PTAL at the test case for json_parser

tore-espressif (approved):
LGTM, I will follow-up with a GH action that checks manifest files before merging to master

PR #151, opened 2023-02-21
usb: Implement new usbd MSC Class in esp_tinyusb

esp-saurabhbansal:
1. Add usbd MSC  Class for accessing storage like SPI Flash or sd-card.
2. Modify TCs for this change.


usb/esp_tinyusb/Kconfig
```diff
@@ -160,4 +160,66 @@ menu "TinyUSB Stack"
             help
                 Setting value greater than 0 will enable TinyUSB HID feature.
     endmenu # "HID Device Class (HID)"
+
+    menu "Mass Storage Device Class (MSC)"
+        choice STORAGE_MEDIA
```
igrr, line 165: Can this be decided at link time and run time, instead of configuration time? Some of the projects (Arduino, Micropython) use ESP-IDF and other components in a pre-compiled form, therefore fixing this choice here will mean that they would need to build two versions of a precompiled library.

Could we introduce two functions,

```
esp_err_t tinyusb_msc_storage_init_sdcard(const tinyusb_msc_storage_config_t *config, more args);
esp_err_t tinyusb_msc_storage_init_spiflash(const tinyusb_msc_storage_config_t *config more args);
```

instead?

usb/esp_tinyusb/Kconfig
```diff
@@ -160,4 +160,66 @@ menu "TinyUSB Stack"
...
+            config STORAGE_MEDIA_SPIFLASH
+                bool "SPI FLASH"
+
+            config STORAGE_MEDIA_SDMMCCARD
+                bool "SDMMC CARD"
+                depends on IDF_TARGET_ESP32S3
+        endchoice
+
+        if STORAGE_MEDIA_SDMMCCARD
+            choice SDMMC_BUS_WIDTH
```
igrr, line 180: It's okay to use such configuration options in example projects, but for reusable libraries the configuration has to happen at run time, unless there is no technically feasible way of doing it at run time.

We already have an API to initialize the SDMMC host, why not leave this to the caller (i.e. application) and simply accept sdmmc_card_t* as one of the arguments of tinyusb_msc_storage_init_sdcard?

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,166 @@
...
+typedef struct {
+    // Place holder. Currently, nothing to configure
+} tinyusb_msc_spiffs_config_t;
+
+/**
+ * @brief Configuration structure for storage initialization
+ */
+typedef struct {
+    tinyusb_msc_storage_type_t storage_type;
+    union {
```
igrr, line 58: As mentioned above, recommend having two separate functions for initializing MSC with SPI flash and SDMMC, in that case each can have its own config structure.

(Not only this allows the program to dynamically choose which of the implementations to use, it also allows the linker to remove unused code if one of the implementations is not used.)

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
```diff
@@ -0,0 +1,196 @@
+/*
+ * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// DESCRIPTION:
+// This file contains the code for accessing the storage medium i.e. SPI Flash.
```
igrr, line 8: Please use "spiflash" or "spi_flash" in the file name, then. "spiffs" is usually associated with SPIFFS file system, which is a particular type of file system on top of SPI flash.

usb/esp_tinyusb/tusb_msc_storage_sdmmc.c
```diff
@@ -0,0 +1,276 @@
...
+fail:
+    if (fs) {
+        f_mount(NULL, drv, 0);
+    }
+    esp_vfs_fat_unregister_path(base_path);
+    ff_diskio_unregister(pdrv);
+    return err;
+}
+
+esp_err_t tinyusb_msc_storage_mount(const char *base_path)
```
igrr, line 175: Why do we need a wrapper for this functionality in `esp_tinyusb`? Isn't it already possible to mount a FAT filesystem in SPI Flash using IDF function `esp_vfs_fat_spiflash_mount_rw_wl`?

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
```diff
@@ -0,0 +1,196 @@
...
+{
+    size_t alloc_unit_size = requested_size;
+    const size_t max_sectors_per_cylinder = 128;
+    const size_t max_size = sector_size * max_sectors_per_cylinder;
+    alloc_unit_size = MAX(alloc_unit_size, sector_size);
+    alloc_unit_size = MIN(alloc_unit_size, max_size);
+    return alloc_unit_size;
+}
+
+esp_err_t tinyusb_msc_storage_mount(const char *base_path)
```
igrr, line 51: Likewise, why is this necessary when there is already `esp_vfs_fat_spiflash_mount_rw_wl`?

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
```diff
@@ -0,0 +1,196 @@
...
+}
+
+uint32_t tinyusb_msc_storage_get_sector_size(void)
+{
+    assert(s_wl_handle != WL_INVALID_HANDLE);
+
+    return (uint32_t)wl_sector_size(s_wl_handle);
+}
+
+esp_err_t tinyusb_msc_storage_read_sector(uint32_t lba, uint32_t offset, size_t size, void *dest)
```
igrr, line 173: Is this supposed to be called from anywhere in tinyusb? (i.e. from tud_msc_read10_cb?)

usb/esp_tinyusb/tusb_msc_storage_sdmmc.c
```diff
@@ -0,0 +1,276 @@
...
+        ESP_LOGE(TAG, "can't write, FAT mounted");
+        return ESP_ERR_INVALID_STATE;
+    }
+    size_t addr = lba * tinyusb_msc_storage_get_sector_size() + offset; // Address of the data to be read, relative to the beginning of the partition.
+    size_t sector_size = s_card->csd.sector_size;
+    if (addr % sector_size != 0 || size % sector_size != 0) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    ESP_RETURN_ON_ERROR(sdmmc_erase_sectors(s_card, lba, size / tinyusb_msc_storage_get_sector_size(), SDMMC_ERASE_ARG),
```
igrr, line 272: Unlike with SPI flash, it's not necessary to erase SD card sectors before writing — this is done by the card itself if necessary.

esp-saurabhbansal:
@igrr Thanks for the review comments.
@tore-espressif @roma-jam @Dazza0 PTAL.
I will soon be raising MR for usb device msc example code.

usb/esp_tinyusb/Kconfig
esp-saurabhbansal: I have now taken care of it.
Now, the APIs that are defined has one more argument - `const tinyusb_msc_storage_type_t storage_type`
This parameter will tell this driver to choose and operate on eaither spiflash or sd-card. Later the API usage can be extended for more media types without changing the API signature.

esp-saurabhbansal: I have now taken care of it.
Now, the APIs that are defined has one more argument - `const tinyusb_msc_storage_type_t storage_type`
This parameter will tell this driver to choose and operate on eaither spiflash or sd-card. Later the API usage can be extended for more media types without changing the API signature.

usb/esp_tinyusb/Kconfig
esp-saurabhbansal: Agree.
Now all the configuration of SDCard is done at the Application (example in our case) level.

usb/esp_tinyusb/include/tusb_msc_storage.h
esp-saurabhbansal: This is done in different way.
Instead of two different APIs, I am using a parameter `const tinyusb_msc_storage_type_t storage_type` to differentiate between two types.
The type of media is seleted by user at application (example) level using menuconfig. It is SPIFLASH by default.

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
esp-saurabhbansal: done

usb/esp_tinyusb/Kconfig
igrr: If you do this at run time based on storage_type argument, this means that both the SDMMC and Flash driver will be linked into the application, even if only one of the two is ever used.

Having two separate functions avoids the problem — the linker will be able to eliminate the code that is not called.

usb/esp_tinyusb/tusb_msc_storage_sdmmc.c
esp-saurabhbansal: There are two separate operations:
1. Initialisation
2. Mount

API `esp_vfs_fat_sdmmc_mount ` `is an all-in-one function` which performs both of these functionality.

We do not want to combine these operations in a single API because if user is accessing the storage media from USB Host, then only Init is required and media should not be mounted.
So, I have broken the functionality provided by `esp_vfs_fat_sdmmc_mount ` into two APIs `tinyusb_msc_storage_init` and `tinyusb_msc_storage_mount`.

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
esp-saurabhbansal: Above explanation of `esp_vfs_fat_sdmmc_mount ` holds good for `esp_vfs_fat_spiflash_mount_rw_wl ` as well.

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
esp-saurabhbansal: Yes.
`tinyusb_msc_storage_read_sector()` is called from tinyusb callback `tud_msc_read10_cb()`
Similarly, `tinyusb_msc_storage_write_sector()` is called from callback `tud_msc_write10_cb()`

usb/esp_tinyusb/tusb_msc_storage_sdmmc.c
esp-saurabhbansal: ok, I have removed the erase API from sd card code.

esp-saurabhbansal:
@igrr Thanks for the review comments. Please check if you agree.
@tore-espressif @roma-jam @Dazza0 PTAL.

usb/esp_tinyusb/Kconfig
esp-saurabhbansal: Currently, we have 7 common APIs for spi-flash and sdcard.
If we separate out APIs for spiflash and sdcard, it will be 14 APIs. In future, if more media type is supported, there will be an increase of number of APIs. The code size will increase with more files.
I request you check the current implementation once more.

igrr: In IDF, this dependency issue is usually solved using a factory method pattern, something along these lines:

```c++
// header file:


// forward declaration of the opaque handle
typedef struct mylibrary_handle_s *mylibrary_handle_t;

// factory method for implementation 1:
esp_err_t mylibrary_init_impl_aaa(const mylibrary_config_aaa_t* config, mylibrary_handle_t* out_handle);

// factory method for implementation 2:
esp_err_t mylibrary_init_impl_bbb(const mylibrary_config_bbb_t* config, mylibrary_handle_t* out_handle);

// common method for every implementation
esp_err_t mylibrary_do_stuff(mylibrary_handle_t handle, int some_arg);
```
```c++
// source file:

struct mylibrary_handle_s {
    // implementation function pointer
    esp_err_t (*do_stuff_impl)(int);
}

esp_err_t mylibrary_do_stuff(mylibrary_handle_t handle, int some_arg) {
    // indirect call
    return (handle->do_stuff_impl)(some_arg);
}

static esp_err_t mylibrary_aaa_do_stuff(mylibrary_handle_t handle, int some_arg)
{
    // concrete implementation for aaa
}

static esp_err_t mylibrary_bbb_do_stuff(mylibrary_handle_t handle, int some_arg)
{
    // concrete implementation for bbb
}

esp_err_t mylibrary_init_impl_aaa(const mylibrary_config_aaa_t* config, mylibrary_handle_t* out_handle)
{
    mylibrary_handle_t handle = (allocate struct mylibrary_handle_s, error check...)
    handle->do_stuff_impl = &mylibrary_aaa_do_stuff;
    *out_handle = handle;
    return ESP_OK;
}

esp_err_t mylibrary_init_impl_bbb(const mylibrary_config_bbb_t* config, mylibrary_handle_t* out_handle)
{
    mylibrary_handle_t handle = (allocate struct mylibrary_handle_s, error check...)
    handle->do_stuff_impl = &mylibrary_bbb_do_stuff;
    *out_handle = handle;
    return ESP_OK;
}
```

This way, only those implementations of functions which are actually needed are linked into the final binary. You can also easily extend the implementation to new subclasses by adding more factory methods and implementations.

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,359 @@
...
+}
+
+static inline size_t esp_vfs_fat_get_allocation_unit_size(
+    size_t sector_size, size_t requested_size)
+{
+    size_t alloc_unit_size = requested_size;
+    const size_t max_sectors_per_cylinder = 128;
+    const size_t max_size = sector_size * max_sectors_per_cylinder;
+    alloc_unit_size = MAX(alloc_unit_size, sector_size);
+    alloc_unit_size = MIN(alloc_unit_size, max_size);
```
roma-jam, line 118: IMHO this is not so clear to the developer what is happening here. 
Probably, the changes and a ternary operator could help us to make it a little bit better:

    alloc_unit_size = MAX(requested_size, sector_size);
    alloc_unit_size = (alloc_unit_size < max_size)? alloc_unit_size : max_size;

With such code we also can more easily (and clearer to the developer) cover the boundary conditions when alloc_unit_size is equal to max_size.

roma-jam (approved):
Hi @esp-saurabhbansal, 

Well done! Great job and nice PR!

I have left few comments, that pointed to the places which is probably needs to be verified. 

Also, Ivan had left few comments about the whole architecture and otherwise I do not see anything else. 

Except points mentioned before, LGTM. 

Regards, 
Roman

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,359 @@
...
+    s_card = (sdmmc_card_t *)malloc(sizeof(sdmmc_card_t));
+    ESP_GOTO_ON_FALSE(s_card, ESP_ERR_NO_MEM, clean, TAG, "could not allocate new sdmmc_card_t");
+
+    ESP_GOTO_ON_ERROR((*host.init)(), clean, TAG, "Host Config Init fail");
+    host_init = true;
+
+    ESP_GOTO_ON_ERROR(sdmmc_host_init_slot(host.slot, (const sdmmc_slot_config_t *) &slot_config),
+                      clean, TAG, "Host init slot fail");
+
+    while (sdmmc_card_init(&host, s_card)) {
```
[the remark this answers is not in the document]
tore-espressif: This is probably a leftover from the original example. It definitely deserves better error handling.

As far as I understand it, we don't even need to retry on failure. Just inform the user that the SD card cannot be mounted

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,359 @@
...
+        ESP_LOGE(TAG, "The detection pin of the slot is disconnected(Insert uSD card). Retrying...");
+        vTaskDelay(pdMS_TO_TICKS(3000));
+    }
+
+    // Card has been initialized, print its properties
+    sdmmc_card_print_info(stdout, s_card);
```
tore-espressif, line 71: Probably unnecessary. Does the user need to know any additional about the SD card? (apart form size, available from `tinyusb_msc_storage_get_sector_count/size`)

tore-espressif (commented):
@esp-saurabhbansal thank you for the follow-up!
I generally agree with the design reviews from Ivan and Roman.

Just one question: Does the SD card API offer a way of handling sudden disconnections of the SD cards? With SD cards, sudden disconnections can be expected and we should handle them gracefully

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
igrr: Could you point me where the call is done? I searched this MR for `tud_msc_read10_cb` function definition and didn't find it.

tore-espressif:
@esp-saurabhbansal We have USB MSC Host test implemented, where we also use mock USB Device https://github.com/espressif/idf-extra-components/blob/master/usb/usb_host_msc/test/msc_device.c

Could you rewrite the mock device so it uses your MSC device layer? We could see how the API is meant to be used, whether it compiles and whether it works with our MSC Host driver

usb/esp_tinyusb/include/tusb_msc_storage.h
esp-saurabhbansal: `tinyusb_msc_storage_type_t storage_type;` is not required anymore.

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,172 @@
...
+ */
+typedef struct {
+    union {
+        tinyusb_msc_spiflash_config_t *spiflash_config;
+#if SOC_SDMMC_HOST_SUPPORTED
+        tinyusb_msc_sdmmc_config_t *sdmmc_config;
+#endif
+    };
+} tinyusb_msc_storage_config_t;
+/*********************************************************************** structs, unions */
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,359 @@
...
+    ESP_LOGI(TAG, "Initializing SDCard");
+
+    sdmmc_host_t host = config->sdmmc_config->host;
+    sdmmc_slot_config_t slot_config = config->sdmmc_config->slot_config;
+
+    // not using ff_memalloc here, as allocation in internal RAM is preferred
+    s_card = (sdmmc_card_t *)malloc(sizeof(sdmmc_card_t));
+    ESP_GOTO_ON_FALSE(s_card, ESP_ERR_NO_MEM, clean, TAG, "could not allocate new sdmmc_card_t");
+
+    ESP_GOTO_ON_ERROR((*host.init)(), clean, TAG, "Host Config Init fail");
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/esp_tinyusb/tusb_msc_storage.c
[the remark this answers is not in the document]
esp-saurabhbansal: Agree.
Earlier, this implementation was part of example code. So, retry was a good approach.
But now it is removed.

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: The function definition was redundant, removed.

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,359 @@
...
+{
+    size_t alloc_unit_size = requested_size;
+    const size_t max_sectors_per_cylinder = 128;
+    const size_t max_size = sector_size * max_sectors_per_cylinder;
+    alloc_unit_size = MAX(alloc_unit_size, sector_size);
+    alloc_unit_size = MIN(alloc_unit_size, max_size);
+    return alloc_unit_size;
+}
+
+esp_err_t tinyusb_msc_storage_mount(const tinyusb_msc_storage_type_t storage_type, const char *base_path)
```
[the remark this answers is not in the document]
esp-saurabhbansal: `connect driver` and `vfs register` was using single APIs.
Separated the code for `mount`

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,359 @@
...
+    return result;
+}
+
+esp_err_t tinyusb_msc_storage_read_sector(const tinyusb_msc_storage_type_t storage_type, uint32_t lba, uint32_t offset, size_t size, void *dest)
+{
+    esp_err_t ret = ESP_FAIL;
+    size_t sector_size = tinyusb_msc_storage_get_sector_size(storage_type);
+    switch (storage_type) {
+    case TINYUSB_MSC_STORAGE_SPI:
+        size_t addr = lba * sector_size + offset; // Address of the data to be read, relative to the beginning of the partition.
```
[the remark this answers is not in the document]
esp-saurabhbansal: `size_t` is of type `unsigned long`.
`addr` is of type size_t and thus can accommodate addresses upto 4GB.
This implementation is used for SPI Flash. 4GB size is enough for SPI Flash.
Implementation of module `SPI Flash wear-levelling` itself uses type `size_t` in its exposed APIs.

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: done.
Removed the printing of this additional information.

usb/esp_tinyusb/tusb_msc_storage_spiffs.c
esp-saurabhbansal: It is done in example code 'examples/peripherals/usb/device/tusb_msc/main/tusb_msc_main.c'.
I will raise a MR for updated example.

igrr: Wait, isn't the point of this component to provide the definitions of these tinyusb callback functions?
For example, tusb_cdc_acm.c defines the callback functions which tinyusb will call for CDC class:
https://github.com/espressif/idf-extra-components/blob/211403a548a9507ca22f0968b22b6f66a1e80694/usb/esp_tinyusb/tusb_cdc_acm.c#L83

The the application (example) becomes very simple for the developer: just initialize USB and initialize this component, no need to understand which tinyusb callbacks need to be defined for a particular class.

esp-saurabhbansal: These are the callbacks which must be defined: 

a. void tud_msc_capacity_cb(uint8_t lun, uint32_t *block_count, uint16_t *block_size)
b. int32_t tud_msc_read10_cb(uint8_t lun, uint32_t lba, uint32_t offset, void *buffer, uint32_t bufsize)
c. int32_t tud_msc_write10_cb(uint8_t lun, uint32_t lba, uint32_t offset, uint8_t *buffer, uint32_t bufsize)
d. int32_t tud_msc_scsi_cb(uint8_t lun, uint8_t const scsi_cmd[16], void *buffer, uint16_t bufsize)

e. void tud_msc_inquiry_cb(uint8_t lun, uint8_t vendor_id[8], uint8_t product_id[16], uint8_t product_rev[4]);
f. bool tud_msc_test_unit_ready_cb(uint8_t lun)
g. bool tud_msc_start_stop_cb(uint8_t lun, uint8_t power_condition, bool start, bool load_eject)
h. void tud_umount_cb(void)
i. void tud_mount_cb(void)

Currently, the application example is implemented such that:
`It either allows the embedded application i.e. example to access the partition or Host PC accesses the partition over USB MSC.
They can't be allowed to access the partition at the same time.`

In order to meet this requirement, the last five callbacks has to be implemented in application example code.
All the direct callbacks of tinyusb are implemented in application code (as done in msc device code currently).
The other way to implement this functionality is by implementing user-defined callbacks at application code (as done in cdc - tinyusb_cdcacm_register_callback()).

At the very beginning, I thought to implement the callback mechanism the way cdc has implemented. But then I thought that what is the use of simply wrapping around the callback functions.
Yes, it gives some convinience to user - User does not have to define the callback functions mandatorily.

If you think, the cdc approach is better then please suggest if last five callbacks should be exposed to user OR all nine callbacks.
In my opinion, only last five are required to user.

esp-saurabhbansal:
@tore-espressif I came to know that currently, there is no API (callback) that can be used to detect sdcard disconnect. I have informed the owner of sdmmc driver module owner about this requirement.
In future, if such API or method is implemented, we can update our msc driver accordingly.

tore-espressif:
> Could you rewrite the mock device so it uses your MSC device layer?

@esp-saurabhbansal any news about the tests?

esp-saurabhbansal:
> > Could you rewrite the mock device so it uses your MSC device layer?
> 
> @esp-saurabhbansal any news about the tests?

@tore-espressif All the TCs are passing now. There is a separate commit for TC modification. Mock device implementation is modified. Plus, there are few changes which are done to make TCs in pipeline pass for different targets. Please check.

usb/usb_host_msc/test/msc_device.c
```diff
@@ -1,92 +1,155 @@
 /*
```
tore-espressif, line 1: General comment to this file:
The main idea of refactoring this MSC mock device, was to get rid of the TinyUSB callbacks (such as `tud_msc_test_unit_ready_cb`, `tud_msc_capacity_cb`, `tud_***_cb`...) and use the callbacks implementation from `esp_tinyusb`.

The goal is to showcase that implementing a MSC device with `esp_tinyusb` is easy -> that the user does _not_ have to bother with defining all these TinyUSB callbacks. Could you please take a look whether we can remove these callback from this file?

tore-espressif (changes requested):
Hello @esp-saurabhbansal , thank you very much for the follow-up!
I left few comments that should be addressed before merging.
I'll try it on HW later this week :) thanks!

usb/usb_host_msc/test/msc_device.c
```diff
@@ -1,92 +1,155 @@
...
- *
- */
 
-#include <stdint.h>
 #include "esp_log.h"
-#include "freertos/FreeRTOS.h"
-#include "freertos/task.h"
 #include "tinyusb.h"
+#include "driver/gpio.h"
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,146 @@
...
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stddef.h>
+#include "esp_err.h"
+#include "driver/sdmmc_host.h"
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/usb_host_msc/test/CMakeLists.txt
```diff
@@ -1,10 +1,16 @@
 include($ENV{IDF_PATH}/tools/cmake/version.cmake)
 set (TINYUSB_LIB)
+set (srcs)
 if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.0")
     set(TINYUSB_LIB "esp_tinyusb")
+    list(APPEND srcs "../../esp_tinyusb/tusb_msc_storage.c")
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/usb_host_msc/test/msc_device.c
```diff
@@ -1,92 +1,155 @@
...
+#define TUSB_DESC_TOTAL_LEN (TUD_CONFIG_DESC_LEN + TUD_MSC_DESC_LEN)
+
+enum {
+    ITF_NUM_MSC = 0,
+    ITF_NUM_TOTAL
+};
+
+enum {
+    EDPT_CTRL_OUT = 0x00,
+    EDPT_CTRL_IN  = 0x80,
```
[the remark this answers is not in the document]
esp-saurabhbansal: removed.
done.

usb/test_app/main/Kconfig.projbuild
```diff
@@ -0,0 +1,69 @@
+menu "USB Dev MSC Example Configuration"
```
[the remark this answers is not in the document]
esp-saurabhbansal: correct.
I have made changes accordingly.

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,151 @@
...
+} tinyusb_msc_spiflash_config_t;
+
+/**
+ * @brief Configuration structure for storage initialization
+ */
+typedef struct {
+    union {
+        tinyusb_msc_spiflash_config_t *spiflash_config;
+#if SOC_SDMMC_HOST_SUPPORTED
+        tinyusb_msc_sdmmc_config_t *sdmmc_config;
```
tore-espressif, line 51: Could these be values instead of pointers?

tore-espressif (commented):
@esp-saurabhbansal I left few nitpicks, otherwise LGTM! 

FYI @igrr  this is ready for review

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,562 @@
...
+    esp_err_t (*read)(size_t sector_size, uint32_t lba, uint32_t offset, size_t size, void *dest);
+    esp_err_t (*write)(size_t sector_size, size_t addr, uint32_t lba, uint32_t offset, size_t size, const void *src);
+} tinyusb_msc_storage_handle_s; /*!< MSC object */
+
+/* handle of tinyusb driver connected to application */
+static tinyusb_msc_storage_handle_s *s_storage_handle;
+
+static esp_err_t _init_spiflash(const tinyusb_msc_storage_config_t *config)
+{
+    const esp_partition_t *data_partition = esp_partition_find_first(ESP_PARTITION_TYPE_DATA, ESP_PARTITION_SUBTYPE_DATA_FAT, NULL);
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/esp_tinyusb/include/tusb_msc_storage.h
esp-saurabhbansal: done

usb/usb_host_msc/test/msc_device.c
esp-saurabhbansal: tinyusb callbacks are defined in file msc_test.c for idf v4.4.
idf v4.4 dos not use esp_tinyusb (spiflash and sdcard), So, we cannot remove tinyusb callbacks definition from TC file msc_device.c

usb/usb_host_msc/test/test_common.h
```diff
@@ -11,9 +11,11 @@ enum {
     DISK_BLOCK_SIZE = 512
 };
 
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
```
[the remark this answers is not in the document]
esp-saurabhbansal: In my opinion test_app for version v4.4 should not use esp_tinyusb because esp_tinyusb is introduced only in v5.0.

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,563 @@
...
+ * \return      Actual bytes processed, can be zero for no-data command.
+ * \retval      negative    Indicate error e.g unsupported command, tinyusb will \b STALL the corresponding
+ *                          endpoint and return failed status in command status wrapper phase.
+ */
+int32_t tud_msc_scsi_cb(uint8_t lun, uint8_t const scsi_cmd[16], void *buffer, uint16_t bufsize)
+{
+    int32_t ret;
+
+    switch (scsi_cmd[0]) {
+    case SCSI_CMD_PREVENT_ALLOW_MEDIUM_REMOVAL:
```
tore-espressif, line 537: Could we implement this feature? So the user cannot access the partition while it is, for example, being formated by the USB host?

usb/usb_host_msc/test/test_msc.c
```diff
@@ -281,12 +269,14 @@ TEST_CASE("sectors_can_be_written_and_read", "[usb_msc]")
     msc_teardown();
 }
 
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
 TEST_CASE("check_README_content", "[usb_msc]")
```
tore-espressif, line 273: Is anything missing in the current implementation? Why can't we implement and run this test?

usb/usb_host_msc/test/msc_device.c
```diff
@@ -1,100 +1,197 @@
...
- *
- */
 
-#include <stdint.h>
 #include "esp_log.h"
-#include "freertos/FreeRTOS.h"
-#include "freertos/task.h"
 #include "tinyusb.h"
-#include "test_common.h"
-#include "soc/soc_caps.h"
```
tore-espressif, line 41: You are using defines from `soc/soc_caps.h` (SOC_SDMMC_HOST_SUPPORTED), so I think we should leave this include in the file

tore-espressif (approved):
@esp-saurabhbansal thanks again for your work!
I left few minor comments that do not necessarily need to be  implemented in this PR

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * @brief Mount the storage partition locally on the firmware application.
+ *
+ * Get the available drive number. Register spi flash partition.
+ * Connect POSIX and C standard library IO function with FATFS.
+ * Mounts the partition.
+ * This API is used by the firmware application. If the storage partition is
+ * mounted by this API, host (PC) can't access the storage via MSC.
+ *
+ * @param base_path  path prefix where FATFS should be registered
```
igrr, line 104: Given that you have this argument, perhaps `CONFIG_TINYUSB_MSC_MOUNT_PATH` option can be removed? (Just one less way to do the same thing => might lead to a simpler API.)

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * @brief Get sector size of storage media
+ *
+ * @return sector count
+ */
+uint32_t tinyusb_msc_storage_get_sector_size(void);
+
+/**
+ * @brief Get status if storage iedia is exposed over USB to Host
```
igrr, line 141:
```suggestion
 * @brief Get status if storage media is exposed over USB to Host
```

usb/esp_tinyusb/include/tusb_msc_storage.h
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * @return bool
+ *      - true, if the storage media is exposed to Host
+ *      - false, if the stoarge media is mounted on application (not exposed to Host)
+ */
+bool tinyusb_msc_storage_is_exposed(void);
```
igrr, line 147: Not sure i would understand what "exposed" means if I saw just the usage of this function in the code. How about `tinyusb_msc_storage_in_use_by_usb_host`?

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,563 @@
...
+static esp_err_t _init_sdmmc(const tinyusb_msc_storage_config_t *config)
+{
+    esp_err_t ret = ESP_OK;
+
+    sdmmc_host_t host = config->sdmmc_config.host;
+    sdmmc_slot_config_t slot_config = config->sdmmc_config.slot_config;
+
+    ESP_RETURN_ON_ERROR((*host.init)(), TAG, "Host Config Init fail");
+
+    ESP_GOTO_ON_ERROR(sdmmc_host_init_slot(host.slot, (const sdmmc_slot_config_t *) &slot_config),
```
igrr, line 133: What about keeping the sdmmc initialization outside of this component, in the user's code?

The advantage of such approach is that if the user's application has initialized the card, you can access both cards over SDMMC and SDSPI using the same API.

It is also hard to explain to the user why the SD card initialization in their application would be handled by... esp_tinyusb component, of all the things.

How about changing the initialization APIs to:

```c++
// sdmmc_card_t* card is now part of "tinyusb_msc_sdmmc_config_t"
tinyusb_msc_storage_init_sdmmc(const tinyusb_msc_sdmmc_config_t *config); 

tinyusb_msc_storage_init_spiflash(const tinyusb_msc_spiflash_config_t* config);
```

Both functions can initialize `s_storage_handle`, as your `register...` APIs do now, and then call the common internal `init` implementation for both cases.

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -0,0 +1,563 @@
...
+        ret = -1;
+        break;
+    }
+    return ret;
+}
+
+// Invoked when device is unmounted
+void tud_umount_cb(void)
+{
+    ESP_ERROR_CHECK(tinyusb_msc_storage_mount(s_storage_handle->base_path));
```
igrr, line 555: We usually try to avoid aborting on an error inside IDF components. Is there no way to gracefully handle the error here, other than aborting?

igrr (changes requested):
@esp-saurabhbansal looks very good, thanks for addressing my comments.

I have one remaining request about moving sdcard driver (sdmmc or sdspi) initialization out of this component into the user code.

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: OK.
I will address this in next update.

usb/usb_host_msc/test/test_msc.c
esp-saurabhbansal: This TC is only applicable for software based fake device where `README_CONTENTS` are already stored.
The purpose of this TC is to read the contents of file `/usb/README.TXT` from the fake device and match it with expected contents.
This type of fake device is made only for IDF v4.4.

For IDF v5.0 onwards, such fake device is not required as spiflash or sdcard medium is used as storage device. In this case the pupose is solved by TC `write_and_read_file`

usb/usb_host_msc/test/msc_device.c
esp-saurabhbansal: done

usb/esp_tinyusb/CMakeLists.txt
```diff
@@ -13,9 +13,10 @@ set(srcs
     )
 
 set(priv_requires
-    usb
-    vfs
-    esp_ringbuf
+    "usb"
+    "vfs"
+    "esp_ringbuf"
+    "fatfs"
```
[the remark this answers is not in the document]
esp-saurabhbansal: done

usb/esp_tinyusb/include/tusb_msc_storage.h
esp-saurabhbansal: done

usb/esp_tinyusb/include/tusb_msc_storage.h
esp-saurabhbansal: done

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: I have removed ESP_ERROR_CHECK from this file for now.

I can see that there are many assert statement being used in whole `idf-extra-components/usb`. I will discuss and plan to remove these in subsequent PR.

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: The initialisation part is now shifted to user application for both spiflash and sdmmc to keep the symmetry.
APIs are changed accordingly.
@igrr   PTAL.

usb/esp_tinyusb/include/tusb_msc_storage.h
esp-saurabhbansal: @igrr 

This is done for taking care of the following scenario:
1. Media is exposed over usb host.
2. Next, it is manually ejected.
3. Expected behaviour - media should no longer be exposed over usb host + media should be mounted automatically to MOUNT_PATH.

Current behaviour is point 3 such that it is automatically mounted to CONFIG_TINYUSB_MSC_MOUNT_PATH by component (idf-extra-component) without any intervention of application code. This CONFIG option is to tell the component about the mount path when component has to mount the media by itself. In this case, application code is not calling mount API `esp_err_t tinyusb_msc_storage_mount(const char *base_path);` to mount the storage. I can mention about this default path in doxygen documentation of API tinyusb_msc_storage_mount().

PR #158, opened 2023-03-08
USB modem DTE v1.1

tore-espressif:
# Change description
- Update to [CDC-ACM driver](https://components.espressif.com/components/espressif/usb_host_cdc_acm) to v2
- Provide default configurations for tested modems
- Fix USB receive path bug, where received data could be overwritten by new data
- Initial support for modems with two AT ports


tore-espressif:
@david-cermak In this PR, the `esp_mode_usb_dte` extension finally supports dual-terminal USB modems.

there will be another follow-up in esp-modem's examples, PTAL!

usb/esp_modem_usb_dte/esp_modem_usb_api_target.cpp
```diff
@@ -14,9 +17,20 @@ static const char *TAG = "modem_usb_api_target";
...
 {
+    if (config == nullptr) {
+        return nullptr;
+    }
+
+    // *INDENT-OFF*
     TRY_CATCH_RET_NULL(
-        auto term = create_usb_terminal(config);
-        return std::make_shared<DTE>(config, std::move(term));
+        const struct esp_modem_usb_term_config *usb_config = (struct esp_modem_usb_term_config *)(config->extension_config);
```
david-cermak, line 26: This doesn't look very much like `C++` style. Sorry, probably missed in the previous review. Suggest something like:
```suggestion
        auto *usb_config = static_cast<struct esp_modem_usb_term_config *>(config->extension_config);
```

cc @euripedesrocha (if you have a better idea, but the `extension_config` could be also used from plain C interface, so I think the `void*` makes sense here)

usb/esp_modem_usb_dte/esp_modem_usb_api_target.cpp
```diff
@@ -1,11 +1,14 @@
 /*
- * SPDX-FileCopyrightText: 2022 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2022-2023 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #include "usb_terminal.hpp"
 #include "cxx_include/esp_modem_api.hpp"
+#include "../private_include/exception_stub.hpp"
```
david-cermak, line 9: Nitpick, could we add this to `PRIVATE_INCLUDE` paths?
and simply do:
```suggestion
#include "exception_stub.hpp"
```
(Just a personal preference, but don't very like relative `../` in includes, as they could be tricky -- for example can violate CMake canonical include path approach)

david-cermak (approved):
LGTM, just left some nitpicks.

usb/esp_modem_usb_dte/esp_modem_usb_api_target.cpp
tore-espressif: This private include comes from `esp_modem` so I updated `usb/esp_modem_usb_dte/CMakeLists.txt` to include it, PTAL!

PR #146, opened 2023-02-20
Add libpng and zlib to Component Manager

shootao:
# Checklist

- [x] Component contains License
- [x] Component contains README.md
- [x] Component contains idf_component.yml file with `url` field defined
- [x] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [x] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description
_Please describe your change here_
Add png decoding library

PR #160, opened 2023-03-22
Add libpng 1.6.39 and zlib 1.2.13

igrr:
# Checklist

- [ ] CI passing

# Change description

Replaces the accidentally closed https://github.com/espressif/idf-extra-components/pull/146, adding libpng and zlib.

Authored by @shootao, with some minor fixups done by me.


espzav:
And one additional thing: when I used in LVGL task, it takes lot of stack size, I had to increase stack of the LVGL task.

espzav:
Here is my example with PNG: https://github.com/espressif/esp-bsp/tree/example/add_png_support/examples/display_audio_photo

igrr:
@espzav Thanks for review!

For LVGL, i think we don't need to use libpng. I was able to load PNG images in LVGL directly, using the steps below:

1. Enable CONFIG_LV_USE_PNG=y, CONFIG_LV_USE_FS_POSIX=y, CONFIG_LV_FS_POSIX_LETTER=65
2. Mount a filesystem, for example to /data
3. Call `lv_canvas_draw_img(canvas, x, y, "A:/data/pic.png", &draw_img_dsc)`

I agree that libpng API is not easy to use. We need libpng only when porting some other libraries which already use libpng API. In this case, libpng is not called by the user directly.

PR #161, opened 2023-03-27
ci: fix libpng not uploaded since there's no zlib in the registry yet

igrr:
Unfortunately sorting the list of components in https://github.com/espressif/idf-extra-components/pull/160 had one disadvantage: if two new components are added in the same PR and the (alphabetically) 1st one depends on the 2nd one, the upload of the 1st component will fail (https://github.com/espressif/idf-extra-components/actions/runs/4512604342/jobs/7946288243).

This PR moves libpng to the end of the list so that it gets uploaded after zlib is uploaded for the first time. We can revert to the correct sorting order in a follow-up PR.

### DIFF
--- a/.github/workflows/upload_component.yml
+++ b/.github/workflows/upload_component.yml
@@ -15,6 +15,15 @@ jobs:
       - name: Upload components to component service
         uses: espressif/upload-components-ci-action@v1
         with:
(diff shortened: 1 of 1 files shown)


tore-espressif:
Thanks for the fix, and sorry for the trouble @igrr , LGTM!

PR #168, opened 2023-04-21
Adding workflows for GH -> Jira issue mirroring

tom-borcin:
Adding workflows for mirroring of issues from GitHub to Jira

### DIFF
--- a/.github/workflows/new_issues.yml
+++ b/.github/workflows/new_issues.yml
@@ -1,0 +1,23 @@
+name: Sync issues to Jira
+
+# This workflow will be triggered when a new issue is opened
+on: issues
+
+# Limit to single concurrent run for workflows which can create Jira issues.
+# Same concurrency group is used in issue_comment.yml
+concurrency: jira_issues
+
+jobs:
+  sync_issues_to_jira:
+    name: Sync issues to Jira
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Sync GitHub issues to Jira project
+        uses: espressif/github-actions/sync_issues_to_jira@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


tom-borcin:
@tore-espressif @kumekay hello guys! please take a look

.github/workflows/new_issues.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: Sync issues to Jira
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Sync GitHub issues to Jira project
+        uses: espressif/github-actions/sync_issues_to_jira@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          JIRA_PASS: ${{ secrets.JIRA_PASS }}
+          JIRA_PROJECT: IEC
```
igrr, line 21: Could we also set the component which the issue is associated with?

For example, add a "component" field to the issue form, and the let sync_issues_to_jira set the correct component in Jira?

tom-borcin: That's the problem I was trying to solve recently. We are asking for component in issue form drop-down, but I couldn't find a solution that would pass that component to an issue label. I found only solution that would work for issue templates, not for forms.

I will set up automation in Jira that will parse component from description and set it in Jira and also in GitHub.

We could also modify a script to do it, but we still need automation in case component changes in Jira.

igrr: I see! Should we at least set some default component here (e.g. `JIRA_COMPONENT: Github`), or will we triage new issues by filtering the ones without any component set?

tom-borcin: I prefer leaving it empty, because we would have to make one extra step to remove Github component and adding new one. If there is no component, we just add the component. Component drop-down is required and allows only one value to be set. I can't think of an issue that will relate to two or more components.

igrr (approved):

tore-espressif (approved):
LGTM!

PR #171, opened 2023-04-24
1-wire: add 1-wire bus driver

suda-morris:
Added onewire_bus library based on RMT peripheral driver. Previously this library was located in the esp-idf's example: https://github.com/espressif/esp-idf/tree/master/examples/peripherals/rmt/onewire_ds18b20/components/onewire_bus

Now I have refactored that, and split it into multiple parts: bus, device and other helper functions (e.g. CRC).

As the one-wire bus implementation can be various, so added an interface to abstract the bus behavior. Easy to extend with other backend peripherals.

Fixed an issue that reported in https://github.com/espressif/esp-idf/issues/10790

examples and doc will be added in the later PR(s).

device driver like "ds18b20" will go into esp-bsp repo, after this PR got merged.


onewire_bus/src/onewire_crc.c
```diff
@@ -0,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: 2022-2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "onewire_crc.h"
+
+#define FAST_CRC 1 // define this to use the fast CRC table
```
tore-espressif, line 9: Do we expect users to change this to '0' often?

onewire_bus/include/onewire_bus_impl_rmt.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief 1-Wire bus RMT specific configuration
+ */
+typedef struct {
+    uint32_t max_rx_bytes; /*!< Set the largest possible single receive size,
```
tore-espressif, line 20: Could we add a note that this is relevent only for esp32 and esp32s2?

onewire_bus/src/onewire_bus_impl_rmt.c
```diff
@@ -0,0 +1,490 @@
...
+    return ret;
+}
+
+static void onewire_rmt_decode_data(rmt_symbol_word_t *rmt_symbols, size_t symbol_num, uint8_t *decoded_bytes)
+{
+    size_t byte_pos = 0;
+    size_t bit_pos = 0;
+    for (size_t i = 0; i < symbol_num; i ++) {
+        if (rmt_symbols[i].duration0 > ONEWIRE_SLOT_BIT_SAMPLE_TIME) { // 0 bit
+            decoded_bytes[byte_pos] &= ~(1 << bit_pos); // LSB first
```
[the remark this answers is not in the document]
tore-espressif: this seems like a valid warning. Could you please check @suda-morris ?

onewire_bus/src/onewire_bus_impl_rmt.c
```diff
@@ -0,0 +1,490 @@
...
+        } else { // 1 bit
+            decoded_bytes[byte_pos] |= 1 << bit_pos;
```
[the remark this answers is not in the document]
tore-espressif: same here

tore-espressif (commented):
I left few question/comments, otherwise LGTM!

You also mentioned setting up a runner in https://github.com/espressif/esp-idf/issues/10790#issuecomment-1514544277 , do you plan to connect the runner to esp-idf or in idf-extra-components?

onewire_bus/src/onewire_crc.c
suda-morris: I think in most cases, users won't touch this. Another implementation was added for my understanding of its underlying CRC. :smiley:

onewire_bus/include/onewire_bus_impl_rmt.h
suda-morris: no, not only for esp32 and esp32s2. It decides the receive buffer we used to save the RMT symbols. But for esp32 and esp32s2, this configuration has more meaning, we also use it to set the proper size of the RMT hardware memory block.

onewire_bus/src/onewire_crc.c
tore-espressif: OK, because modifying sources inside `managed_components` dir leads to warnings. Let's keep it as it is for now

PR #173, opened 2023-04-27
cache compensated timer for measuring performance metric

suda-morris:
Play around with git filter-repo, I managed to keep the git history. 

This ccomp_timer was previously located in the esp-idf's tools folder. We found it can be useful for other components' performance tests, so decide to push it to the component registry.

### DIFF
--- a/ccomp_timer/README.md
+++ b/ccomp_timer/README.md
@@ -1,0 +1,9 @@
+# Cache Compensated Timer
+
+[![Component Registry](https://components.espressif.com/components/espressif/ccomp_timer/badge.svg)](https://components.espressif.com/components/espressif/ccomp_timer)
+
+The **Cache Compensated Timer** is a timer that tries to account for instruction and data stall cycles caused by cache misses. It is useful for measuring the time spent in a function or a block of code.
+
+On Xtensa targets (e.g. ESP32), the timer is built on top of the debug module's performance monitor counter.
+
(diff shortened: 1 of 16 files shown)


igrr:
@suda-morris do you think we could squash some of the commits here? A lot of these seem to be "change something unrelated to ccomp_timer and oh by the way also something about ccomp_timer". Perhaps we can keep the original commits from Renz for attribution and then squash most of the remaining ones? (Can keep commit IDs as part of the squashed commit message, just in case...)

suda-morris:
I squashed the origin commits, this time the commit history looks clean. @igrr  @tore-espressif

tore-espressif (approved):
@suda-morris formal aspects LGTM

ccomp_timer/README.md
```diff
@@ -0,0 +1,9 @@
+# Cache Compensated Timer
+
+[![Component Registry](https://components.espressif.com/components/espressif/ccomp_timer/badge.svg)](https://components.espressif.com/components/espressif/ccomp_timer)
+
+The **Cache Compensated Timer** is a timer that tries to account for instruction and data stall cycles caused by cache misses. It is useful for measuring the time spent in a function or a block of code.
+
+On Xtensa targets (e.g. ESP32), the timer is built on top ot the debug module's performance monitor counter.
+
+On RISC-V targets, this driver is fall-back to the CPU's cycle counter, which actually **doesn't** account for the cache miss.
```
ESP-Marius, line 9:
```suggestion
Due to hardware limitations, on RISC-V targets this driver falls back to using the CPU's cycle counter, which actually **doesn't** account for the cache misses. To achieve a measurement that is independent of cache misses you could place the code is to be measured into IRAM.
```

ESP-Marius (commented):
Moving it out LGTM!

Issue #29, opened 2022-03-19
Consider setting up dependabot to update submodule dependencies


igrr:
Dependabot has basic support for tracking and upgrading dependencies expressed using git submodules (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem). Consider setting it up for this repository to get notified whenever a dependency can be upgraded.

Note that we might need to wait for or contribute to https://github.com/dependabot/dependabot-core/issues/1639 first, as currently dependabot will try to upgrade to the latest commit, not to the latest tag.

tore-espressif:
Done.

I don't have any experience with dependabot, so I'll keep this issue open for future evaluation/discussion

igrr:
Thanks @tore-espressif!

The PRs opened by dependabot have two problems now:

1. The PR updates the submodule to the latest commit, not to the latest tag (issue https://github.com/dependabot/dependabot-core/issues/1639 noted above).
2. Looks like the "publish test results" job is failing.

Still even in this form they are useful as a hint/reminder to us that some dependency might be outdated.

igrr:
This might be worth revisiting, especially since https://github.com/espressif/idf-extra-components/pull/146 adds two libraries which are known to sometimes have CVEs reported for them. It would be good to be able to get the new releases of these libraries published quickly.

PR #174, opened 2023-04-30
ci: add an action to update submodules

igrr:
This PR adds a new CI action to automatically update submodules, opening PRs for updates.

Example PRs can be seen in the fork:
* https://github.com/igrr/idf-extra-components/pull/7
* https://github.com/igrr/idf-extra-components/pull/8

Uses the following Github action:
* https://github.com/espressif/github-actions/pull/40

Description of config options in .gitmodules can be found the [README.md of update_submodule_versions action](https://github.com/igrr/github-actions/blob/feature/update_submodule_versions/update_submodule_versions/README.md).

## TODO

- [ ] Set SUBMODULE_UPDATE_AUTHOR_NAME and SUBMODULE_UPDATE_AUTHOR_EMAIL variables in repository settings after merging

---
Closes https://github.com/espressif/idf-extra-components/issues/29

### DIFF
--- a/.github/workflows/update_submodules.yml
+++ b/.github/workflows/update_submodules.yml
@@ -1,0 +1,25 @@
+name: Update submodules
+
+on:
+  schedule:
+    - cron: '0 0 * * SUN' # Sunday midnight
+  workflow_dispatch:               # Or manually
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update_submodules:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          fetch-depth: 0
+      - name: Update submodules
+        uses: igrr/github-actions/update_submodule_versions@feature/update_submodule_versions
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          git-author-name: ${{ secrets.SUBMODULE_UPDATE_AUTHOR_NAME }}
+          git-author-email: ${{ secrets.SUBMODULE_UPDATE_AUTHOR_EMAIL }}
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,36 +1,100 @@
 [submodule "libsodium/libsodium"]
 	path = libsodium/libsodium
 	url = https://github.com/jedisct1/libsodium.git
+	# libsodium doesn't tag new releases since 0.18.0, 'stable' branch is considered
+	# to be the latest stable release. See:
+	# https://github.com/jedisct1/libsodium/issues/1041#issuecomment-799812417
+	# However since idf_component.yml had the version already set to 0.20.0,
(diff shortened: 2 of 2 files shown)


.github/workflows/update_submodules.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  update_submodules:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          fetch-depth: 0
+      - name: Update submodules
+        uses: igrr/github-actions/update_submodule_versions@feature/update_submodule_versions
```
igrr, line 21: Change to espressif/github-actions once the related PR is merged

.gitmodules
```diff
@@ -1,36 +1,100 @@
...
+	# we can't update the version automatically (it would be set it back to 0.18.0~n)
+	# autoupdate = false
 [submodule "cbor/tinycbor"]
 	path = cbor/tinycbor
 	url = https://github.com/intel/tinycbor.git
+	autoupdate = true
+	autoupdate-branch = main
+	autoupdate-tag-glob = v[0-9]*.[0-9]*.[0-9]*
+	autoupdate-manifest = cbor/idf_component.yml
+	autoupdate-ver-regex = v([0-9]+).([0-9]+).([0-9]+)
```
mahavirj, line 17: Should we consider keeping some default values for branch, tag-glob and ver-regex fields? If we take most common syntax (e.g., version as `vx.y.z`) then we can reduce the additional fields specified for all external submodules.

.gitmodules
```diff
@@ -1,36 +1,100 @@
...
+	autoupdate = true
+	autoupdate-branch = master
+	autoupdate-tag-glob = [0-9]*.[0-9]*.[0-9]*
+	autoupdate-include-lightweight = true
+	autoupdate-manifest = fmt/idf_component.yml
+	autoupdate-ver-regex = ([0-9]+).([0-9]+).([0-9]+)
 [submodule "esp_delta_ota/detools"]
 	path = esp_delta_ota/detools
 	url = https://github.com/eerimoq/detools.git
+	# not updating automatically, as this is not an upstream repo
```
mahavirj, line 73: We do use source from this repository in our component build. We should add this component in the auto update list.

.gitmodules
igrr: I kept these explicit in order to force the component maintainer to go and check the tag naming pattern the repository actually uses.

Alternatively I could make the action fail if not a single tag with the given naming pattern is found. In that case I'll also make the action run when .gitmodules file is changed, with a --dry-run argument.

tore-espressif:
LGTM! Let's wait for https://github.com/espressif/github-actions/pull/40

.gitmodules
mahavirj: Yes, they are not related. Our component follows its own versioning policy here. We can handle the update for this case on-demand basis.

Issue #164, opened 2023-04-11
`esp_tinyusb` console redirect should be gated behind VFS being enabled  (IEC-3)


dmezh:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

usb

### ESP-IDF version.

v5.0.1

### Development Kit.

Custom Board (ESP32-S3FN8)

### Used Component version.

1.1.0

### More Information.

With VFS disabled in menuconfig, `esp_tusb_init_console` eventually fails to redirect std streams and returns an error at runtime. This function should just not be available if VFS is not available. Maybe an `#ifdef CONFIG_VFS_SUPPORT...` and a note in the documentation?

tore-espressif:
Hello @dmezh , you are right, I'm sorry for the troubles.

Since this issue can be fixed with correct menuconfig option, we will fix it in upcoming esp_tinyusb update ~end of June.

dmezh:
Awesome, thank you @tore-espressif!

PR #178, opened 2023-05-18
esp_tinyusb: set cpu core affinity of tinyusb default task

esp-saurabhbansal:
1. This change is to set the CPU core affinity to Tinyusb Task.
If the task is not pinned we often see lost packets if the system is put under stress.

2. cdc console feature is enabled only when VFS is enabled.
This is done by compiling the console feature related files only when VFS option is enabled.
If VFS is not enabled, compile error will be thrown. So, it will avoid runtime failure.
Closes https://github.com/espressif/idf-extra-components/issues/164

### DIFF
--- a/usb/esp_tinyusb/include/tinyusb.h
+++ b/usb/esp_tinyusb/include/tinyusb.h
@@ -34,6 +34,10 @@ typedef struct {
     const uint8_t *configuration_descriptor;   /*!< Pointer to a configuration descriptor. If set to NULL, TinyUSB device will use a default configuration descriptor whose values are set in Kconfig */
     bool self_powered;                         /*!< This is a self-powered USB device. USB VBUS must be monitored. */
     int vbus_monitor_io;                       /*!< GPIO for VBUS monitoring. Ignored if not self_powered. */
+    bool no_background_task;                   /*!< When set to false, background task handling tinyusb events is created. */
+    size_t task_priority;                      /*!< Task priority of created background tinyusb task */
+    size_t stack_size;                         /*!< Stack size of crated background tinyusb task */
+    BaseType_t core_id;                        /*!< Select core on which background tinyusb task will run or tskNO_AFFINITY  */
 } tinyusb_config_t;
 
 /**
--- a/usb/esp_tinyusb/tinyusb.c
+++ b/usb/esp_tinyusb/tinyusb.c
@@ -25,6 +25,9 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
     const tusb_desc_device_t *dev_descriptor;
     const char **string_descriptor;
     int string_descriptor_count = 0;
+    size_t task_priority = 0;
+    size_t stack_size = 0;
+    BaseType_t core_id = 0;
     const uint8_t *cfg_descriptor;
     ESP_RETURN_ON_FALSE(config, ESP_ERR_INVALID_ARG, TAG, "Config can't be NULL");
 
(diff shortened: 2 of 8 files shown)


esp-saurabhbansal:
@tore-espressif @roma-jam @Dazza0 
Please check this change.

tore-espressif:
@esp-saurabhbansal Thank you for the PR, I was thinking about making all the task related configs (stack size, affinity, priority) run-time configurable by extending `tinyusb_config_t`. With backward compatibility.

What do you think?

usb/esp_tinyusb/tusb_tasks.c
```diff
@@ -12,6 +12,14 @@
 #include "tinyusb.h"
 #include "tusb_tasks.h"
 
+#if !CONFIG_TINYUSB_NO_DEFAULT_TASK
```
roma-jam, line 15: Nitpick: it is sometimes hard to understand what is the goal when there are two or more denial in logic. 
I mean, here we have (NOT)*_NO_DEFAULT. 
Probably, there is a chance to change it to #if CONFIG_TINYUSB_DEFAULT_TASK.

esp-saurabhbansal: Thanks @roma-jam for the comment.
All the CONFIG options for background tinyusb task are removed now. Instead, runtime parameters are used.

esp-saurabhbansal:
> @esp-saurabhbansal Also, could you please update this line https://github.com/espressif/idf-extra-components/blob/master/usb/esp_tinyusb/CMakeLists.txt#L49 like this:
> 
> ```
> # Determine whether tinyusb is fetched from component registry or from local path
> idf_build_get_property(build_components BUILD_COMPONENTS)
> if(tinyusb IN_LIST build_components)
>     set(tinyusb_name tinyusb) # Local component
> else()
>     set(tinyusb_name espressif__tinyusb) # Managed component
> endif()
> 
> # Pass tusb_config.h from this component to TinyUSB
> idf_component_get_property(tusb_lib ${tinyusb_name} COMPONENT_LIB)
> ```

This is done.

esp-saurabhbansal:
> @esp-saurabhbansal Thank you for the PR, I was thinking about making all the task related configs (stack size, affinity, priority) run-time configurable by extending `tinyusb_config_t`. With backward compatibility.
> 
> What do you think?

Done

usb/esp_tinyusb/include/tinyusb.h
```diff
@@ -34,6 +34,10 @@ typedef struct {
     const uint8_t *configuration_descriptor;   /*!< Pointer to a configuration descriptor. If set to NULL, TinyUSB device will use a default configuration descriptor whose values are set in Kconfig */
     bool self_powered;                         /*!< This is a self-powered USB device. USB VBUS must be monitored. */
     int vbus_monitor_io;                       /*!< GPIO for VBUS monitoring. Ignored if not self_powered. */
+    bool no_backround_task;                    /*!< When set to false, background task handling tinyusb events is created. */
```
tore-espressif, line 37: type
```suggestion
    bool no_background_task;                    /*!< When set to false, background task handling tinyusb events is created. */
```

usb/esp_tinyusb/tinyusb.c
```diff
@@ -92,9 +95,24 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
...
+        if (config->task_priority != 0) {
+            task_priority = config->task_priority;
+        } else {
+            task_priority = 5; // Backward compatibility with esp_tinyusb v1.0.0. Do NOT remove!
+        }
+        if (config->core_id >= 0 && (config->core_id <= configNUM_CORES - 1)) {
+            core_id = config->core_id;
+        } else {
+            core_id = tskNO_AFFINITY; // Backward compatibility with esp_tinyusb v1.0.0. Do NOT remove!
+        }
```
tore-espressif, line 113: Actually, we want to pin the task to core 0 by default. There are issues when the task is not pinned, so before we fix it, the task should be pinned be default.

usb/esp_tinyusb/include/tinyusb.h
```diff
@@ -34,6 +34,10 @@ typedef struct {
...
+    uint32_t task_priority;                    /*!< Task priority of created background tinyusb task */
+    UBaseType_t stack_size;                    /*!< Stack size of crated background tinyusb task */
+    BaseType_t core_id;                        /*!< Select core on which background tinyusb task will run or tskNO_AFFINITY  */
```
tore-espressif, line 40: In most of our components we use type `size_t`. This way, we don't have to include freertos headers in this file

usb/esp_tinyusb/tinyusb.c
esp-saurabhbansal: By default, config->core_id is 0 and task will be pinned to core 0. So, the implementation is correct. Only line to change is:
`core_id = tskNO_AFFINITY; // Backward compatibility with esp_tinyusb v1.0.0. Do NOT remove!`
->
`core_id = tskNO_AFFINITY;`

is it OK ?

usb/esp_tinyusb/include/tinyusb.h
esp-saurabhbansal: done

usb/esp_tinyusb/include/tinyusb.h
esp-saurabhbansal: I have changed the type of `task_priority` and `stack_size` to `size_t`
For core_id, I have kept it `BaseType_t `

tore-espressif:
@esp-saurabhbansal Thank you for the changes. But we will delay this breaking change until next major refactoring

PR #179, opened 2023-05-18
esp_tinyusb: cdc console feature is enabled only when VFS is enabled

esp-saurabhbansal:
cdc console feature is enabled only when VFS is enabled.
This is done by compiling the console feature related files only when VFS option is enabled.
If VFS is not enabled, compile error will be thrown. So, it will avoid runtime failure.

Closes https://github.com/espressif/idf-extra-components/issues/164

### DIFF
--- a/usb/esp_tinyusb/CMakeLists.txt
+++ b/usb/esp_tinyusb/CMakeLists.txt
@@ -27,9 +27,13 @@ if(CONFIG_TINYUSB_CDC_ENABLED)
     list(APPEND srcs
         "cdc.c"
         "tusb_cdc_acm.c"
+        )
+if(CONFIG_VFS_SUPPORT_IO)
+    list(APPEND srcs
         "tusb_console.c"
         "vfs_tinyusb.c"
         )
+endif() # CONFIG_VFS_SUPPORT_IO
 endif() # CONFIG_TINYUSB_CDC_ENABLED
 
 if(CONFIG_TINYUSB_MSC_ENABLED)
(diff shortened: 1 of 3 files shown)


esp-saurabhbansal:
@tore-espressif @roma-jam @Dazza0 
Please check this change.

tore-espressif:
@esp-saurabhbansal the changes LGTM. 
Could you please change the commit message prefix from 'usbd' to 'esp_tinyusb' and add a second line to the commit message `Closes https://github.com/espressif/idf-extra-components/issues/164` ?

Could we add this commit to #178 and merge the changes together? This way we won't release 2 new versions (1.2.2 and 1.2.3) but only 1 new version (1.2.2)

usb/esp_tinyusb/CMakeLists.txt
```diff
@@ -27,9 +27,13 @@ if(CONFIG_TINYUSB_CDC_ENABLED)
     list(APPEND srcs
         "cdc.c"
         "tusb_cdc_acm.c"
+        )
+if(CONFIG_VFS_SUPPORT_IO)
```
roma-jam, line 31: Nitpick: Probably, CONFIG_VFS_SUPPORT_IO_STREAMS would be little bit precisely. 
But it's up to you, indeed.

tore-espressif: This is coming from VFS component https://github.com/espressif/esp-idf/blob/master/components/vfs/Kconfig#L3

So we cannot change this :(

esp-saurabhbansal: CONFIG_VFS_SUPPORT_IO is a TOP level config option for VFS feature.

roma-jam: OK, I see.
I am not insisting. :slightly_smiling_face:

esp-saurabhbansal:
> @esp-saurabhbansal the changes LGTM. Could you please change the commit message prefix from 'usbd' to 'esp_tinyusb' and add a second line to the commit message `Closes https://github.com/espressif/idf-extra-components/issues/164` ?
> 
> Could we add this commit to #178 and merge the changes together? This way we won't release 2 new versions (1.2.2 and 1.2.3) but only 1 new version (1.2.2)

This commit is added to PR #178

tore-espressif (approved):

PR #180, opened 2023-05-22
Port IQmath Library

suda-morris:
This library was obtained from TI's MSPM0 SDK, licensed under BSD-3-Clause:

image

As a fixed-point mathematical library, it should benefit Espressif's RISC-V chips which don't have FPU embedded. 

We're able to use this library to facilitate the FOC control applications on e.g. ESP32-C6 (FYI, ESP32-C6 does have the MCPWM peripheral).

A simple example was also added to test its functionality.


tore-espressif (approved):
LGTM

tore-espressif (approved):
LGTM

PR #154, opened 2023-02-24
esp_tinyusb: add USB device class NCM (IEC-17)

Jacques-Zhao:
Added support for NCM class, and the corresponding example can be found in IDF MR: [22232](https://gitlab.espressif.cn:6688/espressif/esp-idf/-/merge_requests/22232)

PR #185, opened 2023-05-29
esp_tinyusb: USB device Network class (NCM)

david-cermak:
Reworked https://github.com/espressif/idf-extra-components/pull/154

### DIFF
--- a/usb/esp_tinyusb/include/tinyusb_net.h
+++ b/usb/esp_tinyusb/include/tinyusb_net.h
@@ -1,0 +1,53 @@
+/*
+ * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <stdint.h>
+#include "tinyusb_types.h"
+#include "esp_err.h"
+#include "sdkconfig.h"
+
+#if (CONFIG_TINYUSB_NET_MODE_NONE != 1)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef void (*net_recv_handler_t)(void *buffer, uint16_t len);
+
+/**
+ * @brief ESP TinyUSB NCM driver configuration structure
+ */
+typedef struct {
+    const uint8_t *mac_addr; /*!< MAC address. Must be 6 bytes long. */
+    net_recv_handler_t recv_handle; /*!< TinyUSB receive data handle */
+} tinyusb_net_config_t;
+
+/**
+ * @brief Initialize TinyUSB NET driver
+ *
+ * @param[in] usb_dev USB device to use
+ * @param[in] cfg     Configuration of the driver
+ * @return esp_err_t
+ */
+esp_err_t tinyusb_net_init(tinyusb_usbdev_t usb_dev, const tinyusb_net_config_t *cfg);
+
+/**
+ * @brief TinyUSB NET driver send data
+ *
+ * @param[in] buffer USB send data
+ * @param[in] len     Send data len
+ * @return esp_err_t
+ */
+esp_err_t tinyusb_net_send(void *buffer, uint16_t len);
+
+#endif
+
+//TODO: Define ESP-specific functions
+#ifdef __cplusplus
+}
+#endif
--- a/usb/esp_tinyusb/include/tinyusb_net.h
+++ b/usb/esp_tinyusb/include/tinyusb_net.h
@@ -17,14 +17,36 @@
 extern "C" {
 #endif
 
-typedef void (*net_recv_handler_t)(void *buffer, uint16_t len);
+/**
+ * @brief On receive callback type
+ */
+typedef esp_err_t (*tusb_net_rx_cb_t)(void *buffer, uint16_t len, void *ctx);
+
+/**
+ * @brief Free Tx buffer callback type
+ */
+typedef void (*tusb_net_free_tx_cb_t)(void *buffer, void *ctx);
+
+/**
+ * @brief On init callback type
+ */
+typedef void (*tusb_net_init_cb_t)(void *ctx);
 
 /**
  * @brief ESP TinyUSB NCM driver configuration structure
  */
 typedef struct {
-    const uint8_t *mac_addr; /*!< MAC address. Must be 6 bytes long. */
-    net_recv_handler_t recv_handle; /*!< TinyUSB receive data handle */
+    uint8_t mac_addr[6];                      /*!< MAC address. Must be 6 bytes long. */
+    tusb_net_rx_cb_t on_recv_callback;        /*!< TinyUSB receive data callbeck */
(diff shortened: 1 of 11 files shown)


usb/esp_tinyusb/tinyusb_net.c
```diff
@@ -0,0 +1,173 @@
...
+
+static void do_send_sync(void *ctx)
+{
+    (void) ctx;
+    if (xSemaphoreTake(s_net_obj.buffer_sema, 0) != pdTRUE || s_net_obj.packet_to_send == NULL) {
+        return;
+    }
+
+    packet_t *packet = s_net_obj.packet_to_send;
+    if (tud_network_can_xmit(packet->len)) {
```
tore-espressif, line 47: Question: Did the function ever returned 'false' during your tests?

If yes, could we make the error handling a little more thorough?

usb/esp_tinyusb/tinyusb_net.c
```diff
@@ -0,0 +1,173 @@
...
+        packet->result = ESP_FAIL;
+    }
+    xSemaphoreGive(s_net_obj.buffer_sema);
+    xEventGroupSetBits(s_net_obj.tx_flags, TX_FINISHED_BIT);
+}
+
+static void do_send_async(void *ctx)
+{
+    packet_t *packet = ctx;
+    if (tud_network_can_xmit(packet->len)) {
```
tore-espressif, line 60: How can we signal failure of this call to the user?

tore-espressif (commented):
@david-cermak thank you very much for the update! I left few comments about error handling, apart from that, LGTM!

Could you also please bump version in idf_component.yml to 1.3.0 ?

usb/esp_tinyusb/tinyusb.c
```diff
@@ -61,12 +61,13 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
         cfg_descriptor = config->configuration_descriptor;
     } else {
         // Default configuration descriptor is provided only for CDC and MSC classes
-#if (CFG_TUD_HID > 0 || CFG_TUD_MIDI > 0 || CFG_TUD_CUSTOM_CLASS > 0 || CFG_TUD_ECM_RNDIS > 0 || CFG_TUD_NCM > 0 || CFG_TUD_DFU > 0 || CFG_TUD_DFU_RUNTIME > 0 || CFG_TUD_BTH > 0)
```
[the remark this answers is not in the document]
david-cermak: Thanks, removed only NCM class.

usb/esp_tinyusb/tinyusb_net.c
david-cermak: Yes, this happens if the USB cannot accept more packets (i.e. the available space in the outgoing buffer is less that size of this packet).

> could we make the error handling a little more thorough?

The error should be reported to the user API, as this is a synchronous API which waits for completion.

usb/esp_tinyusb/tinyusb_net.c
david-cermak: This is an asynchronous API, so the packet would be just dropped.
I've added an error message, but that's probably all we can do I think.

Maybe we could add Tx result argument to this `tx_buff_free()` callback, WYT?

usb/esp_tinyusb/include/tinyusb_net.h
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * @brief ESP TinyUSB NCM driver configuration structure
+ */
+typedef struct {
+    uint8_t mac_addr[6];                      /*!< MAC address. Must be 6 bytes long. */
+    tusb_net_rx_cb_t on_recv_callback;        /*!< TinyUSB receive data callbeck */
+    tusb_net_free_tx_cb_t free_tx_buffer;     /*!< User function for freeing the Tx buffer.
+                                               *    - could be NULL, if user app is responsible for freeing the buffer
+                                               *    - must be used in asynchronous send mode
+                                               *    - is only called if the packet is accepted, i.e. tinyusb_net_send...() returns ESP_OK */
```
tore-espressif, line 44: In async mode, this callback is called even when the packet is _not_ accepted, right?

usb/esp_tinyusb/tinyusb_net.c
tore-espressif: > Maybe we could add Tx result argument to this tx_buff_free() callback, WYT?

This sounds good to me. But if it is OK to drop packages silently, then even the current solution is OK.

Nitpick: would `tx_done_cb` be a better name for this callback?

david-cermak: > if it is OK to drop packages silently, then even the current solution is OK.

I'd prefer this option TBH. It's already described here, that the async mode should be used with consideration (and now with the error message, developers can easily adjust task priorities to reduce this kind of packet drop)

https://github.com/espressif/idf-extra-components/blob/d3a6806d194ce934871866f0f62c478c98e4e931/usb/esp_tinyusb/include/tinyusb_net.h#L83-L84

> Nitpick: would tx_done_cb be a better name for this callback?

`tx_done_cb()` wouldn't reflect the purpose of this callback IMO, as this is called exactly the moment when we *copy* the data from *our* buffer to the internal tinyUSB buffer. so the exact moment we're able to deallocate the buffer.
We could also add a similar callback for transmit completion (and maybe add a completion status), but I would probably like to keep this simple (adding a new callback isn't a breaking change and can be implemented later)

usb/esp_tinyusb/include/tinyusb_net.h
```diff
@@ -0,0 +1,99 @@
...
+ */
+typedef struct {
+    uint8_t mac_addr[6];                      /*!< MAC address. Must be 6 bytes long. */
+    tusb_net_rx_cb_t on_recv_callback;        /*!< TinyUSB receive data callbeck */
+    tusb_net_free_tx_cb_t free_tx_buffer;     /*!< User function for freeing the Tx buffer.
+                                               *    - could be NULL, if user app is responsible for freeing the buffer
+                                               *    - must be used in asynchronous send mode
+                                               *    - is only called if the used tinyusb_net_send...() function returns ESP_OK
+                                               *        - in sync mode means that the packet was accepted by TinyUSB
+                                               *        - in async mode means that the packet was queued to be processed in TinyUSB task
```
tore-espressif, line 46: I'm really sorry for nitpicking but: But we call the callback [here](https://github.com/espressif/idf-extra-components/blob/d3a6806d194ce934871866f0f62c478c98e4e931/usb/esp_tinyusb/tinyusb_net.c#L64) when the packed was _not_ queued

usb/esp_tinyusb/include/tinyusb_net.h
david-cermak: That's true, thanks for noticing. Have rephrased and added few details about sync vs async mode.

tore-espressif (approved):

tore-espressif:
@david-cermak Can we merge it or do you plan more changes?

david-cermak:
@tore-espressif Feel free to merge it if the last description sounds okay to you.

PR #183, opened 2023-05-24
ci: Use idf-build-apps from example builds

tore-espressif:
# Change description

This new GitHub action will build all examples in this repository.

TODO:
- [x] CBOR examples have their own GH action; merge it with this one
- [x] Add support for example tests
- [x] Wait until the 'exclude feature' is fixed in idf-build-apps _EDIT:_ Fixed in idf-build-apps v1.0.0

PR #187, opened 2023-06-06
esp_encrypted_img: fix structure member name in esp_decrypt_cfg_t

hmalpani:
1) Fix structure member name in esp_decrypt_cfg_t 
1.1 `rsa_pub_key` to `rsa_priv_key`
1.2 `rsa_pub_key_len` to `rsa_priv_key_len`
2) Fix the return code  in the header file API description

### DIFF
--- a/esp_encrypted_img/include/esp_encrypted_img.h
+++ b/esp_encrypted_img/include/esp_encrypted_img.h
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <esp_err.h>
+#include <esp_idf_version.h>
 
 #if 0   //High level layout for state machine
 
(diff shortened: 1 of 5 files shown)


hmalpani:
Before merging this, we need to update the example in idf. It will fail in CI as warning for deprecated members will appear while building the example.

esp_encrypted_img/include/esp_encrypted_img.h
```diff
@@ -45,8 +45,14 @@ extern "C" {
 typedef void *esp_decrypt_handle_t;
 
 typedef struct {
-    const char *rsa_pub_key;    /*!< 3072 bit RSA key in PEM format */
-    size_t rsa_pub_key_len;     /*!< Length of the buffer pointed to by rsa_pub_key*/
+    union {
+        const char *rsa_priv_key;                               /*!< 3072 bit RSA private key in PEM format */
+        const char *rsa_pub_key __attribute__((deprecated));    /*!< RSA private key buffer legacy name */
```
mahavirj, line 50: Lets not add attribute here, it will break the existing releases..

Suggest updating comment that _this name is kept for backward compatibility purpose, but it is not accurate (meaning wise) and hence it would be removed in the next major release_

.github/workflows/build_and_run_pre_encrypted_ota.yml
```diff
@@ -0,0 +1,74 @@
+name: Build and Run Pre Encrypted OTA Example
```
tore-espressif, line 1: All examples in this repository will be run after this PR https://github.com/espressif/idf-extra-components/pull/183 is merged. Could you wait & rebase after it is merged?

mahavirj:
@hmalpani Lets take the example addition separately and merge the current PR

mahavirj (approved):

PR #188, opened 2023-06-09
Add CoAP examples from ESP-IDF

hmalpani:
Added CoAP examples from IDF to idf-extra-components.


.github/workflows/build_and_run_coap_examples.yml
```diff
@@ -0,0 +1,71 @@
+name: Build and Run CoAP Examples
```
tore-espressif, line 1: All examples should now be build&run by this GH action https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/build_and_run_examples.yml

Could you please rebase your PR and check that it works OK with your new examples?

mahavirj: @tore-espressif This example runs on ethernet based runner (tag `ESP32-ETHERNET-KIT`). Is it possible to specify specific tag for an example in the `idf-build-apps` action?

hmalpani: Also I don't think it will build coap_server example as it does not have a pytest script. For `coap_server`, we just want to build it in CI and it does not need to run in CI.

tore-espressif: @hmalpani the CI script will build all examples and run example tests for examples that have pytest

@mahavirj I'll check the tags

tore-espressif: Hello @mahavirj , if you want to run some example's tests on specific runner, you need to do the following:
1. Define new env marker in pytest.ini (e.g. `ethernet`)
2. Add the marker to your pytests, e.g. `@pytest.mark.ethernet`
3. Define new CI step that [runs-on](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on) a selected runner (e.g. `runs-on: [self-hosted, linux, docker, "${{ matrix.idf_target }}", "some_other_marker"]`)
4. Run `pytest -m some_other_marker` to run only testcases with this marker

hmalpani: Hello @tore-espressif
I have updated `.build-test-rules.yml` file as follows: https://github.com/espressif/idf-extra-components/pull/188/files#diff-c2ae404d9e16bd33ade435e00a2de9a59245c24a72b8c24c4ef39d3ec9b59325R13-R19 so that `coap_client` and `coap_server` examples don't get built for esp32h2. But still, they are being built for esp32h2, see job: https://github.com/espressif/idf-extra-components/actions/runs/5343770280/jobs/9687302992. Am I missing something here?

mahavirj: @hmalpani Shouldn't the entry be `coap/examples/coap_client`?

hmalpani: @mahavirj Yes that worked. I thought we just add the example name.
Thanks!

.build-test-rules.yml
```diff
@@ -9,3 +9,11 @@ esp_delta_ota:
   disable:
     - if: SOC_WIFI_SUPPORTED != 1
       reason: Relevant only for WiFi enabled targets
+
+coap/examples/coap_server:
+  disable:
+    - if: IDF_TARGET in ["esp32h2"]
```
mahavirj, line 15: In IDF, we have these examples enabled for ESP32-H2 as well. Probably for H2, the default transport goes to SPI-ethernet and hence the examples should build?

.github/workflows/build_and_run_examples.yml
```diff
@@ -62,4 +62,28 @@ jobs:
           PIP_EXTRA_INDEX_URL: "https://dl.espressif.com/pypi/"
         run: pip install --only-binary cryptography pytest-embedded pytest-embedded-serial-esp pytest-embedded-idf
       - name: Run examples
-        run: pytest --target=${{ matrix.idf_target }} --build-dir=build_${{ matrix.idf_target }} --ignore=usb --ignore=test_app
+        run: pytest --target=${{ matrix.idf_target }} -m generic --build-dir=build_${{ matrix.idf_target }} --ignore=usb --ignore=test_app
+
+  run-coap_examples:
```
mahavirj, line 67: Could this be generic like run-ethernet-examples?

mahavirj (approved):

PR #193, opened 2023-06-27
Add http2_request example in sh2lib component

mahavirj:
This example depends on sh2lib abstraction layer for HTTP/2 protocol with TLS. Hence, its kept inside the sh2lib component itself.

Note: ESP-IDF version prior to 5.2 had this example as a part of it but it will be removed starting 5.2 release.

# Checklist

NA - Adding example inside an existing component

# Change description


sh2lib/examples/http2_request/pytest_http2_request.py
```diff
@@ -0,0 +1,56 @@
...
+    except Exception as msg:
+        logging.info('Exception occurred when connecting to {}: {}'.format(TEST_SERVER, msg))
+        return False
+    finally:
+        conn.close()
+
+
+@pytest.mark.esp32
+@pytest.mark.ethernet
+def test_examples_protocol_http2_request(dut: Dut) -> None:
```
hfudev, line 33: from idf_build_apps.constants import IDF_VERSION_MAJOR

@pytest.mark.skipif(IDF_VERSION_MAJOR < 4, reason='does not support old releases')

hmalpani: Or we can use `--collect-app-info` with `idf-build-apps` while building the examples and then use `app-info-filepattern` while running the examples. This is the way we do in ESP-IDF CI?

hfudev: Also need to modify the `conftest.py`, which will make it more complicated. I don't recommend following the ESP-IDF CI in this repo, that's too complex.

mahavirj: Thanks, added the change.

hmalpani (approved):

PR #196, opened 2023-06-28
esp_tinyusb: add handling for integer overflow in MSC device class

esp-saurabhbansal:
Error handling for integer overflow is added so as to return appropriate error when argument passed to APIs overflows. This type of handling should be done in underlying storage component like wear-levelling. Nevertheless, we can also handle the overflow to make the code more robust.

closes [IDF-7436](https://jira.espressif.com:8443/browse/IDF-7436)


esp-saurabhbansal:
@tore-espressif @roma-jam @igrr 
PTAL at this change.

tore-espressif (commented):
@esp-saurabhbansal Thank you for the update.

I can see that now we can catch all the arithmetic overflow problems. Did you have chance to contact the wear_leaveling developers to fix the problem on their side too?

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -170,6 +173,10 @@ static esp_err_t msc_storage_read_sector(uint32_t lba,
         void *dest)
 {
     assert(s_storage_handle);
+    ESP_RETURN_ON_FALSE(lba < UINT_MAX, ESP_ERR_INVALID_ARG, TAG, "Overflow lba");
```
roma-jam, line 176: Seems, that for the overflow verification it is better not to verify overflow for uint32_t value here (cause it barely can happened on 32-bit platform) but to take into consideration the following logic: 
- overflow can happened during increasing operation (sum or multiplication), so every operation should be in boundaries after completion. Overflow here would mean, that the result value will be less that both operands.
- for MSC there should be a total num sectors for the media device which we should return during the read capacity operation. Maybe it makes sense to compare boundaries with this value to be sure that all input request at least in the boundaries which device claims.

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -170,6 +173,10 @@ static esp_err_t msc_storage_read_sector(uint32_t lba,
...
+    ESP_RETURN_ON_FALSE(offset < UINT_MAX, ESP_ERR_INVALID_ARG, TAG, "Overflow offset");
+    ESP_RETURN_ON_FALSE(size < UINT_MAX, ESP_ERR_INVALID_ARG, TAG, "Overflow size");
```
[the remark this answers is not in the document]
esp-saurabhbansal: I have removed these checks. The overflow will be handles by `__builtin_umul_overflow()` and `__builtin_uadd_overflow ()`given in [doc](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html)

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: I have removed these checks. The overflow will be handles by `__builtin_umul_overflow()` and `__builtin_uadd_overflow ()`given in [doc](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html)

roma-jam (approved):
LGTM.

esp-saurabhbansal:
> @esp-saurabhbansal Thank you for the update.
> 
> I can see that now we can catch all the arithmetic overflow problems. Did you have chance to contact the wear_leaveling developers to fix the problem on their side too?

@tore-espressif Yes, I contacted the wear-levelling developers and probably they will be fixing it using [this](https://jira.espressif.com:8443/browse/IDF-7313) Jira.

tore-espressif:
@esp-saurabhbansal LGTM. Do you think we can postpone release of this change? As far as I know, there are no bug reports from users about it now.

The reason for this is that I would like to avoid releasing every small bugfix. I prefer adding more bugfixes that we have in our backlog to esp_tinyusb 1.3.1

esp-saurabhbansal:
> @esp-saurabhbansal LGTM. Do you think we can postpone release of this change? As far as I know, there are no bug reports from users about it now.
> 
> The reason for this is that I would like to avoid releasing every small bugfix. I prefer adding more bugfixes that we have in our backlog to esp_tinyusb 1.3.1

sure, lets add more fixes to 1.3.1

tore-espressif:
> sure, lets add more fixes to 1.3.1

OK, so we can merge this without the version bump in idf_component.yml

PR #190, opened 2023-06-25
led_strip: add SPI backend (IEC-24)

Kainarx:
The interface still needs to be modified, the current version is only for technical verification

# Checklist

- [ ] Component contains License
- [ ] Component contains README.md
- [ ] Component contains idf_component.yml file with `url` field defined
- [ ] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [ ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description
Add an SPI mode to control Led_Strip by `CONFIG_BLINK_LED_SPI`

PR #197, opened 2023-06-29
led_strip: fix version dependency

suda-morris:
Continue to the previous PR: https://github.com/espressif/idf-extra-components/pull/190

* Changed the dependency, now the led_strip still relies on the idf v5.0, instead of v5.1
* Updated the README to describe the new SPI backend usage and a diagram to help user choose which backend

### DIFF
--- a/.build-test-rules.yml
+++ b/.build-test-rules.yml
@@ -1,7 +1,19 @@
-led_strip:
+led_strip/examples/led_strip_rmt_ws2812:
   enable:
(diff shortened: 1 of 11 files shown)


led_strip/examples/led_strip_spi_ws2812/main/idf_component.yml
```diff
@@ -3,4 +3,3 @@ dependencies:
   espressif/led_strip:
     version: '^2.4'
     override_path: '../../../'
-  idf: ">=5.1"
```
suda-morris, line 6: this is still needed becase now led_strip acquires only esp-idf 5.0, but this example can only build for 5.1

PR #204, opened 2023-07-12
fix(usb/hid): HID Host Driver v.1.0.1

roma-jam:
# Checklist

- [X] _Optional:_ Component contains unit tests
- [X] CI passing

# Changelog
- Fixed a bug where configuring the driver with `create_background_task = false` did not properly initialize the driver. This lead to `the hid_host_uninstall()` hang-up.
- Fixed bug where `hid_host_uninstall()` would cause a crash during the call while USB device has not been removed.
- Added `hid_host_get_device_info()` to get the basic information of a connected USB HID device.

### Brief description
Fix `hid_host_uninstall()` hanging during the closing while driver was installed with `create_background_task = false` configuration.
Unit test `class_specific_requests_with_external_polling` has been added to the test cases.

### Problem
When the `create_background_task = false` configuration is used, the `event_handler_task` is not being created. Thus, the `hid_host_handle_events()` should be called periodically to handle the call `usb_host_client_handle_events()` outside the HID driver.
After `hid_host_install()` with `create_background_task = false` the `usb_host_client_deregister()` will never be called and the semaphore `all_events_handled` will remain taken. 

### Solution
Adding the `hid_host_all_events_was_handled()` to deregister the usb client and unlock the semaphore.

### DIFF
--- a/usb/usb_host_hid/hid_host.c
+++ b/usb/usb_host_hid/hid_host.c
@@ -111,6 +111,7 @@ typedef struct {
     usb_host_client_handle_t client_handle;                     /**< Client task handle */
     hid_host_driver_event_cb_t user_cb;                         /**< User application callback */
     void *user_arg;                                             /**< User application callback args */
+    bool event_handling_started;                                /**< Events handler started flag */
     SemaphoreHandle_t all_events_handled;                       /**< Events handler semaphore */
     volatile bool end_client_event_handling;                    /**< Client event handling flag */
 } hid_driver_t;
@@ -139,6 +140,9 @@ typedef struct hid_class_request {
     uint8_t *data;                  /**< Pointer to data */
 } hid_class_request_t;
 
+
+// ----------------- USB Event Handler - Internal Task -------------------------
+
 /**
  * @brief USB Event handler
  *
@@ -148,19 +152,10 @@ typedef struct hid_class_request {
  */
 static void event_handler_task(void *arg)
 {
-    while (1) {
-        /* Here wee need a timeout 50 ms to handle end_client_event_handling flag
-         * during situation when all devices were removed and it is time to remove
-         * and destroy everything.
-         */
-        usb_host_client_handle_events(s_hid_driver->client_handle, portMAX_DELAY);
-
-        if (s_hid_driver->end_client_event_handling) {
-            break;
-        }
+    ESP_LOGD(TAG, "USB HID handling start");
+    while (hid_host_handle_events(portMAX_DELAY) == ESP_OK) {
     }
-    ESP_ERROR_CHECK( usb_host_client_deregister(s_hid_driver->client_handle) );
-    xSemaphoreGive(s_hid_driver->all_events_handled);
+    ESP_LOGD(TAG, "USB HID handling stop");
     vTaskDelete(NULL);
 }
 
@@ -252,7 +247,7 @@ static hid_iface_t *get_iface_by_handle(hid_host_device_handle_t hid_dev_handle)
     hid_iface_t *hid_iface = (hid_iface_t *) hid_dev_handle;
 
     if (!is_interface_in_list(hid_iface)) {
-        ESP_LOGE(TAG, "HID device handle not found");
+        ESP_LOGE(TAG, "HID interface handle not found");
         return NULL;
     }
 
@@ -374,6 +369,22 @@ static esp_err_t hid_host_add_interface(hid_device_t *hid_device,
     STAILQ_INSERT_TAIL(&s_hid_driver->hid_ifaces_tailq, hid_iface, tailq_entry);
     HID_EXIT_CRITICAL();
 
+    return ESP_OK;
+}
+
+/**
+ * @brief Remove interface from a list
+ *
+ * Use only inside critical section
+ *
+ * @param[in] hid_iface    HID interface handle
+ * @return esp_err_t
+ */
+static esp_err_t _hid_host_remove_interface(hid_iface_t *hid_iface)
+{
+    hid_iface->state = HID_INTERFACE_STATE_NOT_INITIALIZED;
+    STAILQ_REMOVE(&s_hid_driver->hid_ifaces_tailq, hid_iface, hid_interface, tailq_entry);
+    free(hid_iface);
     return ESP_OK;
 }
 
@@ -502,60 +513,26 @@ static bool hid_host_device_init_attempt(uint8_t dev_addr)
  * @brief USB device was removed we need to shutdown HID Interface
  *
  * @param[in] hid_dev_handle    Handle of the HID devive to close
- * @param[in] force             To shutdown HID interface immediately
  * @return esp_err_t
  */
-static esp_err_t hid_host_interface_shutdown(hid_host_device_handle_t hid_dev_handle,
-        bool force)
+static esp_err_t hid_host_interface_shutdown(hid_host_device_handle_t hid_dev_handle)
 {
-    hid_iface_t *iface = get_iface_by_handle(hid_dev_handle);
+    hid_iface_t *hid_iface = get_iface_by_handle(hid_dev_handle);
 
-    HID_RETURN_ON_INVALID_ARG(iface);
+    HID_RETURN_ON_INVALID_ARG(hid_iface);
 
-    iface->state = HID_INTERFACE_STATE_WAIT_USER_DELETION;
-
-    // Not force shutdown and Interface was started by user
-    if (!force && iface->user_cb) {
-        hid_host_user_interface_callback(iface, HID_HOST_INTERFACE_EVENT_DISCONNECTED);
-        return ESP_OK;
-    }
-
-    // Interface has not been opened, delete it right away
-    return hid_host_device_close(iface);
-}
-
-/**
- * @brief Destroy a list of available interfaces in RAM
- *
- * @return esp_err_t
- */
-static esp_err_t hid_host_interface_list_destroy(void)
-{
-    bool active_shutdown = false;
-
-    hid_iface_t *iface = STAILQ_FIRST(&s_hid_driver->hid_ifaces_tailq);
-    hid_iface_t *tmp = NULL;
-
-    while (iface != NULL) {
-        tmp = STAILQ_NEXT(iface, tailq_entry);
-        // free interface
-        if (HID_INTERFACE_STATE_NOT_INITIALIZED != iface->state) {
-            active_shutdown = true;
-            ESP_ERROR_CHECK( hid_host_device_close(iface) );
-            ESP_ERROR_CHECK( hid_host_interface_shutdown(iface, true));
-        }
-
+    if (hid_iface->user_cb) {
+        // Let user handle the remove process
+        hid_iface->state = HID_INTERFACE_STATE_WAIT_USER_DELETION;
+        hid_host_user_interface_callback(hid_iface, HID_HOST_INTERFACE_EVENT_DISCONNECTED);
+    } else {
+        // Remove HID Interface from the list right now
+        ESP_LOGD(TAG, "Remove addr %d, iface %d from list",
+                 hid_iface->dev_params.addr,
+                 hid_iface->dev_params.iface_num);
         HID_ENTER_CRITICAL();
-        iface->state = HID_INTERFACE_STATE_NOT_INITIALIZED;
-        STAILQ_REMOVE(&s_hid_driver->hid_ifaces_tailq, iface, hid_interface, tailq_entry);
-        free(iface);
+        _hid_host_remove_interface(hid_iface);
         HID_EXIT_CRITICAL();
-
-        iface = tmp;
-    }
-
-    if (active_shutdown) {
-        ESP_LOGE(TAG, "Shutdown active HID device.");
     }
 
     return ESP_OK;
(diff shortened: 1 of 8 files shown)


tore-espressif:
@roma-jam Thanks for the update, I think I understand the problem. I'd prefer if we could fix it without introducing new API. Could you please share your thoughts about this? :

1. Rename `s_hid_driver->end_client_event_handling` to `s_hid_driver->uninstall_request` (or something similar in meaning)
2. Extend `hid_host_handle_events()` with something like this:

``` c
esp_err_t hid_host_handle_events(uint32_t timeout)
{
    ...
    usb_host_client_handle_events(s_hid_driver->client_handle, timeout); // @todo add error checking
    if (s_hid_driver->uninstall_request) {
        usb_host_client_deregister(s_hid_driver->client_handle);
        xSemaphoreGive(s_hid_driver->all_events_handled); // @todo Is this really needed?
    }
    return ESP_OK;
}
```

The code snippet is pretty far from working, but the idea is clear: Could we handle the client deregistration in `hid_host_handle_events`? This way we do not need new API

roma-jam:
> Could we handle the client deregistration in `hid_host_handle_events`? This way we do not need new API

I thought about that, but I am not sure that there will be no collision with event handling and uninstall calling. 

Right now, all the events shoud be handled before uninstall, which is clear. As we are not able to handle the events after the uninstall completion. Which is also clear.

Moreover, there should be some api to set this `s_hid_driver->uninstall_request` and it cannot be done inside `hid_host_uninstall()`…

Otherwise, I am agreed with the idea of trying not to change api. 
I will think about it a little bit more.

roma-jam:
> There can be scenario where:
> 
> 1. `hid_host_install()` is called with `create_background_task = false`
> 2. User does clean up with `hid_host_uninstall()`
> 3. User does not call `hid_host_handle_events()` at all.
> 
> Who will give the semaphore then !!
> 
> `hid_host_uninstall()` will remain blocked on `xSemaphoreTake()`

Yes, indeed.
The point is that hid_host_uninstall() wait semaphore which handled only in background task. 
The next steps doesn't really matters, yes.

roma-jam:
Hey @tore-espressif, 

I have changed the logic and now the API is the same. 

> xSemaphoreGive(s_hid_driver->all_events_handled); // @todo Is this really needed?

Answering your question - Honestly, probably we do not need that. But in CDC Driver you do the same but with events handling (`CDC_ACM_TEARDOWN` and `CDC_ACM_TEARDOWN_COMPLETE`), so probably there is some idea behind that. 

@esp-saurabhbansal , 
Thanks for pointing that out. I have changed the logic during the `hid_host_uninstall()` function as well as have added the test case `class_specific_requests_with_external_polling_without_polling` when `hid_host_handle_events()` is not being called before uninstall the driver to verify the correct behavior in this scenario. 

Feel free to review changes, I am looking forward for yours feedback.
Thanks.

usb/usb_host_hid/hid_host.c
```diff
@@ -1254,13 +1277,29 @@ esp_err_t hid_host_device_close(hid_host_device_handle_t hid_dev_handle)
...
+
+    esp_err_t ret = usb_host_client_handle_events(s_hid_driver->client_handle, timeout);
+
+    if (s_hid_driver->end_client_event_handling) {
+        HID_RETURN_ON_ERROR( usb_host_client_deregister(s_hid_driver->client_handle),
+                             "HID Client deregister failure");
+
+        xSemaphoreGive(s_hid_driver->all_events_handled);
+        // wait for hid_host_uninstall completion
+        s_hid_driver->ext_task_config.external_task_hdl = xTaskGetCurrentTaskHandle();
```
esp-saurabhbansal, line 1298: `external_task_hdl` is set in function hid_host_handle_events() in one thread and then referred in function hid_host_uninstall() in another thread. These two threads can race against each other. I think, its better to set `external_task_hdl` at the very start if function hid_host_handle_events(). It will reduce the possibility race condition.

Something like this:

```
esp_err_t hid_host_uninstall(void)
{
    ***
    if (s_hid_driver->ext_task_config.active &&
        !s_hid_driver->ext_task_config.external_task_hdl) {
            ESP_LOGE(TAG, "HID Host external event handling has not been called");
            // Deregister client by ourselves and release semaphore
            HID_RETURN_ON_ERROR( usb_host_client_deregister(s_hid_driver->client_handle),
                                 "HID Client deregister failure");
            xSemaphoreGive(s_hid_driver->all_events_handled);
        }
    }

    xSemaphoreTake(s_hid_driver->all_events_handled, portMAX_DELAY);
    vSemaphoreDelete(s_hid_driver->all_events_handled);
    s_hid_driver->ext_task_config.external_task_hdl = NULL;
    s_hid_driver->ext_task_config.active = false;
    ***
}

esp_err_t hid_host_handle_events(uint32_t timeout)
{
    HID_RETURN_ON_FALSE(s_hid_driver,
                        ESP_ERR_INVALID_STATE,
                        "HID Driver is not installed");

    if (!s_hid_driver->ext_task_config.active) {
        return ESP_ERR_NOT_SUPPORTED;
    }
    s_hid_driver->ext_task_config.external_task_hdl = xTaskGetCurrentTaskHandle();

    esp_err_t ret = usb_host_client_handle_events(s_hid_driver->client_handle, timeout);

    if (s_hid_driver->end_client_event_handling) {
        HID_RETURN_ON_ERROR( usb_host_client_deregister(s_hid_driver->client_handle),
                             "HID Client deregister failure");

        xSemaphoreGive(s_hid_driver->all_events_handled);
    }

    return ret;
}
```

esp-saurabhbansal (commented):
Code logic is very good.
Can you check my suggestion once, please.

usb/usb_host_hid/hid_host.c
roma-jam: Hey @esp-saurabhbansal, 
Thanks for the review.

Probably, I don't get the point. Let me try to explain my logic, probably after that you can just point the moment when I am not right or has a logic misleading. 

1. When application logic calls the `hid_host_handle_events()`, we spend some time in `usb_host_client_handle_events()`. To make it compatible with power management features such as tickless idle and automatic light sleep we need to provide some mechanism with a possibility to use maximum delay `portMAX_DELAY` to wait next event and do not poll peripheral every N times per minute. This can be achieved when we call `hid_host_handle_events(portMAX_DELAY);` from the external polling task;
2. In this case, the host will wait another client event. If there is no event, the logic will return from this function after `usb_host_client_unblock()` during the `hid_host_uninstall()`;
3. Before client unblock, we change the state of internal flag `s_hid_driver->end_client_event_handling = true;` during uninstall;
4. When logic returns from `usb_host_client_handle_events()` in `hid_host_handle_events()` fulfilled in external task,  the  flag end_client_event_handling is true and we unblock the client, release semaphore and save the Task handle and suspend the external Task (here is another question, if there is more logic except the client events polling in this task, so we can halt this logic as well, which is not so good);

...

I see. 
Do you mean, that we can Resume the external Task before it (external Task) release semaphore and will go to Suspend, right?

esp-saurabhbansal: Lets call the thread as threadA which calls hid_host_uninstall() and threadB which calls hid_host_handle_events()

After Point 3, in case threadA keeps on executing and threadB does not get CPU, threadA will check for the if conditions and it finds external_task_hdl as NULL and it execute this block. It will call usb_host_client_deregister() -> xSemaphoreGive() -> ....
```
        if (NULL == s_hid_driver->ext_task_config.external_task_hdl) {
            ESP_LOGE(TAG, "HID Host external event handling has not been called");
            // Deregister client by ourselves and release semaphore
            HID_RETURN_ON_ERROR( usb_host_client_deregister(s_hid_driver->client_handle),
                                 "HID Client deregister failure");
            xSemaphoreGive(s_hid_driver->all_events_handled);
        }
```


Next, threadB gets the CPU and starts executing. After finding s_hid_driver->end_client_event_handling as true, it will start executing this block:
```
    if (s_hid_driver->end_client_event_handling) {
        HID_RETURN_ON_ERROR( usb_host_client_deregister(s_hid_driver->client_handle),
                             "HID Client deregister failure");

        xSemaphoreGive(s_hid_driver->all_events_handled);
        // wait for hid_host_uninstall completion
        s_hid_driver->ext_task_config.external_task_hdl = xTaskGetCurrentTaskHandle();
        vTaskSuspend(NULL); // Task will be resumed from hid_host_uninstall()
    }
```

This threaB will again call usb_host_client_deregister() -> xSemaphoreGive() -> sets external_task_hdl -> vTaskSuspend itself.

This flow happens because threadA find the value of external_task_hdl as NULL.
So, my sugession was to set external_task_hdl with xTaskGetCurrentTaskHandle(); before calling usb_host_client_handle_events().

I can be unclear now also. We can have a call to discuss.

usb/usb_host_hid/hid_host.c
```diff
@@ -377,6 +444,23 @@ static esp_err_t hid_host_add_interface(hid_device_t *hid_device,
...
+ *
+ * Use only inside critical section
+ *
+ * @param[in] hid_iface    HID interface handle
+ * @return esp_err_t
+ */
+static esp_err_t _hid_host_remove_interface(hid_iface_t *hid_iface)
+{
+
+    hid_iface->state = HID_INTERFACE_STATE_NOT_INITIALIZED;
```
Dazza0, line 458:
```suggestion
{
    hid_iface->state = HID_INTERFACE_STATE_NOT_INITIALIZED;
```

usb/usb_host_hid/include/usb/hid_host.h
```diff
@@ -16,6 +17,8 @@
 extern "C" {
 #endif
 
+#define HID_STR_DESC_SIZE           32
```
Dazza0, line 20: The max length of a string descriptor according to USB 2.0 specification is 256-2 bytes. Please add a description as to what this macro actually means.

```suggestion
#define HID_STR_DESC_MAX_SIZE           32
```

usb/usb_host_hid/hid_host.c
```diff
@@ -512,51 +596,53 @@ static esp_err_t hid_host_interface_shutdown(hid_host_device_handle_t hid_dev_ha
...
+            ESP_ERROR_CHECK( hid_host_device_close(interface) );
         }
 
+        ESP_ERROR_CHECK( hid_host_interface_shutdown(interface, true));
+
         HID_ENTER_CRITICAL();
-        iface->state = HID_INTERFACE_STATE_NOT_INITIALIZED;
-        STAILQ_REMOVE(&s_hid_driver->hid_ifaces_tailq, iface, hid_interface, tailq_entry);
-        free(iface);
+        _hid_host_remove_interface(interface);
```
Dazza0, line 631: There's a gap in the critical section here. Thus, during that gap, another thread could already have removed the interface from the `hid_ifaces_tailq`. What if we removed it from the list first and then carried out the shut down.

Dazza0 (commented):
@roma-jam Seems like we have a pretty complex set of calls to remove devices/interfaces on `hid_host_uninstall()`. As I've mentioned previously, would it be easier if we just required all opened interfaces/devices to be closed before allowing users to call `hid_host_uninstall()`.

Having `hid_host_uninstall()` trying to remove live devices/interfaces could lead to some tricky concurrency issues,

usb/usb_host_hid/hid_host.c
roma-jam: Yes, I know that. And still I didn't come up with any better idea how to fix it. 
There is a logical problem with architecture, that is why I decided to change it some time ago. 
I explained everything more detailed in the document, which I attached to the PR for next version: https://github.com/espressif/idf-extra-components/pull/214#issuecomment-1643770574

The whole idea to keep list of usb device opened device is strange. 
That is why there is a two-way choice during the hid_host_unsinstall(). 

But I'll try to figure something out to fix this version. 
Thanks.

roma-jam: Change the way how list is freed. 
Yes, there is still a gap in critical section, but in case if the some elements will be absent, should be no problem. 

I guess this is the best option to do in case of current logic.

usb/usb_host_hid/include/usb/hid_host.h
```diff
@@ -35,11 +48,27 @@ typedef enum {
     HID_HOST_INTERFACE_EVENT_DISCONNECTED,            /**< HID Device has been disconnected */
 } hid_host_interface_event_t;
 
+typedef struct {
+    wchar_t *p;
+    uint16_t len;
+} hid_host_dev_string_t;
```
tore-espressif, line 54: Unused type definition?

usb/usb_host_hid/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.0.0"
+version: "1.0.1~1"
```
tore-espressif, line 1: 1.0.1 should be enough, right?

tore-espressif (approved):
Thank you for your patience @roma-jam !
I left few nitpicks, otherwise LGTM!

And please make sure that the HID example in esp-idf works OK with this new version!

usb/usb_host_hid/include/usb/hid_host.h
roma-jam: Yep, right. 
Thanks.

usb/usb_host_hid/idf_component.yml
roma-jam: I think so.
But I have asked you about the description, because I understood that I don't really know when and what better to use. 

Anyway, changed to 1.0.1 right now.

tore-espressif: No problem! Here are some details about versioning https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/versioning.html#versioning-scheme

roma-jam: Thanks, I have updated that and verifyed with esp-idf as well. 
New example works well with both versions: 1.0.0 and 1.0.1.

PR #205, opened 2023-07-12
refactor(esp_tinyusb/cdc): Remove receive ring buffer

tore-espressif:
- CDC-ACM: Remove intermediate RX ringbuffer
- CDC-ACM: Increase default FIFO size to 512 bytes
- CDC-ACM: Fix Virtual File System binding


usb/esp_tinyusb/tusb_cdc_acm.c
```diff
@@ -164,16 +144,24 @@ esp_err_t tinyusb_cdcacm_register_callback(tinyusb_cdcacm_itf_t itf,
     if (acm) {
         switch (event_type) {
         case CDC_EVENT_RX:
```
roma-jam, line 146: **Refactoring offer** 
Seems that we have 4 equal callbacks and the same logic for them. 
Probably, it can be squashed to callback buffer `tusb_cdcacm_callback_t callback[CDC_EVENT_MAX]` and when we need to call it or change we can do it with a pointer by event_type index. 
This could help to simplify this switch-case structure to: 

    if (event_type < CDC_EVENT_MAX) {
        CDC_ACM_ENTER_CRITICAL();
        acm->callback[event_type] = callback;
        CDC_ACM_EXIT_CRITICAL();
    }
 
Also, the same with unregister callback function.

usb/esp_tinyusb/test/test_esp_tinyusb.c
```diff
@@ -0,0 +1,117 @@
...
+    TUD_CONFIG_DESCRIPTOR(1, 4, 0, cdc_desc_config_len, TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 100),
+    TUD_CDC_DESCRIPTOR(0, 4, 0x81, 8, 0x02, 0x82, 64),
+    TUD_CDC_DESCRIPTOR(2, 4, 0x83, 8, 0x04, 0x84, 64),
+};
+
+static void tinyusb_cdc_rx_callback(int itf, cdcacm_event_t *event)
+{
+}
+
+TEST_CASE("test_name", "[esp_tinyusb]")
```
roma-jam, line 53: IDK, but does this test case name by any chance has been missed during refactoring and it still has a `test_name` test case?

usb/test_app/pytest_usb_host.py
```diff
@@ -51,3 +54,56 @@ def test_usb_host(dut: Tuple[IdfDut, IdfDut]) -> None:
 
     # 3.4 Run HID tests
     host.run_all_single_board_cases(group='hid_host')
+
+@pytest.mark.esp32s2
+@pytest.mark.esp32s3
+@pytest.mark.usb_device
+def test_usb_device(dut) -> None:
```
roma-jam, line 61: Here it seems that we add tusb_device logic testing in file with `pytest_usb_host.py` name. 
Is it possible to create another *.py file for device logic testing? 
Or at least rename that one to something more common, like: `pytest_usb.py`?  :thinking:

roma-jam (approved):
Hey @tore-espressif well done! 

I have left some comments, which are not critical at all. Just a few suggestions. 
Also, I have tried to launch pytest for the tusb_device part and it seems that it is too difficult for me. 
As far as I understand I do not need to devices here to test the logic, or do I? 
Based on this, maybe it is a good idea to put somewhere the description how to launch a test locally. 

What do you think? Or we have such a description and there is only me who doesn't know about that?

Otherwise, LGTM.

usb/esp_tinyusb/tusb_cdc_acm.c
```diff
@@ -298,37 +254,27 @@ esp_err_t tinyusb_cdcacm_write_flush(tinyusb_cdcacm_itf_t itf, uint32_t timeout_
...
-                ESP_LOGW(TAG, "remained data to flush!");
-                return ESP_FAIL;
-            } else {
-                return ESP_OK;
-            }
-        }
+        // It might take some time until TinyUSB flushes the endpoint
+        // Since this call is non-blocking, we don't wait (or check)
+        // whether all data has been flushed
+        tud_cdc_n_write_flush(itf);
```
esp-saurabhbansal, line 260: Here, what will happen if there are more bytes remaining to be flushed !! 

Do we still need to return ESP_OK.

Suggestion:
```
    if (!timeout_ticks) { // if no timeout - nonblocking mode
        // It might take some time until TinyUSB flushes the endpoint
        // Since this call is non-blocking, we don't wait (or check)
        // whether all data has been flushed
        tud_cdc_n_write_flush(itf);

        if (tud_cdc_n_write_occupied(itf)) {
            ESP_LOGW(TAG, "remained data to flush!");
            return ESP_FAIL;
        }
    }
```

esp-saurabhbansal (commented):
@tore-espressif Please check few comments.

usb/esp_tinyusb/tusb_cdc_acm.c
tore-espressif: Yes, we return ESP_OK; that is one of the main reasons for this change. **The remaining data will be flushed automatically in next packet, after this packets is sent**.

Your suggestion would return `ESP_FAIL` even though all data is going to be flushed. To be more precise, your code returning `ESP_FAIL` signals, that the multi-packet transfer is still ongoing.

Does that make sense?

usb/esp_tinyusb/tusb_cdc_acm.c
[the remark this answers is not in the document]
tore-espressif: Thanks, will optimieze

usb/test_app/pytest_usb_host.py
tore-espressif: True, will improve the tests

tore-espressif (commented):
Thank you for the review @roma-jam @esp-saurabhbansal , I will update the code ASAP

usb/esp_tinyusb/tusb_cdc_acm.c
esp-saurabhbansal: Your are right.

Different perspective that I have is:
If timeout_ticks is 0, it means that the API is non-blocking. And all data has to be flushed right away. If data is remaining to be flushed, API should throw error.

tore-espressif: Agreed, I will update the function so it returns `ESP_ERR_NOT_FINISHED` in this case

esp-saurabhbansal (approved):

usb/esp_tinyusb/tusb_cdc_acm.c
tore-espressif: Hi Roman, if you agree, I'd leave this optimization for next round

tore-espressif:
@esp-saurabhbansal @roma-jam Thank you very much for your reviews. I fixed all the issues. Moreover, I update CMake so we don't require unnecessary components, PTAL!

usb/esp_tinyusb/tusb_cdc_acm.c
roma-jam: Hey @tore-espressif,

«Later » means « never » 😂.

But yes, no problem.

Issue #203, opened 2023-07-12
Incorrect Version Badge Link in README.md for "pid_ctrl" Folder (IEC-28)


WXL-steven:
### Answers checklist.

- [X] I have read the documentation of the component in question and the issue is not addressed there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

To be honest I'm not quite sure what the connection is between pcap and pid, but I don't think the version tag of [pid_ctrl](https://github.com/espressif/idf-extra-components/blob/master/pid_ctrl/README.md) should point to the [pcap repository](https://components.espressif.com/components/espressif/pcap).

![image](https://github.com/espressif/idf-extra-components/assets/76638078/cf3254cd-da29-49e7-a86a-900a0cc17737)

PR #207, opened 2023-07-13
feature/add reset_pre_error in pid_ctrl (IEC-30)

Kainarx:
There are occasions when it is necessary to reset the previously accumulated error (e.g. when the motor is stopped and then started again). This commit add a reset_pre_error function in pid_ctrl. And fix a hyperlink error in pid_ctrl README.md

Closes https://github.com/espressif/idf-extra-components/issues/203

# Checklist

- [ ] CI passing

# Change description

### DIFF
--- a/pid_ctrl/idf_component.yml
+++ b/pid_ctrl/idf_component.yml
@@ -1,4 +1,4 @@
-version: "0.1.1"
+version: "0.2.0"
 description: Proportional-integral-derivative controller
 url: https://github.com/espressif/idf-extra-components/tree/master/pid_ctrl
 dependencies:
--- a/pid_ctrl/include/pid_ctrl.h
+++ b/pid_ctrl/include/pid_ctrl.h
@@ -95,6 +95,16 @@ esp_err_t pid_update_parameters(pid_ctrl_block_handle_t pid, const pid_ctrl_para
  */
 esp_err_t pid_compute(pid_ctrl_block_handle_t pid, float input_error, float *ret_result);
 
+/**
+ * @brief Reset the accumulation in pid_ctrl_block
+ *
+ * @param[in] pid PID control block handle, created by `pid_new_control_block()`
+ * @return
+ *      - ESP_OK: Reset successfully
+ *      - ESP_ERR_INVALID_ARG: Reset failed because of invalid argument
+ */
+esp_err_t pid_reset_ctrl_block(pid_ctrl_block_handle_t pid);
+
 #ifdef __cplusplus
 }
 #endif
--- a/pid_ctrl/src/pid_ctrl.c
+++ b/pid_ctrl/src/pid_ctrl.c
@@ -138,3 +138,13 @@ esp_err_t pid_update_parameters(pid_ctrl_block_handle_t pid, const pid_ctrl_para
     }
     return ESP_OK;
 }
+
+esp_err_t pid_reset_ctrl_block(pid_ctrl_block_handle_t pid)
+{
+    ESP_RETURN_ON_FALSE(pid, ESP_ERR_INVALID_ARG, TAG, "invalid argument");
+    pid->integral_err = 0;
+    pid->previous_err1 = 0;
+    pid->previous_err2 = 0;
+    pid->last_output = 0;
+    return ESP_OK;
+}
--- a/pid_ctrl/README.md
+++ b/pid_ctrl/README.md
@@ -1,3 +1,3 @@
 # Proportional integral derivative controller
 
-[![Component Registry](https://components.espressif.com/components/espressif/pcap/badge.svg)](https://components.espressif.com/components/espressif/pcap)
+[![Component Registry](https://components.espressif.com/components/espressif/pid_ctrl/badge.svg)](https://components.espressif.com/components/espressif/pid_ctrl)


suda-morris:
@L-KAYA please take a look as well

suda-morris:
@Kainarx Please don't forget to bump the library version.

pid_ctrl/src/pid_ctrl.c
```diff
@@ -138,3 +138,13 @@ esp_err_t pid_update_parameters(pid_ctrl_block_handle_t pid, const pid_ctrl_para
...
+
+esp_err_t pid_reset_pre_error(pid_ctrl_block_handle_t pid)
+{
+    ESP_RETURN_ON_FALSE(pid, ESP_ERR_INVALID_ARG, TAG, "invalid argument");
+    pid->integral_err = 0;
+    pid->previous_err1 = 0;
+    pid->previous_err2 = 0;
+    pid->last_output = 0;
+    return ESP_OK;
+}
```
L-KAYA, line 150: If we need a reset API, I would suggest to make it more generic. In some cases we might only need to clear the integral error to do a fast response, and in another case we might want to reset the last output to eliminate the influence from the last output. Even in some unconventional apps, user wants to tamper with these data.

Therefore, I propose to have a generic update API like this:
```c
// Define a intermediate_data structure in the header
typedef struct {
    float integral_err;
    float previous_err1;
    float previous_err2;
    float last_output;
} pid_ctrl_intermediate_data_t;

// Realize it in the source file
esp_err_t pid_update_Intermediate data(pid_ctrl_block_handle_t pid, const pid_ctrl_intermediate_data_t *Intermediate_data) {
    ESP_RETURN_ON_FALSE(pid, ESP_ERR_INVALID_ARG, TAG, "invalid argument");
    pid->integral_err = Intermediate_data->integral_err;
    pid->previous_err1 = Intermediate_data->previous_err1;
    pid->previous_err2 = Intermediate_data->previous_err2;
    pid->last_output = Intermediate_data->last_output;
}
```
It's open for discussion~

And don't forget to have a new line in the end of this file.

Kainarx: thank you for the review,  I think generic update API is a good idea

pid_ctrl/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "0.1.1"
+version: "0.1.2"
```
suda-morris, line 1: it's a feature update, so please "0.2.0"

pid_ctrl/include/pid_ctrl.h
```diff
@@ -95,6 +106,17 @@ esp_err_t pid_update_parameters(pid_ctrl_block_handle_t pid, const pid_ctrl_para
...
+/**
+ * @brief Reset the pre_error
+ *
+ * @param[in] pid PID control block handle, created by `pid_new_control_block()`
+ * @param[in] Intermediate_data Intermediate data that need to update
+ * @return
+ *      - ESP_OK: Run a PID compute successfully
+ *      - ESP_ERR_INVALID_ARG: Run a PID compute failed because of invalid argument
+ */
+esp_err_t pid_update_Intermediate_data(pid_ctrl_block_handle_t pid, const pid_ctrl_intermediate_data_t *Intermediate_data);
```
suda-morris, line 118: Hmmm, do you think the user will know the current value of the internal "integral_err" when he only wants to reset the "previous_err" **only**? IMHO, the previous API looks better. "integral_err" doesn't have to be reset with "previous_err" together.

Kainarx: Typically, this interface is only called if the user wants to zero the cumulative error, but it can't be ruled out that some users will want to set the cumulative error to try out a different control effect, but I would think that this should be very rare, and perhaps this kind of functionality shouldn't exist in the component?

L-KAYA: How about split the previous `reset` API into `reset_previous_err` and `reset_integral_err`? 

And is it reasonable to reset the last output together with previous error?

L-KAYA: Oh, maybe reset `last_output` and `integral_err` together make more sense. Because `last_output` is the accumulated variable for `incremental` formula and `integral_err` is the accumulated variable for `positional` formula. And one pid control block is attached to one formula, reset them together is reasonable. Maybe naming it `reset_accumulation`/`clear_accumulation` or something.

Kainarx: the `last output` is for `PID_CAL_TYPE_INCREMENTAL` and the `integral_err` is for `PID_CAL_TYPE_POSITIONAL`. They are essentially the same thing, effective only in their respective control modes. There is no need to use multiple reset functions

L-KAYA: Yes, actually I mean two function like this:
```c
esp_err_t pid_reset_previous_err(pid_ctrl_block_handle_t pid)
{
    ESP_RETURN_ON_FALSE(pid, ESP_ERR_INVALID_ARG, TAG, "invalid argument");
    pid->previous_err1 = 0;
    pid->previous_err2 = 0;
    return ESP_OK;
}

esp_err_t pid_reset_accumulation(pid_ctrl_block_handle_t pid)
{
    ESP_RETURN_ON_FALSE(pid, ESP_ERR_INVALID_ARG, TAG, "invalid argument");
    pid->integral_err = 0;
    pid->last_output = 0;
    return ESP_OK;
}
```

I'm also OK with reseting all intermediate data in one function, but I think the function can be renamed to `pid_reset_ctrl_block` then.

L-KAYA:
Changes LGTM~ Thanks for the contribution!

L-KAYA (approved):

suda-morris (approved):

PR #210, opened 2023-07-14
esp_tinyusb: add MSC pre mount/unmount changed event and callback

esp-saurabhbansal:
Add additional pre mount/unmount changed event and callback.
This will be delivered (if subscribed) BEFORE mount/unmount operation is started.
This will be helpful to the user who wants to be notified before the mount/unmount operation starts.
One such requirement is mentioned [here](https://esp32.com/viewtopic.php?f=13&t=34535).

### DIFF
--- a/usb/esp_tinyusb/tusb_msc_storage.c
+++ b/usb/esp_tinyusb/tusb_msc_storage.c
@@ -39,6 +39,7 @@ typedef struct {
     esp_err_t (*read)(size_t sector_size, uint32_t lba, uint32_t offset, size_t size, void *dest);
     esp_err_t (*write)(size_t sector_size, size_t addr, uint32_t lba, uint32_t offset, size_t size, const void *src);
     tusb_msc_callback_t callback_mount_changed;
+    tusb_msc_callback_t callback_premount_changed;
 } tinyusb_msc_storage_handle_s; /*!< MSC object */
 
 /* handle of tinyusb driver connected to application */
@@ -248,11 +249,22 @@ fail:
 
 esp_err_t tinyusb_msc_storage_mount(const char *base_path)
 {
-    esp_err_t ret;
+    esp_err_t ret = ESP_OK;
     assert(s_storage_handle);
 
     if (s_storage_handle->is_fat_mounted) {
         return ESP_OK;
+    }
+
+    tusb_msc_callback_t cb = s_storage_handle->callback_premount_changed;
+    if (cb) {
+        tinyusb_msc_event_t event = {
+            .type = TINYUSB_MSC_EVENT_PREMOUNT_CHANGED,
+            .mount_changed_data = {
+                .is_mounted = s_storage_handle->is_fat_mounted
+            }
+        };
+        cb(&event);
     }
 
     if (!base_path) {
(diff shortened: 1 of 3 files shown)


usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -291,8 +306,21 @@ esp_err_t tinyusb_msc_storage_mount(const char *base_path)
         };
         cb(&event);
     }
+    uint32_t ticks_finish_cb_post = xTaskGetTickCount();
```
pacucha42, line 309: I'm not sure I understand this logic - the code seem like single-thread sequence of calls, so there's a chance it gets blocked by the CB user completely... isn't it more appropriate to run a separate thread to watch the timeout?

esp-saurabhbansal: I am not able to find out the timing limitations by which callback should be finished both in case of mount and unmount of storage medium.
In case of S3, I observed it to be 3500ms in case of unmount and 10000ms in case of mount. But it may be different for other chips.

So, for now I am skipping this debug timing part from this PR.
I will revisit it later after releasing the current customer requirement (i.e. pre-unmount callback).

esp-saurabhbansal:
@tore-espressif @igrr @pacucha42 PTAL

usb/esp_tinyusb/tusb_msc_storage.c
pacucha42: Ok, thanks for the update

usb/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -248,13 +249,24 @@ static esp_err_t _mount(char *drv, FATFS *fs)
...
     if (s_storage_handle->is_fat_mounted) {
         return ESP_OK;
     }
 
+    tusb_msc_callback_t cb = s_storage_handle->callback_premount_changed;
+    if (cb) {
+        tinyusb_msc_event_t event = {
+            .type = TINYUSB_MSC_EVENT_PREMOUNT_CHANGED,
+            .mount_changed_data = {
+                .is_mounted = s_storage_handle->is_fat_mounted
```
pacucha42, line 264: Nitpicking: if this code runs single threaded, 'is_fat_mounted' can only equal to FALSE here (similar situation few lines below). Please, consider either the code readability upgrade or adding some explanatory comments throughout whole component. It would surely help the users (incl. myself) with better understanding the logic. Not blocker now, though, can be done in follow-up PR. Thanks

pacucha42 (approved):
I believe the code was properly tested and does the job. However, please consider a bit of polishing and adding more comments in future follow-ups. Thank you

Approving

usb/esp_tinyusb/tusb_msc_storage.c
esp-saurabhbansal: `is_fat_mounted` can have two values TRUE and FALSE.
The pre-mount/unmount callback can be delivered to users from two places - one from `tinyusb_msc_storage_mount()` before mount operation where `is_fat_mounted` is always FALSE and other from `tinyusb_msc_storage_unmount()` before unmount operation where `is_fat_mounted` is always TRUE. This is well intended. User can figure out the place from where the callback is generated based on the value of `is_fat_mounted`.

Similar explanation for post-mount/unmount callback.

I will add some code comments for users/developers.

PR #212, opened 2023-07-18
Add SBOM files for applicable components

mahavirj:
Please see more details at:
https://github.com/espressif/esp-idf-sbom

# Checklist

- [x] CI passing

# Change description

### DIFF
--- a/expat/idf_component.yml
+++ b/expat/idf_component.yml
@@ -1,4 +1,5 @@
-version: "2.5.0"
+# Note: Please keep this version consistent with sbom.yml file
+version: "2.5.0~1"
 description: "Expat - XML Parsing C Library"
 url: https://github.com/espressif/idf-extra-components/tree/master/expat
 dependencies:
--- a/coap/idf_component.yml
+++ b/coap/idf_component.yml
@@ -1,4 +1,5 @@
-version: "4.3.1~3"
+# Note: Please keep this version consistent with sbom.yml file
+version: "4.3.1~4"
 description: Constrained Application Protocol (CoAP) C Library
 url: https://github.com/espressif/idf-extra-components/tree/master/coap
 dependencies:
--- a/coap/sbom.yml
+++ b/coap/sbom.yml
@@ -1,0 +1,5 @@
+version: 4.3.1
+cpe: cpe:2.3:a:libcoap:libcoap:{}:*:*:*:*:*:*:*
+supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
+originator: 'Organization: libcoap <https://libcoap.net/>'
+description: A CoAP (RFC 7252) implementation in C
--- a/expat/sbom.yml
+++ b/expat/sbom.yml
@@ -1,0 +1,5 @@
+version: 2.5.0
+cpe: cpe:2.3:a:libexpat_project:libexpat:{}:*:*:*:*:*:*:*
+supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
+originator: 'Organization: libexpat_project'
+description: Fast streaming XML parser written in C99
--- a/fmt/idf_component.yml
+++ b/fmt/idf_component.yml
@@ -1,4 +1,5 @@
-version: "9.1.0"
+# Note: Please keep this version consistent with sbom.yml file
+version: "9.1.0~1"
 description: Formatting library providing a fast and safe alternative to C stdio and C++ iostreams.
 url: https://github.com/espressif/idf-extra-components/tree/master/fmt
 dependencies:
--- a/fmt/sbom.yml
+++ b/fmt/sbom.yml
@@ -1,0 +1,5 @@
+version: 9.1.0
+cpe: cpe:2.3:a:fmt:fmt:{}:*:*:*:*:*:*:*
+supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
+originator: 'Organization: fmt <https://fmt.dev/latest/index.html>'
+description: A modern formatting library
--- a/libpng/idf_component.yml
+++ b/libpng/idf_component.yml
@@ -1,4 +1,5 @@
-version: "1.6.39"
+# Note: Please keep this version consistent with sbom.yml file
+version: "1.6.39~1"
 description: Portable Network Graphics(png) C library
 url: https://github.com/espressif/idf-extra-components/tree/master/libpng
 repository: "https://github.com/espressif/idf-extra-components.git"
--- a/libpng/sbom.yml
+++ b/libpng/sbom.yml
@@ -1,0 +1,5 @@
+version: 1.6.39
+cpe: cpe:2.3:a:libpng:libpng:{}:*:*:*:*:*:*:*
+supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
+originator: 'Organization: libpng'
+description: Portable Network Graphics support, official PNG reference library
--- a/nghttp/idf_component.yml
+++ b/nghttp/idf_component.yml
@@ -1,4 +1,5 @@
-version: "1.52.0"
+# Note: Please keep this version consistent with sbom.yml file
+version: "1.52.0~1"
 description: "nghttp2 - HTTP/2 C Library"
 url: https://github.com/espressif/idf-extra-components/tree/master/nghttp
 dependencies:
--- a/nghttp/sbom.yml
+++ b/nghttp/sbom.yml
@@ -1,0 +1,5 @@
+version: 1.52.0
+cpe: cpe:2.3:a:nghttp2:nghttp2:{}:*:*:*:*:*:*:*
+supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
+originator: 'Organization: nghttp2 <https://nghttp2.org/'
+description: nghttp2 - HTTP/2 C Library and tools
--- a/zlib/idf_component.yml
+++ b/zlib/idf_component.yml
@@ -1,4 +1,5 @@
-version: "1.2.13"
+# Note: Please keep this version consistent with sbom.yml file
+version: "1.2.13~1"
 description: zlib C library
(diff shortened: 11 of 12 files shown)


mahavirj:
@fhrbata PTAL

expat/idf_component.yml
```diff
@@ -1,4 +1,5 @@
-version: "2.5.0"
+# Note: Please keep this version consistent with sbom.yml file
+version: "2.5.0~1"
```
fhrbata, line 2: Hi @mahavirj , I'm a little bit confused and probably missing something. Why are we changing the versions in `idf_component.yml`? For example here from `2.5.0` to `2.5.0~1`.  Also just a note, if the version is not found in `sbom.yml` it will be take from `idf_component.yml` if presented. The same goes for description. So maybe we can avoid the duplication. Now I'm starting to think that maybe we should ask @kumekay if it would be possible to add the sbom fields support into `idf_component.yml` so we do not have to keep the info on two places for the managed components. Just a thought. Thank you.

mahavirj: Since we are adding a new file to this component, we must bump up its version so that it gets pushed to the upstream registry.

>  Also just a note, if the version is not found in sbom.yml it will be take from idf_component.yml if presented.

Okay, in that case, I can remove `version` field from the `sbom.yml` file. Probably we can remove description field as well?

> it would be possible to add the sbom fields support into idf_component.yml so we do not have to keep the info on two places for the managed components

That makes sense.

fhrbata: Thank you for the explanation @mahavirj . I will sync with @kumekay to see what can be done about this.

igrr: It would also make the implementation of automatic version updates (https://github.com/espressif/idf-extra-components/pull/174) a lot easier if there are less files which have to be modified for each version bump.

For now, https://github.com/espressif/github-actions/pull/40 has added support for updating the submodule itself and the version in idf_component.yml. Having to update also `sbom-version` and `sbom-hash` in .gitmodules, plus the new sbom.yml file, definitely makes the implementation more complex...

mahavirj: > do I understand it right that after the SBOM information is moved into sbom.yml, we will revert the .gitmodules changes

@igrr Yes, we can remove the .gitmodules change. In-fact, we can handle it in this PR itself.

Interim, the version in `idf_component.yml` may not always be same as the upstream. E.g., some components have version as `x.y.z~1` (highlights our own changes), upstream being `x.y.z`.  In that case, we may need an independent field to highlight version for SBOM or a separate `sbom.yml` file (like current PR).

fhrbata: @mahavirj , @igrr , is the `~x`, appended to a version to highlight our changes,  a convention that should be used by all managed components if needed or can there be other suffixes? Just me thinking out loud here: While generating SBOM, I guess we know that we are taking the version from `idf_component.yml` and we can strip the version suffix in the CPE value. Also there is no restriction to have a different version in the version field and CPE. Meaning we would have the version with suffix in the package version field and stripped in the CPE. This should allow correct vulnerability check with our checker. But this might be a problem e.g. with cve-bin-tool, which will use the version with suffix from the package version field. My understanding is that this should not be a problem if the CVE record has `CPE match ranges`, but it might be a problem for `CPE match string`. The later probably requires the exact version to match correctly.

kumekay: @fhrbata Starting v1.3 component manager is less strict about unknown root-level records, and only warning it printed. So it should be ok adding more data to the idf_component.yml now, and we can white-list more parameters (or somehow allow all the fields of sbom.yml)

fhrbata: Thank you very much @kumekay. I also liked the idea, which you have proposed, that there could be a section in `idf_component.yml` that would be whitelisted and ignored by component manager. That would allow us to add information in `idf_component.yml` without any restriction and in a layout we need. @mahavirj I think we should settle on this before we start adding data into `idf_component.yml`. 

Maybe something like this
```
dependencies:
  idf:
    version: '>=5.0'
description: Driver for Addressable LED Strip (WS2812, etc)
url: https://github.com/espressif/idf-extra-components/tree/master/led_strip
version: 2.4.1

external:
  sbom:
    cpe: cpe:2.3:a:someorg:blink:{}:*:*:*:*:*:*:*
    supplier: 'Person: John Doe'
    originator: 'Organization: Espressif Systems (Shanghai) CO LTD'

```
It's just an example. I guess other may have better ideas. Thank you

kumekay: @fhrbata I've checked the actual implementation and component manager v1.3 doesn't print any warnings for unknown root level fields, so you can use this solution now.

mahavirj: @fhrbata We need dedicated `version` field in the `sbom` section. As we discussed above, it could be different from the component version. Otherwise, I agree with these recommendation and can help to update this PR accordingly.

CC @igrr @dobairoland

fhrbata: Hi @mahavirj , I have no problem with the special sbom version field, but maybe we can just use different version directly in the cpe. Question is what's more convenient. I personally have no strong opinion on this. Example which should work even now
```
version: 1.6.39
cpe: cpe:2.3:a:libpng:libpng:1.1.1:*:*:*:*:*:*:*
supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
originator: 'Organization: libpng'
description: Portable Network Graphics support, official PNG reference library
``` 
But this will not work for 3rd party scanning tools, so maybe we really need special version field for sbom. Maybe we can just prefix all sbom fields to make it clear as we do in .gitmodules with `sbom_`. Or have a special sbom namespace in `idf_component.yml`.
```
version: 1.6.39
description: Portable Network Graphics support, official PNG reference library
sbom:
   version: 1.1.1
   supplier: 'Organization: Espressif Systems (Shanghai) CO LTD'
   originator: 'Organization: libpng'
```
I think this is just a matter of preferences. Thank you for all the help and ideas!

fhrbata: @mahavirj , @kumekay I quickly tried, with idf-component-manager 1.3.2, some local test managed component and I can confirm that the root variables are used and not reported. With e.g. `1.2.3` I get this
``` 
Invalid manifest format
Unknown keys: cpe, originator, supplier
SUGGESTION: This component may be using a newer version of the component manager.
```
I just wanted to point this out, because if we update `idf_component.yml` files, people with older manager will run into this. I'm not sure if this is(could be) a problem.

mahavirj: > have a special sbom namespace

This could be an override switch as well. If the version is present under sbom namespace then pick that one, else use the component version field itself.

igrr: > This could be an override switch as well. If the version is present under sbom namespace then pick that one, else use the component version field itself.

Don't we actually have two libraries, from SBOM point of view, here?
```mermaid
flowchart TD
    app["Application"]
    espressif/libpng["espressif/libpng\n1.6.39~2"]
    libpng["libpng\n1.6.39"]
    app -- depends --> espressif/libpng
    espressif/libpng -- depends --> libpng
```

And in some cases, there could be more than one library inside a component. For example, `console` component in IDF:

```mermaid
flowchart TD
    app["Application"]
    console["$IDF_PATH/console"]
    argtable3["argtable3\n3.2.2"]
    linenoise["linenoise\n1.0.1"]
    app -- depends --> console
    console -- depends --> argtable3
    console -- depends --> linenoise
```

Not saying we should handle such cases in this MR, just thought it's relevant to the discussion.

fhrbata: Hi @igrr , thank you for the graphs!

IIUC the problem we are trying to solve is that even though we have the sbom information for submodules in `.gitmodules`, it's of course not propagated through component manager. So if I do 

`idf.py add-dependency "espressif/libpng^1.6.39"`

the libpng component added to my project has no info that the `libpng` subdir in `managed_components/espressif__libpng/` is actually a submodule in the `idf-extra-components` repo with sbom info for it in `.gitmodules`.

I agree with your point above, that from the sbom POV the libpng component should depend on libpng library, but this relationship is currently not possible to express in the tool :(. It works like presented in your graph  if we have submodule entry for the libpng library available in `.gitmodules` and `.gitmodules` available while the sbom is generated.

For the first `libpng`  component example, I think we have to extend the sbom.yml to allow to express this relationship. Maybe something like this added in `idf_component.yml`

```yaml
sbom:
  contains:
    - path: libpng
      name: libpng
      version: 1.6.39
      cpe: cpe:2.3:a:libpng:libpng:{}:*:*:*:*:*:*:*
      supplier: 'Organization: libpng'
      description: Portable Network Graphics support, official PNG reference library
```
This would actually replace the sbom data and component->submodule relationship from `.gitmodules` .

For the second `console` component example, I think we can use the same approach as for `libpng`. We would just add two libs: `argtable3` and `linenoise`. Also IIUC  `argtable3` and `linenoise` are actually not submodules, so maybe we can added their own sbom.yml into their dirs.
```
components/console/
├── argtable3
│   └── sbom.yml
├── linenoise
│   └── sbom.yml
└── sbom.yml
```

Anyway both approaches require changes in the tool. I'm open and grateful for any ideas WRT this. Thank you very much!

mahavirj: @fhrbata Proposal sounds good to me. Please let me know once you have tool side changes in place, alternatively please feel free to take over this PR. Thanks.

fhrbata:
Hi @mahavirj ,

Here https://github.com/espressif/idf-extra-components/pull/239 is a PR, which adds the sbom information using referenced manifest files. It a draft, because we also need to change the checking in CI and probably remove the sbom information from `.gitmodules`. I opened MR for `esp-idf-sbom`, which should allow to validate(check the hashes) recorded in manifest files, not only in `.gitmodules`.

Thank you

mahavirj:
Closing in favor or https://github.com/espressif/idf-extra-components/pull/239

PR #227, opened 2023-08-21
Bugfix/update freetype license

suda-morris:
The first commit it attempt to fix the upload failure: https://github.com/espressif/idf-extra-components/actions/runs/5910069219/job/16031424944

The second commit is a small fix to declare the driver component as a public dependency of led_strip component

### DIFF
--- a/freetype/LICENSE
+++ b/freetype/LICENSE
@@ -1,169 +1,46 @@
-                    The FreeType Project LICENSE
-                    ----------------------------
-
-                            2006-Jan-27
-
-                    Copyright 1996-2002, 2006 by
-          David Turner, Robert Wilhelm, and Werner Lemberg
-
-
-
-Introduction
-============
-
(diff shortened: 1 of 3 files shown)


freetype/LICENSE
```diff
@@ -1,169 +1,46 @@
...
-  We  disclaim  all warranties  covering  The  FreeType Project  and
-  assume no liability related to The FreeType Project.
-
-
-  Finally,  many  people  asked  us  for  a  preferred  form  for  a
-  credit/disclaimer to use in compliance with this license.  We thus
-  encourage you to use the following text:
-
-   """
-    Portions of this software are copyright �0�8 <year> The FreeType
```
suda-morris, line 51: @igrr this might be related to the upload failure "'utf-8' codec can't decode byte 0x81 in position 2038: invalid start byte"

Issue #202, opened 2023-07-11
usb_host_msc: 1.1.0 can't use with psram open (IEC-27)


lijunru-hub:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

usb

### ESP-IDF version.

release/master, release/5.0

### Development Kit.

esp32-s3-n8r8

### Used Component version.

usb_host_msc: 1.1.0

### More Information.

Using version 1.1.0 of usb_host_msc and enabling PSRAM (OCT) simultaneously in ESP-IDF results in a failure to mount USB flash drives. This issue can be reproduced in both the release/master and release/v5.0 branches, using the MSC example from the ESP-IDF. However, this problem is not present in version 1.0.4.

```
         iManufacturer:  USB 
         iSerialNumber: 0501ed72cb11563ad6431daea2a5e4f 
E (10530) USB_MSC: msc_bulk_transfer_zcpy(599): 
E (10533) USB_MSC_SCSI: bot_execute_command(253): 
E (10538) diskio_usb: scsi_cmd_read10 failed (-1)
fresult 1
ESP_ERROR_CHECK failed: esp_err_t 0x1701 (ERROR) at 0x42009172
0x42009172: msc_init at /home/esp/project/usb_msc_example/image_display_2/build/../main/app_msc.c:193 (discriminator 1)

file: "../main/app_msc.c" line 193
func: msc_init
expression: msc_host_vfs_register(msc_device, "/usb", &mount_config, &vfs_handle)

abort() was called at PC 0x4037ae2f on core 0
0x4037ae2f: _esp_error_check_failed at /home/esp/esp-idf5/components/esp_system/esp_err.c:47
```

lijunru-hub:
![image](https://github.com/espressif/idf-extra-components/assets/68437239/81c8cb84-e117-484a-9984-e766c6c2fb0b)

igrr:
The issue is that `scsi_cmd_read10` and `scsi_cmd_write10` may be called with PSRAM buffers as arguments, however `msc_bulk_transfer_zcpy` now requires that the buffer is DMA capable:
[...]

PR #232, opened 2023-09-09
USB Host MSC v1.1.1

tore-espressif:
# Change description

* Correctly handle empty string descriptors
* Move data buffer to DMA capable memory, if non-capable memory is passed to the driver

PTAL @Dazza0 @roma-jam 
cc @leeebo @lijunru-hub @Maldus512 @0727fqh @f1andrew

Closes https://github.com/espressif/esp-idf/issues/12163
Closes https://github.com/espressif/idf-extra-components/issues/202
Closes https://github.com/espressif/esp-idf/issues/12142

### DIFF
--- a/usb/usb_host_msc/include/usb/msc_host.h
+++ b/usb/usb_host_msc/include/usb/msc_host.h
@@ -142,10 +142,17 @@ __attribute__((deprecated("use API from esp_private/msc_scsi_bot.h")));
 /**
  * @brief Handle MSC HOST events.
  *
- * @param[in]  timeout_ms  Timeout in miliseconds
- * @return esp_err_t
+ * If MSC Host install was called with create_background_task=false configuration,
+ * application needs to handle USB Host events itself.
+ * Do not call this function if MSC host install was called with create_background_task=true configuration
+ *
+ * @param[in]  timeout  Timeout in FreeRTOS tick
+ * @return
+ *     - ESP_OK:          All events handled
+ *     - ESP_ERR_TIMEOUT: No events handled within the timeout
+ *     - ESP_FAIL:        Event handling finished, driver uninstalled. You do not have to call this function further
  */
-esp_err_t msc_host_handle_events(uint32_t timeout_ms);
+esp_err_t msc_host_handle_events(uint32_t timeout);
 
 /**
  * @brief Gets devices information.
@@ -171,8 +178,8 @@ esp_err_t msc_host_print_descriptors(msc_host_device_handle_t device);
  *
  * @param[in] device Handle of MSC device
  * @return
- *       - ESP_OK:  The device was recovered from reset
- *       - ESP_FAI: Recovery unsuccessful, might indicate broken device
+ *       - ESP_OK:   The device was recovered from reset
+ *       - ESP_FAIL: Recovery unsuccessful, might indicate broken device
  */
 esp_err_t msc_host_reset_recovery(msc_host_device_handle_t device);
 
--- a/usb/usb_host_msc/private_include/msc_common.h
+++ b/usb/usb_host_msc/private_include/msc_common.h
@@ -45,14 +45,15 @@ typedef struct msc_host_device {
  * @brief Trigger a BULK transfer to device: zero copy
  *
  * Data buffer ownership is transferred to the MSC driver and the application cannot access it before the transfer finishes.
- * The buffer must be DMA capable, as it is going to be accessed by USB DMA.
+ * This function is true zero-copy only if the passed data buffer is in DMA capable memory, as the USB Host uses DMA transfers.
+ * If the buffer is NOT DMA capable, an intermediate DMA capable buffer is allocated and used for the transfer, resulting in memory coping.
  *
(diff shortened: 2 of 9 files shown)


usb/usb_host_msc/include/usb/msc_host.h
```diff
@@ -142,8 +142,15 @@ __attribute__((deprecated("use API from esp_private/msc_scsi_bot.h")));
...
+ * application needs to handle USB Host events itself.
+ * Do not use if MSC host install was made with create_background_task=true configuration
+ *
+ * @param[in]  timeout  Timeout in FreeRTOS tick
+ * @return
+ *     - ESP_OK:          All events handled
+ *     - ESP_ERR_TIMEOUT: No events handled within the timeout
+ *     - ESP_FAIL:        Event handling finished, driver uninstalled. You do not have to call this function further
  */
 esp_err_t msc_host_handle_events(uint32_t timeout_ms);
```
roma-jam, line 155: "Timeout in FreeRTOS tick" but argument name `timeout_ms`
Looks like an error, isn't it?

usb/usb_host_msc/src/msc_host.c
```diff
@@ -320,17 +322,34 @@ static bool is_mass_storage_device(uint8_t dev_addr)
...
 {
-    while (1) {
-        usb_host_client_handle_events(s_msc_driver->client_handle, portMAX_DELAY);
+    MSC_RETURN_ON_FALSE(s_msc_driver != NULL, ESP_ERR_INVALID_STATE);
 
-        if (s_msc_driver->end_client_event_handling) {
-            break;
-        }
+    ESP_LOGV(TAG, "USB MSC handling");
+    s_msc_driver->event_handling_started = true;
```
roma-jam, line 332: Here is the question: what if the MSC driver was configured with bg_task=false, but the `msc_host_handle_events` has never been called? 
Seems that `event_handling_started` flag will be staying false and semaphore does not let the `msc_host_uninstall` function be successfully completed. 

Does that make sense? :thinking:

usb/usb_host_msc/src/msc_host.c
```diff
@@ -514,11 +537,20 @@ esp_err_t msc_host_write_sector(msc_host_device_handle_t device, size_t sector,
...
 {
-    MSC_RETURN_ON_FALSE(s_msc_driver != NULL, ESP_ERR_INVALID_STATE);
-
-    return usb_host_client_handle_events(s_msc_driver->client_handle, pdMS_TO_TICKS(timeout_ms));
+    if (dest == NULL) {
+        return;
+    }
+    if (src != NULL) {
+        size_t len = MIN((src->bLength - USB_STANDARD_DESC_SIZE) / 2, MSC_STR_DESC_SIZE - 1);
+        wcsncpy(dest, src->wData, len);
```
roma-jam, line 547: How do you deal with unaligned access warning here? 
Or you do not have any?

roma-jam (approved):
Left a few comments. 
Otherwise, LGTM. :clap:

usb/usb_host_msc/private_include/msc_common.h
```diff
@@ -45,14 +45,15 @@ typedef struct msc_host_device {
  * @brief Trigger a BULK transfer to device: zero copy
  *
  * Data buffer ownership is transferred to the MSC driver and the application cannot access it before the transfer finishes.
- * The buffer must be DMA capable, as it is going to be accessed by USB DMA.
+ * This function is true zero-copy only if the passed data buffer is in DMA capable memory, as the USB Host uses DMA transfers.
+ * If the buffer is NOT DMA capable, an intermediate DMA capable buffer is allocated and used for the transfer, resulting in memory coping.
  *
- * This function significantly improves performance with usage of Virtual File System, which creates a intermediate buffer for each opened file.
- * The intermediate VFS buffer is then used for USB transfers too, which eliminates need of 2 large buffers and unnecessary copying of the data.
+ * This function significantly improves performance with usage of Virtual File System, which creates its own buffer for each opened file.
```
leeebo, line 51: AFAIK, `Virtual File System` does not maintain the buffer itself, the buffer is maintained by newlib with the `fopen` function, right?

tore-espressif: Yes, thank you! I'll update the docs

usb/usb_host_msc/src/msc_host.c
tore-espressif: I suppressed this warning. It is not relevant as the descriptor is always in memory that supports unaligned access

usb/usb_host_msc/src/msc_host.c
tore-espressif: Yes, it does. I will extend the tests to cover this situation

tore-espressif: Hi @roma-jam I extended the tests with the scenario you described, all tests passing!

roma-jam: Hi @tore-espressif ,

Yes, I did the same on HID and that also works.

PR #235, opened 2023-09-14
feat(usb): add change NCM network status (IEC-48)

Jacques-Zhao:
1.When switching routers, the HOST can re-initiate DHCP by switching the network status.
2. Delete ECM/RNDIS config because NCM can support MAC, Linux and Windows11.

This PR relies on a change to tinyusb (https://github.com/espressif/tinyusb/pull/16).


Jacques-Zhao:
@tore-espressif @david-cermak

usb/esp_tinyusb/idf_component.yml
```diff
@@ -5,7 +5,7 @@ url: https://github.com/espressif/idf-extra-components/tree/master/usb/esp_tinyu
 dependencies:
   idf: '>=5.0' # IDF 4.x contains TinyUSB as submodule
   tinyusb:
-    version: '>=0.14.2'
+    version: '>=0.15.1'
```
Jacques-Zhao, line 8: I am not sure the next tinyusb version is 0.15.1, but this PR requires the next version after tinyusb PR is merferd(https://github.com/espressif/tinyusb/pull/16)

usb/esp_tinyusb/Kconfig
```diff
@@ -241,21 +241,11 @@ menu "TinyUSB Stack"
...
-            prompt "Network mode"
-            default TINYUSB_NET_MODE_NONE
+    menu "Network driver (NCM)"
+        config TINYUSB_NET_MODE_NCM
+            bool "Enable TinyUSB Network feature(NCM)"
+            default n
             help
-                Select network driver you want to use.
-
-            config TINYUSB_NET_MODE_ECM_RNDIS
```
tore-espressif, line 251: Could you please explain why do we remove the option to use ECM/RNDIS?

I think we can leave the option there, even though we don't support it in `esp_tinyusb`. This way, the user can still implement it himself, in case he does not want to use NCM

tore-espressif (changes requested):

usb/esp_tinyusb/Kconfig
Jacques-Zhao: NCM can be used on linux/MAC/windows, ECM/RNDIS do duplicate work. In addition, if ECM/RNIDS need to be supported on both windows(Only support RNDIS) and MAC(Only support ECM),  we need to use two descriptor configurations, but esp_tinyusb can only support one configurations.
I removed this option because I thought it would be weird for customers to think we had already implemented it

usb/esp_tinyusb/idf_component.yml
Jacques-Zhao: Thanks. Updated!

tore-espressif (approved):
LGTM!

tore-espressif:
@Jacques-Zhao is this ready for merge?
If yes, could you please solve the formating check?

PR #236, opened 2023-09-14
esp_delta_ota: Add support to pass user data to write callback function

hmalpani:
Add support to pass user data to write callback in esp_delta_ota


esp_delta_ota/include/esp_delta_ota.h
```diff
@@ -19,10 +19,13 @@ typedef esp_err_t (*src_read_cb_t)(uint8_t *buf_p, size_t size, int src_offset);
 
 // Callback for working on the data generated by applying patch on the source data
 typedef esp_err_t (*merged_stream_write_cb_t)(const uint8_t *buf_p, size_t size);
+typedef esp_err_t (*merged_stream_write_cb_v2_t)(const uint8_t *buf_p, size_t size, void *user_data);
 
 typedef struct esp_delta_ota_cfg {
-    src_read_cb_t read_cb;                  /*!< Read Callback */
-    merged_stream_write_cb_t write_cb;      /*!< Write Callback */
+    void *user_data;                                        /*!< User Data */
```
mahavirj, line 25: Would it help if we add new API to retrieve the user ctx and then application can use it inside write callback as required?

```
void *esp_delta_ota_get_user_ctx(esp_delta_ota_handle_t handle);
```

hmalpani: In that case, the delta_ota handle cannot be a local variable. If it is a local variable, the write_cb cannot use the new API as it will require the handle to get the user data.

PSONALl (approved):

PR #239, opened 2023-09-17
feat: add sbom manifest files for 3rd party libraries (IEC-50)

fhrbata:
# Checklist

- [ ] CI passing

# Change description
The esp-idf-sbom tool now supports "referenced" manifests allowing to add sbom files for libraries, which we do not have control over. IOW submodules not managed by Espressif. This adds sbom manifests for such libraries.

Since we are moving away from keeping sbom information in `.gitmodules`, this PR removes sbom related variables from `.gitmodules`. We also need a different approach for `hash` checking and sbom manifest validation, because till now it was all based on the info in `.gitmodules` only. Latest `esp-idf-sbom` has a new command, which allows to validate specific sbom manifest files, .gitmodules and also directories, which are searched for any possible sbom manifest file(sbom.yml, referenced manifests, idf_component.yml, .gitmodules.) This allows to simply validate all manifests in a repository. This command, `esp-idf-sbom manifest validate`, is now used in the CI instead of the `test_sbom.py` from `esp-idf`.

Along with `esp-idf-sbom manifest validate`, there is also `esp-idf-sbom manifest check`, which behaves the same as the former, but it checks CPEs found in the manifest files against NVD. Meaning we can simply check the whole repository for possible vulnerabilities without actually generating the SBOM SPDX file.

### DIFF
--- a/.github/workflows/test_sbom.yml
+++ b/.github/workflows/test_sbom.yml
@@ -1,4 +1,4 @@
-name: Run SBOM submodule test
+name: Run SBOM manifests validation test
 
 on:
   pull_request:
@@ -6,13 +6,11 @@ on:
 
 jobs:
   test_sbom:
-    name: Run SBOM submodule test
-    runs-on: ubuntu-20.04
-    container: espressif/idf:latest
+    name: Run SBOM manifests validation test
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          submodules: 'true'
       - run: |
           git config --global safe.directory $(pwd)
-          python3 ${IDF_PATH}/tools/test_sbom/test_submodules.py
+          pip install esp-idf-sbom
+          python3 -m esp_idf_sbom manifest validate
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,101 +1,50 @@
-# Submodules SBOM information
-# ---------------------------
-# Submodules, which are used directly and not forked into espressif namespace should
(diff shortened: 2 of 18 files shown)


fhrbata:
There are few things, which need to be done before this can be merged.
1. We need `esp-idf-sbom` support for hash validation in manifest files, not only in `.gitmodules`. This is WIP.
2. We probably should remove sbom information currently stored in  `.gitmodules`.
So for this moment this is just a preview how the sbom information would be stored using "referenced" manifests.

mahavirj:
Change generally LGTM!

> We probably should remove sbom information currently stored in .gitmodules.

Agree, we can remove the information as a part of this PR itself.

igrr:
Looks good to me, as well.

fhrbata:
The required changes for `esp-idf-sbom` are scheduled to be merged after the pipeline passes. Once the new 0.6 version is available, I will remove the draft stated of this MR and we will see if the pipeline passes.

kumekay (approved):
@fhrbata idf_component.yml changes look good and should not cause any issues with component manager >=1.3.0

fhrbata:
Ok, this seems to be somehow ready, but IIUC @kumekay has a PR in progress, which will allow to backup also the sbom part, or any part generally not known to the component manager,  in the idf_component.yml manifest into DB. So I will keep this in draft for now and remove it once it's done.

fhrbata:
@mahavirj /others, this should be ready now. Thank you.

mahavirj (approved):

PR #240, opened 2023-09-19
Quirc esp-eye example (IEC-51)

FBEZ:
# Checklist

- [ ] CI passing

# Change description
It is an example for the quirc component using an esp-eye


quirc/examples/esp-eye-qrcode/main/include/app_peripherals.h
```diff
@@ -0,0 +1,248 @@
+#pragma once
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/queue.h"
+#include "freertos/task.h"
+#include "freertos/semphr.h"
+
+#include "esp_camera.h"
+
+#define CONFIG_CAMERA_MODULE_ESP_EYE 1
```
igrr, line 10: I would recommend bringing board definitions using the board support package. This way to support a new board, work needs to be done only on BSP side, and the example doesn't have to be updated.

Example: https://github.com/espressif/qrcode-demo/blob/188cc80ca4e3654f597f00cdcb2193fad1709010/main/idf_component.yml#L4

(It's okay if that means that fewer boards are supported. After all, it's just an example.)

quirc/examples/esp-eye-qrcode/main/app_main.c
```diff
@@ -0,0 +1,85 @@
...
+
+        esp_camera_fb_return(fb);
+        vTaskDelay(10 / portTICK_PERIOD_MS);
+    }
+
+    quirc_destroy(q);
+}
+
+
+void app_main()
```
igrr, line 82:
```suggestion
void app_main(void)
```

quirc/examples/esp-eye-qrcode/main/app_main.c
```diff
@@ -0,0 +1,85 @@
...
+#include "esp_log.h"
+#include "esp_system.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "app_peripherals.h"
+#include "quirc.h"
+
+static const char *TAG = "APP_CODE_SCANNER";
+
+static void decode_task()
```
igrr, line 13: FreeRTOS task function definition should look like this:
```suggestion
static void decode_task(void* arg)
```

quirc/examples/esp-eye-qrcode/main/app_main.c
```diff
@@ -0,0 +1,85 @@
...
+            else
+                printf("%d/%d] DATA: %s\n",i+1,num_codes,data.payload);
+        }
+
+        esp_camera_fb_return(fb);
+        vTaskDelay(10 / portTICK_PERIOD_MS);
+    }
+
+    quirc_destroy(q);
+}
```
igrr, line 79: `vTaskDelete(NULL);` missing here?

quirc/examples/esp-eye-qrcode/main/app_main.c
```diff
@@ -0,0 +1,85 @@
...
+        vTaskDelete(NULL);
+        return;
+    }
+
+    camera_fb_t *fb = NULL;
+
+    // Initializing the quirc handle
+    struct quirc *q = quirc_new();
+    if (!q) {
+        ESP_LOGE(TAG,"Failed to allocate memory\n");
```
igrr, line 25: Newline at the end of format strings is not necessary for ESP_LOG, it's added automatically.

Also, missing space after comma. (Have you installed the pre-commit hooks using `pre-commit install`? They should automatically format the source files.)

igrr (changes requested):
Hey @FBEZ, thanks for your PR!

I've left a few comments, please ping me if you have any questions or when I should review again.

quirc/examples/esp-eye-qrcode/main/include/app_peripherals.h
FBEZ: Thank you Ivan for your kind and detailed feedback. 

Using the bsp approach would be the best choice for sure. Right now though I don't see the bsp for the esp-eye (meaning the [older one](https://www.espressif.com/en/products/devkits/esp-eye/overview)). I could setup a very basic bsp for this board starting form the esp32-s3-eye and update this example accordingly. 

I will receive and Esp32-s3-eye the next week to test this example on it anyway.

quirc/examples/esp-eye-qrcode/main/app_main.c
FBEZ: This part of the code isn't supposed to be reached. I'm not sure what the best approach would be here. 
Do you suggest the vTaskDelete or to remove the quirc_destroy completely?

igrr: Depends on the error handling strategy — see the other comment about this.

If you decide to handle errors by terminating the task, probably having the cleanup path with quirc_destroy would make sense. You could use an `out:` goto label and use macros like `ESP_GOTO_ON_ERROR` to handle errors and perform cleanup.

If you decide to handle errors by aborting (which is not an unreasonable strategy for an example project) then quirc_destroy can be removed since it's unreachable.

quirc/examples/esp-eye-qrcode/main/include/app_peripherals.h
igrr: Ah yes, we don't seem to have a BSP for the original ESP-EYE.

I think at this point it's probably safe to recommend ESP32-S3-EYE for new applications, so I wouldn't mind keeping this example supported on that board only. Later when we add ESP-EYE BSP, we can update the example to support it as well.

quirc/examples/esp-eye-qrcode/main/app_main.c
FBEZ: Ok then since as you say it's an example, I think it's best to keep it the simplest possible. I removed the quirc_destroy.

quirc/examples/esp-eye-qrcode/main/include/app_peripherals.h
FBEZ: Ok, then it will take a few days to receive the hardware and make the changes. 
If it's better, I can close this PR here and reopen when everything it's in place. Otherwise I'll leave it here and update it in a few days.

igrr: You can keep this PR, just make the changes and push to the same branch when you are ready. Later once everything is ready you can clean up the commit history.

PR #245, opened 2023-09-26
feat(ci): add periodic and ad hoc vulnerability scan

fhrbata:
# Checklist

- [ ] CI passing

# Change description

esp-idf-sbom allows to scan whole repository/directory for all possible manifest files(idf_component.yml, sbom.yml and its referenced manifests, .gitmodules) and check them for possible vulnerabilities based on the cpe variable in manifest.

This adds scheduled scan at every midnight and also ad hoc(dispatch workflow) allowing to scan on demand.

### DIFF
--- a/.github/workflows/vulnerability_scan.yml
+++ b/.github/workflows/vulnerability_scan.yml
@@ -1,0 +1,35 @@
+name: Vulnerability scan
+
+on:
+  schedule:
+     - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  vulnerability-scan:
+    name: Vulnerability scan
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install esp-idf-sbom
+        run: pip3 install esp-idf-sbom
+
+      - name: Vulnerability scan
+        id: scan
+        run: python3 -m esp_idf_sbom manifest check
+
+      - name: Send result
+        if: ${{ !cancelled() }}
+        env:
+          MATTERMOST_WEBHOOK: ${{ secrets.MATTERMOST_WEBHOOK }}
+          # Uses ternary operator. The double quotes are needed because of the colons
+          # for mattermost markup, which are confusing the yaml parser.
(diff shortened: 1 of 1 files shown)


fhrbata:
This is how the report and workflow looks like in my private fork https://github.com/fhrbata/idf-extra-components/actions/runs/6309525206/job/17129607763

mahavirj:
Thanks for adding this! 

One question: how will this action notify us if it finds any vulnerabilities?

fhrbata:
@mahavirj The command exits with `1` if a vulnerability is found. So my expectation was that the pipeline fails and an email with info about failed pipeline will be send. But I cannot see how to customize this. Meaning I don't see notification settings. I was hoping people will be able to set notification for this failed pipeline, but this doesn't seem possible.

 https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs

But I have very limited experience with github, so may it's somehow possible and I just don't know how.

One quick thought is that maybe we can store some email address in secrets and explicitly send the report to that address if the check fails(finds some new vulnerability)

igrr:
I might also suggest that we set up a simple webhook in Mattermost, so that when the action fails, it notifies us on some channel.

fhrbata:
@igrr Thank you very much for the suggestion. IIUC we would set up an  [incoming webhhok](https://developers.mattermost.com/integrate/webhooks/incoming/) in mattermost and use e.g.  [mattermost notify action](https://github.com/mattermost/action-mattermost-notify) in the workflow to post a message into the mattermost channel if the workflow fails or am I missing something? Thank you

fhrbata:
@igrr I guess we would like to use something like this in GL also, so maybe just sending the msg from the jobs with curl would be enough. Do you want this to be handled as part of this PR or can it be handled in a follow-up PR?
Thank you

fhrbata:
@mahavirj Would some mattermost channel work for you as a source for the notifications? Thank you

mahavirj:
@fhrbata Yeah, that sounds good to me.

mahavirj (approved):

fhrbata:
Hi @mahavirj I needed to update this PR, but now it should be hopefully all set. I tested both cases(ok, failed) in my private fork and the testing mattermost channel. The `esp-idf-sbom` changes are also published. Thank you!

PR #253, opened 2023-10-18
led_strip: Support for IDF4.4

espzav:
# Checklist

- [x] CI passing

# Change description
Updated LED Strip component for working with IDF4.4 (only with RMT)

### DIFF
--- a/led_strip/CHANGELOG.md
+++ b/led_strip/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.5.0
+
+- Enabled support for IDF4.4 and above
+    - only with RMT backend
+- Added API `led_strip_set_pixel_hsv`
+
 ## 2.4.0
 
 - Support configurable SPI mode to control leds
--- a/led_strip/CMakeLists.txt
+++ b/led_strip/CMakeLists.txt
@@ -2,8 +2,12 @@ include($ENV{IDF_PATH}/tools/cmake/version.cmake)
 
 set(srcs "src/led_strip_api.c")
 
-if(CONFIG_SOC_RMT_SUPPORTED)
-    list(APPEND srcs "src/led_strip_rmt_dev.c" "src/led_strip_rmt_encoder.c")
+if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.0")
+    if(CONFIG_SOC_RMT_SUPPORTED)
+        list(APPEND srcs "src/led_strip_rmt_dev.c" "src/led_strip_rmt_encoder.c")
+    endif()
+else() 
+    list(APPEND srcs "src/led_strip_rmt_dev_idf4.c")
 endif()
 
 # the SPI backend driver relies on something that was added in IDF 5.1
--- a/led_strip/include/led_strip_rmt.h
+++ b/led_strip/include/led_strip_rmt.h
@@ -7,8 +7,13 @@
 
 #include <stdint.h>
 #include "esp_err.h"
-#include "driver/rmt_types.h"
 #include "led_strip_types.h"
+#include "esp_idf_version.h"
+
+
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
+#include "driver/rmt_types.h"
+#endif
 
 #ifdef __cplusplus
 extern "C" {
@@ -18,12 +23,17 @@ extern "C" {
  * @brief LED Strip RMT specific configuration
  */
 typedef struct {
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
+    uint8_t rmt_channel;
+    size_t mem_block_symbols;   /*!< How many RMT symbols can one RMT channel hold at one time. Set to 0 will fallback to use the default size. */
+#else
     rmt_clock_source_t clk_src; /*!< RMT clock source */
     uint32_t resolution_hz;     /*!< RMT tick resolution, if set to zero, a default resolution (10MHz) will be applied */
     size_t mem_block_symbols;   /*!< How many RMT symbols can one RMT channel hold at one time. Set to 0 will fallback to use the default size. */
     struct {
         uint32_t with_dma: 1;   /*!< Use DMA to transmit data */
     } flags;
+#endif
 } led_strip_rmt_config_t;
 
 /**
--- a/led_strip/src/led_strip_rmt_dev_idf4.c
+++ b/led_strip/src/led_strip_rmt_dev_idf4.c
@@ -1,0 +1,171 @@
+/*
+ * SPDX-FileCopyrightText: 2022-2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <stdlib.h>
+#include <string.h>
+#include <sys/cdefs.h>
+#include "esp_log.h"
+#include "esp_check.h"
+#include "driver/rmt.h"
+#include "led_strip.h"
+#include "led_strip_interface.h"
+
+static const char *TAG = "led_strip_rmt";
+
+#define WS2812_T0H_NS   (350)
+#define WS2812_T0L_NS   (1000)
+#define WS2812_T1H_NS   (1000)
+#define WS2812_T1L_NS   (350)
+#define WS2812_DELAY_MS (100)
+
+// the memory size of each RMT channel, in words (4 bytes)
+#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2
+#define LED_STRIP_RMT_DEFAULT_MEM_BLOCK_SYMBOLS 8
+#else
+#define LED_STRIP_RMT_DEFAULT_MEM_BLOCK_SYMBOLS 6
+#endif
+
+static uint32_t ws2812_t0h_ticks = 0;
+static uint32_t ws2812_t1h_ticks = 0;
+static uint32_t ws2812_t0l_ticks = 0;
+static uint32_t ws2812_t1l_ticks = 0;
+
+typedef struct {
+    led_strip_t base;
+    rmt_channel_t rmt_channel;
+    uint32_t strip_len;
+    uint8_t bytes_per_pixel;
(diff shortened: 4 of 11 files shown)


espzav:
@tore-espressif PTAL (it is updated for Generic BSP)

tore-espressif (approved):
Works OK with my examples. Thank you @espzav 

@suda-morris PTAL too

led_strip/CMakeLists.txt
```diff
@@ -2,8 +2,10 @@ include($ENV{IDF_PATH}/tools/cmake/version.cmake)
 
 set(srcs "src/led_strip_api.c")
 
-if(CONFIG_SOC_RMT_SUPPORTED)
```
suda-morris, line 5: why remove this? FYI, we have some chip that doesn't have the RMT peripheral, e.g. ESSP32C2

led_strip/include/led_strip_rmt.h
```diff
@@ -18,12 +23,16 @@ extern "C" {
...
+#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
+    uint8_t rmt_channel;
+#else
     rmt_clock_source_t clk_src; /*!< RMT clock source */
     uint32_t resolution_hz;     /*!< RMT tick resolution, if set to zero, a default resolution (10MHz) will be applied */
     size_t mem_block_symbols;   /*!< How many RMT symbols can one RMT channel hold at one time. Set to 0 will fallback to use the default size. */
     struct {
         uint32_t with_dma: 1;   /*!< Use DMA to transmit data */
     } flags;
+#endif
```
suda-morris, line 35: Even with the legacy driver, we can still reuse many of these configurations
```suggestion
#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
    uint8_t rmt_channel;
#else
    rmt_clock_source_t clk_src; /*!< RMT clock source */
#endif
    uint32_t resolution_hz;     /*!< RMT tick resolution, if set to zero, a default resolution (10MHz) will be applied */
    size_t mem_block_symbols;   /*!< How many RMT symbols can one RMT channel hold at one time. Set to 0 will fall back to use the default size. */
    struct {
        uint32_t with_dma: 1;   /*!< Use DMA to transmit data */
    } flags;
```

led_strip/src/led_strip_rmt_dev_idf4.c
```diff
@@ -0,0 +1,157 @@
...
+        bytes_per_pixel = 4;
+    } else if (led_config->led_pixel_format == LED_PIXEL_FORMAT_GRB) {
+        bytes_per_pixel = 3;
+    } else {
+        assert(false);
+    }
+
+    rmt_config_t config = RMT_DEFAULT_CONFIG_TX(led_config->strip_gpio_num, dev_config->rmt_channel);
+    // set counter clock to 40MHz
+    config.clk_div = 2;
```
suda-morris, line 123: we should set the `config.mem_block_num` based on the `led_strip_rmt_config::mem_block_symbols` user setting

led_strip/src/led_strip_rmt_dev_idf4.c
```diff
@@ -0,0 +1,157 @@
...
+    if (led_config->led_pixel_format == LED_PIXEL_FORMAT_GRBW) {
+        bytes_per_pixel = 4;
+    } else if (led_config->led_pixel_format == LED_PIXEL_FORMAT_GRB) {
+        bytes_per_pixel = 3;
+    } else {
+        assert(false);
+    }
+
+    rmt_config_t config = RMT_DEFAULT_CONFIG_TX(led_config->strip_gpio_num, dev_config->rmt_channel);
+    // set counter clock to 40MHz
```
suda-morris, line 122: the comment is not correct because it assumes the RMT default clock source's frequency is 80MHz.

suda-morris:
> @espzav Can support control led by HSV? This makes it easier to control the brightness.

@lijunru-hub 
It's just a helper function, which is not tied to the led_strip driver very much. You may consider creating a new component e.g. "color_converter" to include e.g. `void hsv2rgb(uint32_t h, uint32_t s, uint32_t v, uint32_t *r, uint32_t *g, uint32_t *b)`

led_strip/CMakeLists.txt
espzav: I am sorry, I thought, that this constant is not available in IDF4.4 - I'll fix it.

led_strip/src/led_strip_rmt_dev_idf4.c
espzav: I copied from component from IDF4.4 examples. But I think, that default clock is 80MHz, but there is divider set to 2:
`config.clk_div = 2;` I think, that the comment is right.

espzav:
> > @espzav Can support control led by HSV? This makes it easier to control the brightness.
> 
> @lijunru-hub It's just a helper function, which is not tied to the led_strip driver very much. You may consider creating a new component e.g. "color_converter" to include e.g. `void hsv2rgb(uint32_t h, uint32_t s, uint32_t v, uint32_t *r, uint32_t *g, uint32_t *b)`

@suda-morris @lijunru-hub I think, that it is good idea to alow set color by HSV. We can convert color by this function inside this component: https://github.com/espressif/esp-idf/blob/8fc8f3f47997aadba21facabc66004c1d22de181/examples/peripherals/rmt/led_strip/main/led_strip_example_main.c#L30 what do you think?

led_strip/CMakeLists.txt
espzav: Fixed

espzav:
@tore-espressif and @suda-morris Thank you for the review! Everything fixed.

lijunru-hub:
@espzav @suda-morris I believe adding a esp_err_t led_strip_set_hsv(led_strip_handle_t strip, uint32_t index, uint32_t hue, uint32_t saturation, uint32_t brightness); would make this component more comprehensive.

espzav:
@suda-morris @lijunru-hub I added HSV set pixel: https://github.com/espressif/idf-extra-components/pull/253/commits/8c299f5d97be02f27ebb996ca0791dc2bd68473b (I can remove it, if you want). 

And I tested it with my custom BSP API with `led_indicator` component like this. It is working!

```

static void bsp_rgb_led_rgb2hsv(uint16_t red, uint16_t green, uint16_t blue, uint16_t *h, uint8_t *s, uint8_t *v)
{
    double hue, saturation, value;
    double m_max = MAX(red, MAX(green, blue));
    double m_min = MIN(red, MIN(green, blue));
    double m_delta = m_max - m_min;

    value = m_max / 255.0;

    if (m_delta == 0) {
        hue = 0;
        saturation = 0;
    } else {
        saturation = m_delta / m_max;

        if (red == m_max) {
            hue = (green - blue) / m_delta;
        } else if (green == m_max) {
            hue = 2 + (blue - red) / m_delta;
        } else {
            hue = 4 + (red - green) / m_delta;
        }

        hue = hue * 60;

        if (hue < 0) {
            hue = hue + 360;
        }
    }

    *h = (int)(hue + 0.5);
    *s = (int)(saturation * 100 + 0.5);
    *v = (int)(value * 100 + 0.5);
}

static esp_err_t bsp_rgb_led_set_brightness(void *hardware_data, uint32_t brightness)
{
    bsp_led_t led = (bsp_led_t)hardware_data;
    uint32_t hue = 0;
    uint32_t sat = 0;
    uint32_t val = 0;

    /* Set default color, if not set before */
    if (bsp_leds_color[led].red == 0 && bsp_leds_color[led].green == 0 && bsp_leds_color[led].blue == 0) {
        bsp_rgb_led_set_color(led, &BSP_LED_RGB_DEFAULT_COLOR);
    }
    
    bsp_rgb_led_rgb2hsv(bsp_leds_color[led].red, bsp_leds_color[led].green, bsp_leds_color[led].blue, &hue, &sat, &val);

    /* Set the LED pixel using HSV */
    led_strip_set_pixel_hsv(led_strip, led, hue, sat, brightness);
    /* Refresh the strip to send data */
    led_strip_refresh(led_strip);

    return ESP_OK;
}
```

espzav:
@tore-espressif @suda-morris are you okay with keep the HSV function in this component? Can I merge it?

led_strip/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.5.0
+
+- Enabled support for IDF4.4 and above
+    - only RMT mode
```
suda-morris, line 4:
```suggestion
    - only with RMT backend
```

led_strip/CMakeLists.txt
suda-morris: > I thought, that this constant is not available in IDF4.4 

Sorry @espzav You're right, soc caps are injected into the Kconfig system starting from esp-idf v5.0. I'm afraid this statement will be false when we build it with esp-idf v4.4

led_strip/CMakeLists.txt
```diff
@@ -3,7 +3,11 @@ include($ENV{IDF_PATH}/tools/cmake/version.cmake)
 set(srcs "src/led_strip_api.c")
 
 if(CONFIG_SOC_RMT_SUPPORTED)
-    list(APPEND srcs "src/led_strip_rmt_dev.c" "src/led_strip_rmt_encoder.c")
+    if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.0")
+        list(APPEND srcs "src/led_strip_rmt_dev_idf5.c" "src/led_strip_rmt_encoder.c")
```
suda-morris, line 7: nitpick, just the `idf5` sounds weird, I'm sure we will continue with the current RMT driver implementation in esp-idf v6/v7.

suda-morris (approved):
LGTM! Thanks for adding the support for the idf 4.4 users.

led_strip/CMakeLists.txt
espzav: You are right @suda-morris, I was tested, but I didn't remove whole build folder before and I though, it is working... I changed the condition only for IDF5 and higher.

espzav:
@suda-morris Thank you for review. I fixed everything. I think, we can merge it after tests pass.

PR #261, opened 2023-10-27
Relax IDF version dependency till v4.0

vikramdattu:
qrcode component is mostly independent and can also be used with IDF v4.0 Relaxing the version dependency on IDF makes the component usable for projects with older IDFs.

# Checklist

- [ ] CI passing

# Change description

- Relax the IDF version dependency for `qrcode` component till IDF `v4.0`


qrcode/idf_component.yml
```diff
@@ -1,5 +1,5 @@
-version: "0.1.0~1"
+version: "0.1.0~2"
 description: QR Code generator
 url: https://github.com/espressif/idf-extra-components/tree/master/qrcode
 dependencies:
-  idf: ">=4.3"
+  idf: ">=4.0"
```
shahpiyushv, line 5: Do we even need this under dependencies?

igrr:
Just keep in mind that we don't actually test anything older than IDF v4.3 in idf-extra-components, and we will remove v4.3 testing at the end of December when that release becomes EoL. So if the component becomes incompatible with the declared minimal supported IDF version, we won't notice this.

vikramdattu:
> Just keep in mind that we don't actually test anything older than IDF v4.3 in idf-extra-components, and we will remove v4.3 testing at the end of December when that release becomes EoL. So if the component becomes incompatible with the declared minimal supported IDF version, we won't notice this.

@igrr I think that should be fine. Not having strong opinions here, but maybe we make some exceptions and have some of the components also get tested for the older IDF builds? (But again, that opens door for more and more components :( )

qrcode/idf_component.yml
vikramdattu: @shahpiyushv makes sense, as component manager anyway is not supported for `IDF <= 4.0`! Made the change.

tore-espressif (approved):

shahpiyushv (approved):

Issue #248, opened 2023-10-04
Add API reference for components


igrr:
### Is your feature request related to a problem?

It is difficult to discover/learn the API of components which aren't ports of well-known libraries.

### Describe the solution you'd like.

Component Registry now supports displaying the contents of API.md file, if it is provided in the component directory. We also have a pre-commit hook to generate API.md from Doxygen-generated XML files.

### Describe alternatives you've considered.

- Find the header file of the component, read Doxygen comments
- Read examples (which are missing from most components...)

### Additional context.

Example repo where API.md is set up: https://github.com/igrr/pa1010d

PR #265, opened 2023-10-30
docs(led_strip): add api reference

suda-morris:
Added api.md according to approach described in 
https://github.com/espressif/idf-extra-components/issues/248

Fixed the wrong mem_block_num configuration in the *led_strip_rmt_dev_idf4.c*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ sdkconfig.old
 dependencies.lock
 **/managed_components/**
 .vscode/
+doxygen/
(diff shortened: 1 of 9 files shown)


espzav:
Hi @suda-morris Thank you for fix my issues. I am sorry about that. I don't understand memory blocks setting, which you fixed.

Please, could you add one another fix before merging: https://github.com/espressif/idf-extra-components/commit/b0f85ca5c474c36795a30f99871375740792862d

Thank you very much.

suda-morris:
@espzav Never mind, it's indeed confusing between "mem block symbols" and "mem block num". Thanks for finding a new issue. I have fixed it in the PR as well, together with other clean up.

espzav:
@suda-morris Thank you very much!

PR #269, opened 2023-11-06
change(esp_tinyusb): Update esp_tinyusb task affinity config type

Dazza0:
# Change description

ESP-IDF updates the task affinity config option from hex to int, and provides constant values for each affinity value. This commit syncs `TINYUSB_TASK_AFFINITY` with those changes.

See internal MR !26701 for more details.

### DIFF
--- a/usb/esp_tinyusb/Kconfig
+++ b/usb/esp_tinyusb/Kconfig
@@ -46,10 +46,10 @@ menu "TinyUSB Stack"
         endchoice
 
         config TINYUSB_TASK_AFFINITY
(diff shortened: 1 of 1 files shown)


Dazza0:
@tore-espressif Do we need to bump the bugfix version for every change?

Dazza0:
@tore-espressif  Sorry, looks like there's not a good way to introduce this change to the managed components in a non-breaking manner. I'll closing this PR (and associated MR) and will delay this change until ESP-IDF v6.0.

PR #272, opened 2023-11-28
esp_jpeg: Fix BSP versions for using new version of button component (without warnings in new IDF).

espzav:
# Checklist

- [x] CI passing

# Change description
Fix for this error: https://github.com/espressif/idf-extra-components/actions/runs/7004078689/job/19060622718?pr=271


igrr (commented):
@espzav should we also bump esp_jpeg component version to release this change?

espzav:
> @espzav should we also bump esp_jpeg component version to release this change?

@igrr I think, that it is not necessary, but we can do it. --> Done.

PR #279, opened 2023-12-06
feat(usb/host): Update manifest files for USB Host components

roma-jam:
# Checklist

- [x] CI passing

# Change description
Changes in the `idf_component.yml` component files: 
- Added `## IDF Component Manager Manifest File` line where absent
- Added `esp32p4` target

### DIFF
--- a/usb/usb_host_hid/idf_component.yml
+++ b/usb/usb_host_hid/idf_component.yml
@@ -1,11 +1,10 @@
-version: "1.0.1"
+## IDF Component Manager Manifest File
+version: "1.0.1~1"
 description: USB Host HID driver
 url: https://github.com/espressif/idf-extra-components/tree/master/usb/usb_host_hid
-
 dependencies:
   idf: ">=4.4"
-
 targets:
   - esp32s2
   - esp32s3
-
+  - esp32p4
--- a/usb/usb_host_vcp/idf_component.yml
+++ b/usb/usb_host_vcp/idf_component.yml
@@ -1,5 +1,5 @@
 ## IDF Component Manager Manifest File
-version: "1.0.0~1"
+version: "1.0.0~2"
 description: USB Host Virtual COM Port Service
 url: https://github.com/espressif/idf-extra-components/tree/master/usb/usb_host_vcp
 dependencies:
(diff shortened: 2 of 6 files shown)


usb/usb_host_ch34x_vcp/idf_component.yml
```diff
@@ -2,8 +2,14 @@
...
 dependencies:
   espressif/usb_host_cdc_acm:
     version: ">=1.0.4,<3.0.0"
     public: true
   idf: ">=4.4"
+
+targets:
+  - esp32s2
+  - esp32s3
+  - esp32p4
```
tore-espressif, line 15: The reason why I did not add the `targets` to `usb_host_xxx_vcp` components is that the targets requirement are taken from `usb_host_cdc_acm` on which they depend.

I did not want to update all the drivers for a new target, but only the base `usb_host_cdc_acm`.

Does it make sense? can we remove it?

usb/usb_host_hid/idf_component.yml
```diff
@@ -1,3 +1,4 @@
+## IDF Component Manager Manifest File
 version: "1.0.1"
```
tore-espressif, line 2: In order for this change to be uploaded to ESP registry, we must bump the `version`.
In this case we can only bump the revision number to `1.0.1~1`. Which means: no changes in code; only a metadata change

tore-espressif (commented):
@roma-jam Thanks for the update. I left 2 comments ;)

usb/usb_host_ch34x_vcp/idf_component.yml
roma-jam: Yes, when I started to add them in the end this popped up in my head.... that that was the reason. 
Agreed.

usb/usb_host_hid/idf_component.yml
roma-jam: Oh...ok. 
Do we want this line in manifest?

roma-jam:
@tore-espressif , thanks for quick review. 
Regarding the lines - the lines were only in HID and MSC component (because I did HID on the MSC base once) so they had these lines. All other  *yml files don't have them. So I remove the lines and make all *yml files the same.

usb/usb_host_vcp/idf_component.yml
```diff
@@ -1,9 +1,13 @@
...
 url: https://github.com/espressif/idf-extra-components/tree/master/usb/usb_host_vcp
 dependencies:
   espressif/usb_host_cdc_acm:
     version: ">=1.0.4,<3.0.0"
     public: true
   idf: ">=4.4"
+targets:
+  - esp32s2
+  - esp32s3
+  - esp32p4
```
tore-espressif, line 13: Same here, the component depends on `usb_host_cdc_acm` so the target requirements are taken from there

tore-espressif (approved):

PR #282, opened 2023-12-12
feat(usb/esp_tinyusb): Add AUDIO Device Class configuration to Konfig

roma-jam:
# Checklist

- [x] CI passing

# v1.4.3 Change description
- Adding the possibility to config the tinyUSB Audio Device Class with KConfig

Closes https://github.com/espressif/esp-idf/issues/12774

### DIFF
--- a/usb/esp_tinyusb/include/tusb_config.h
+++ b/usb/esp_tinyusb/include/tusb_config.h
@@ -49,6 +49,10 @@ extern "C" {
 #   define CONFIG_TINYUSB_HID_COUNT 0
 #endif
 
+#ifndef CONFIG_TINYUSB_AUDIO_ENABLED
+#   define CONFIG_TINYUSB_AUDIO_ENABLED 0
+#endif
+
 #ifndef CONFIG_TINYUSB_MIDI_COUNT
 #   define CONFIG_TINYUSB_MIDI_COUNT 0
 #endif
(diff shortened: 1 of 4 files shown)


usb/esp_tinyusb/include/tusb_config.h
```diff
@@ -114,6 +118,17 @@ extern "C" {
 // MSC Buffer size of Device Mass storage
 #define CFG_TUD_MSC_BUFSIZE         CONFIG_TINYUSB_MSC_BUFSIZE
 
+// AUDIO macros
+#define CFG_TUD_AUDIO_FUNC_1_DESC_LEN                                 TUD_AUDIO_MIC_ONE_CH_DESC_LEN
```
roma-jam, line 122: @tore-espressif , 
When you will have time, could you please check the configuration parameters for the audio device here. 
I think couple of things: 
- What do we really need to configure? 
- Is there any better place to define such variables in the `esp_tinyusb` component?

roma-jam:
Closed as to fully support AUDIO class driver configuration there is a necessity to provide clearer API and be able to support all the configs options. 

Regarding the idea that some of that changes should be closely tied to the hardware microphone configuration and changes were made in that MR are nut fully fulfill the AUDIO class driver support, seems that in case of AUDIO class driver right now the better option is to use tinyUSB directly.

akumpf:
I've verified that these changes allow audio to work on ESP32-S3. Tested on OSX. Can also work alongside MIDI and CDC/Serial for quite a fantastic mix of USB capabilities and projects.

Would love for this to get merged into the main branch, even if it's just a small step toward full audio support. As more people experiment, that specific features to support would also become more clear. Thank you @roma-jam!

tore-espressif:
@akumpf Thank you very much for your feedback!
We decided not to merge this PR in its current form for following reasons:
1. This PR only exposes TinyUSB API that **can be used even without this change**. (you can follow instructions [here](https://components.espressif.com/components/espressif/tinyusb) in section 'Use TinyUSB only, without the additions')
2. The configuration in `tusb_config.` is fixed and does not allow the flexibility audio applications typically need (various sample rates, bit widths, channel numbers...)
3. In this PR, there is no biding to ESP's audio peripherals, such as I2S, or more abstract SW components as [esp_codec_dev](https://components.espressif.com/components/espressif/esp_codec_dev)

We will revisit this topic after we are ready to provide fully-fledged USB Audio solutions

PR #283, opened 2023-12-13
fix CI failures on master

igrr:
This PR fixes three CI build failures on master branch:

- catch2 component: https://github.com/espressif/idf-extra-components/actions/runs/7188874947/job/19579380791.

   I will submit the fix for the root cause to IDF as well, but would like to merge this first to unblock CI here.
- sh2lib: CONFIG_SPIRAM option defined in sdkconfig.ci, but not valid for all targets: https://github.com/espressif/idf-extra-components/actions/runs/7188874947/job/19579381569#step:4:9587
- json_parser: missing semicolon in test case: https://github.com/espressif/idf-extra-components/actions/runs/7191258936/job/19585662118?pr=283


0xjakob (approved):
Code looks good to me, thanks for the fixes!

PR #291, opened 2024-01-04
fix(catch2): fix missing structure initializer warning

igrr:
In IDF v5.3, esp_console_cmd_t will include the new field, func_w_context. Since cmd_catch2.cpp was a C++ source file, not initializing this field led to a compiler warning. Change the initialization to the style that is backwards compatible and doesn't trigger the warning.


igrr:
@tore-espressif PTAL, this should fix the failing build jobs.

tore-espressif (approved):
Thank you @igrr , LGTM!

PR #292, opened 2024-01-04
esp_tinyusb v1.4.3:  ESP32P4 HS only support

roma-jam:
# Checklist

- [X] _Optional:_ Component contains unit tests
- [ ] CI passing

# Changes
- Added USB OTG PHY HS configuration in KConfig for ESP32P4

Hint: For better experience and verification use TinyUSB v0.15.0~4

### DIFF
--- a/usb/esp_tinyusb/Kconfig
+++ b/usb/esp_tinyusb/Kconfig
@@ -6,6 +6,17 @@ menu "TinyUSB Stack"
         help
             Specify verbosity of TinyUSB log output.
 
+    choice TINYUSB_RHPORT
+        depends on IDF_TARGET_ESP32P4
+        prompt "TinyUSB PHY"
+        default TINYUSB_RHPORT_HS
+        help
+            Allows set the USB PHY Controller for TinyUSB: HS (USB OTG2.0 PHY for HighSpeed)
+             
+        config TINYUSB_RHPORT_HS
+            bool "HS"
+    endchoice
+
     menu "TinyUSB task configuration"
         config TINYUSB_NO_DEFAULT_TASK
             bool "Do not create a TinyUSB task"
--- a/usb/esp_tinyusb/idf_component.yml
+++ b/usb/esp_tinyusb/idf_component.yml
@@ -1,7 +1,7 @@
 ## IDF Component Manager Manifest File
 description: Espressif's additions to TinyUSB
 documentation: "https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/peripherals/usb_device.html"
-version: "1.4.2~2"
+version: "1.4.3"
 url: https://github.com/espressif/idf-extra-components/tree/master/usb/esp_tinyusb
 dependencies:
   idf: '>=5.0' # IDF 4.x contains TinyUSB as submodule
--- a/usb/esp_tinyusb/include/tusb_config.h
+++ b/usb/esp_tinyusb/include/tusb_config.h
@@ -82,7 +82,12 @@ extern "C" {
 #   define CONFIG_TINYUSB_DEBUG_LEVEL 0
 #endif
 
-#define CFG_TUSB_RHPORT0_MODE       OPT_MODE_DEVICE | OPT_MODE_FULL_SPEED
+#ifdef CONFIG_TINYUSB_RHPORT_HS
+#   define CFG_TUSB_RHPORT1_MODE    OPT_MODE_DEVICE | OPT_MODE_HIGH_SPEED
+#else
+#   define CFG_TUSB_RHPORT0_MODE    OPT_MODE_DEVICE | OPT_MODE_FULL_SPEED
+#endif
+
 #define CFG_TUSB_OS                 OPT_OS_FREERTOS
 
 /* USB DMA on some MCUs can only access a specific SRAM region with restriction on alignment.
--- a/usb/esp_tinyusb/CHANGELOG.md
+++ b/usb/esp_tinyusb/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.4.3
+- esp_tinyusb: ESP32P4 HS only support
+
 ## 1.4.2
 
 - MSC: Fix maximum files open


usb/esp_tinyusb/Kconfig
```diff
@@ -6,6 +6,19 @@ menu "TinyUSB Stack"
...
 
+    choice TINYUSB_RHPORT
+        depends on IDF_TARGET_ESP32P4
+        prompt "TinyUSB PHY"
+        default TINYUSB_RHPORT_HS
+        help
+            Based on this parameter the one of two USB PHY will be used for TinyUSB configuration. OTG2.0 PHY for HighSpeed (HS) or OTG1.1 PHY for FullSpeed (FS)
+
+        config TINYUSB_RHPORT_FS
+            bool "FS"
```
tore-espressif, line 17: Nitpick: Could you please also add 'help' section that would contain pin numbers of each PHY?

usb/esp_tinyusb/idf_component.yml
```diff
@@ -1,7 +1,7 @@
 ## IDF Component Manager Manifest File
 description: Espressif's additions to TinyUSB
 documentation: "https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/api-reference/peripherals/usb_device.html"
-version: "1.4.2~2"
+version: "1.4.3"
```
tore-espressif, line 4: Please update the https://github.com/espressif/idf-extra-components/blob/master/usb/esp_tinyusb/CHANGELOG.md too

tore-espressif (commented):
@roma-jam Thanks for the update!

Overall LGTM, let me test on HW and we can go ahead with merging

usb/esp_tinyusb/idf_component.yml
```diff
@@ -1,7 +1,7 @@
...
 url: https://github.com/espressif/idf-extra-components/tree/master/usb/esp_tinyusb
 dependencies:
```
tore-espressif, line 6: After TinyUSB with P4 support is merged, this section must be updated so P4 requires version >=0.15.0~4

See this docs for reference https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/manifest_file.html#matches-and-rules

roma-jam: But the changes doesn't require the new version. They should work with old version also. 
Anyway change the dependencies?

tore-espressif: Initially I had in mind that 
* P4 requires >=v0.15.0.4
* S2/S3 requires >=v0.14.2

But you're right the component-manager will pick the right version, so no need to change. sry

roma-jam: @tore-espressif I revert the dependency version to 0.14.2 and for esp32s3 the fetched version is 0.15.0~4. Works the same. So I keep the` >= v0.14.2` .

usb/esp_tinyusb/Kconfig
```diff
@@ -6,6 +6,19 @@ menu "TinyUSB Stack"
...
+        prompt "TinyUSB PHY"
+        default TINYUSB_RHPORT_HS
+        help
+            Allows set the USB PHY Controller for TinyUSB: HS (USB OTG2.0 PHY for HighSpeed) or FS (USB OTG1.1 PHY for FullSpeed).
+            USB OTG2.0 PHY Controller pad numbers: DM - 50, DP - 51
+            USB OTG1.1 PHY Controller pad numbers: DM - 53, DP - 54
+             
+        config TINYUSB_RHPORT_HS
+            bool "HS"
+    endchoice
```
Dazza0, line 20: Suggest leaving out any references to selection of PHY for now. Our initial only supports the HS OTG controller using the internal UTMI PHY. Once we figure out the PHY selection logic in the future, we can add options such as this.

usb/esp_tinyusb/include/tusb_config.h
```diff
@@ -82,7 +82,14 @@ extern "C" {
 #   define CONFIG_TINYUSB_DEBUG_LEVEL 0
 #endif
 
-#define CFG_TUSB_RHPORT0_MODE       OPT_MODE_DEVICE | OPT_MODE_FULL_SPEED
+#ifdef CONFIG_TINYUSB_RHPORT_HS
+#   define CFG_TUSB_RHPORT1_MODE    OPT_MODE_DEVICE | OPT_MODE_HIGH_SPEED
+#   define CFG_TUSB_EP_SIZE         512
```
Dazza0, line 87: When building an HS device (with HS MPS), do you know if the EP size gets set back to 64 if this HS device is plugged into a FS Host?

IMO it's not super critical since its unlikely our initial demos will be plugged into an FS host, so we could always leave this as a follow up.

Dazza0 (approved):
@roma-jam Left some nitpicks but LGTM otherwise

usb/esp_tinyusb/include/tusb_config.h
roma-jam: That is exactly the question was mentioned a couple of minutes before.
I am going to check it.

roma-jam: @tore-espressif @Dazza0 Yep, to be able to work on both hosts (FS and HS) we should be able to provide a correct configuration descriptor with the EP size, which should be 64 or 512.
Right now, we don't have an option to register two configuration descriptor, only one of them via esp_tinyusb: 
`const uint8_t *configuration_descriptor;   /*!< Pointer to a configuration descriptor. If set to NULL, TinyUSB device will use a default configuration descriptor whose values are set in Kconfig */`

But in P4 we need to register both of them and change the EP size based on the speed.

```
// Invoked when received GET CONFIGURATION DESCRIPTOR
// Application return pointer to descriptor
// Descriptor contents must exist long enough for transfer to complete
uint8_t const * tud_descriptor_configuration_cb(uint8_t index)
{
  (void) index; // for multiple configurations

#if TUD_OPT_HIGH_SPEED
  // Although we are highspeed, host may be fullspeed.
  return (tud_speed_get() == TUSB_SPEED_HIGH) ?  desc_hs_configuration : desc_fs_configuration;
#else
  return desc_fs_configuration;
#endif
}
```
I guess right now I can revert the changes to configure EP size based on speed and leave only the option to set the ep size in examples code. 
After that, we could think/discuss how we want to achieve the goal of config decs for FS and HS: register two const descriptors or just implement a callback.

roma-jam: Yep, otherwise, right now (if always return FS config descriptor with EP size 64) then S3 MSC Host + P4 MSC Device are working in pair.

roma-jam:
Hey @tore-espressif @Dazza0, 
I have resolved the situation with configuration descriptor. I don't thing we are going to add more possibilities of returning the configuration descriptor during the SetConfiguration() standard request, so I decided to add it the same way as it is implemented for current configuration descriptor (which right now can be called FS configuration descriptor).

The API doesn't changed (I left the name of `configuration_descriptor` in `tinyusb_config_t` structure the same). When the HS descriptor has not been provided by user, the default HS descriptor is using (if we have `TUD_OPT_HIGH_SPEED`)

I kept the commit history changes available, just for easier understanding of changes. (I will clean it right before merge)
I will test the changes a bit more, meanwhile looking forward for your reviews.

usb/esp_tinyusb/usb_descriptors.c
```diff
@@ -191,6 +191,33 @@ uint8_t const descriptor_fs_cfg_kconfig[] = {
...
+#endif
+
+#if CFG_TUD_MSC
+    // Interface number, string index, EP Out & EP In address, EP size
+    TUD_MSC_DESCRIPTOR(ITF_NUM_MSC, STRID_MSC_INTERFACE, EPNUM_MSC, 0x80 | EPNUM_MSC, 512),
+#endif
+
+#if CFG_TUD_NCM
+    // Interface number, description string index, MAC address string index, EP notification address and size, EP data address (out, in), and size, max segment size.
+    TUD_CDC_NCM_DESCRIPTOR(ITF_NUM_NET, STRID_NET_INTERFACE, STRID_MAC, (0x80 | EPNUM_NET_NOTIF), 64, EPNUM_NET_DATA, (0x80 | EPNUM_NET_DATA), 512, CFG_TUD_NET_MTU),
```
roma-jam, line 216: I am not sure about the EP_NET_NOTIF EP size for HS. 
(Should it be 64 or 512). Based on the previous code - it doesn't depend on speed, but should be double checked.

tore-espressif:
@roma-jam Please also have a look `tud_descriptor_other_speed_configuration_cb()` API in TinyUSB https://github.com/espressif/tinyusb/blob/master/src/device/usbd.h#L136

And how it is used in tinyusb official examples.

roma-jam:
@tore-espressif The `tud_descriptor_other_speed_configuration_cb()` is another thing, which is not related to the current changes. 
Based on USB specs: _a high-speed capable device supports the
device_qualifier descriptor to return information about the device for the speed at which it is not operating
(including wMaxPacketSize for the default endpoint and the number of configurations for the other speed).
The other_speed_configuration returns information in the same structure as a configuration descriptor, but
for a configuration if the device were operating at the other speed._ (usb_2.0.pdf, 9.4.3 Get Descriptor) 

Anyway, to be able to return response to the `GetConfiguration()` request correctly in both cases we need to keep both FS and HS configuration descriptor somewhere in memory, to be able to return the const pointer in `uint8_t const *tud_descriptor_configuration_cb(uint8_t index)`

Otherwise, we need dynamically provide the correct descriptor.
And yes, that is true that we need to implement the  `tud_descriptor_other_speed_configuration_cb()` for the case, when Host decided to change the speed of the device, because right now we don't have any.

usb/esp_tinyusb/include/tusb_config.h
roma-jam: @tore-espressif @Dazza0 I have removed the HS configuration descriptor from current PR. 

Right now the HS P4 Device won't work in any FS Host port. Compatibility only with HS Hosts.

tore-espressif (approved):

PR #296, opened 2024-01-18
esp_tinyusb v1.4.4: Refactor External USB PHY dependencies

roma-jam:
# Checklist

- [x] CI passing

# Change description
- Add dependency for usb_phy for ESP32S2 and ESP32S3. ESP32P4 doesn't require usb_phy and as the result the usb component itself from esp-idf.

### DIFF
--- a/usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
+++ b/usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
@@ -1,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 2020-2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "esp_err.h"
+#include "esp_private/usb_phy.h"
+#include "soc/usb_pins.h"
+#include "tinyusb.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Configure and install new USB PHY
+ *
+ * @param[in] config tinyusb stack specific configuration
+ */
+esp_err_t tinyusb_ext_phy_new(const tinyusb_config_t *config);
+
+/**
+ * @brief Delete previously installed USB PHY
+ *
+ */
+esp_err_t tinyusb_ext_phy_delete(void);
+
+#ifdef __cplusplus
+}
+#endif
--- a/usb/esp_tinyusb/tinyusb_ext_phy.c
+++ b/usb/esp_tinyusb/tinyusb_ext_phy.c
@@ -1,0 +1,46 @@
+/*
+ * SPDX-FileCopyrightText: 2020-2023 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "tinyusb_ext_phy.h"
+
+static usb_phy_handle_t phy_hdl;
+
+esp_err_t tinyusb_ext_phy_new(const tinyusb_config_t *config)
+{
+    // Configure USB PHY
+    usb_phy_config_t phy_conf = {0};
+
+    phy_conf.controller = USB_PHY_CTRL_OTG;
+    phy_conf.otg_mode = USB_OTG_MODE_DEVICE;
+
+    if (config->external_phy) {
+        // External PHY IOs config
+        usb_phy_ext_io_conf_t ext_io_conf = { 0 };
+        ext_io_conf.vp_io_num = USBPHY_VP_NUM;
+        ext_io_conf.vm_io_num = USBPHY_VM_NUM;
+        ext_io_conf.rcv_io_num = USBPHY_RCV_NUM;
+        ext_io_conf.oen_io_num = USBPHY_OEN_NUM;
+        ext_io_conf.vpo_io_num = USBPHY_VPO_NUM;
+        ext_io_conf.vmo_io_num = USBPHY_VMO_NUM;
+
+        phy_conf.target = USB_PHY_TARGET_EXT;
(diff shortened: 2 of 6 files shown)


usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param[in] config tinyusb stack specific configuration
+ */
+esp_err_t tinyusb_ext_phy_new(const tinyusb_config_t *config);
+
+/**
+ * @brief Delete previously installed USB PHY
+ *
+ */
+esp_err_t tinyusb_ext_phy_delete(void);
```
Dazza0, line 29: @roma-jam Not sure if this is a good idea. We'll end up having two separate PHY APIs (one for host stack, one for device stack).

Is there a reason why we can't use the existing `usb_phy.h` API?

usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
```diff
@@ -0,0 +1,33 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Configure and install new USB PHY
+ *
+ * @param[in] config tinyusb stack specific configuration
+ */
+esp_err_t tinyusb_ext_phy_new(const tinyusb_config_t *config);
```
Dazza0, line 23: Will this API be extensible in the future to support ULPI PHYs for the device stack?

Dazza0 (commented):
@roma-jam I'm not understanding the requirements for this PR. Is there a reason to favor having a separate device stack PHY driver rather than having host/device stack and the USJ driver share the same USB PHY API?

usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
roma-jam: @Dazza0 Actually, the only thing that this changes makes is they collect the `usb_phy_*` calls in a one place (`tinyusb_ext_phy.c`) and add the dependency to build or not to build `usb_phy` based on target (S2/S3 or P4).
There is no changes in API, not implementing new API or anything like this.
And we still using existing usb_phy.h API.

usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
roma-jam: @Dazza0 Probably, yes. I don't know the requirements for the ULPI PHYs, but right now it doesn't matter.
The only difference after this PR is **where** do we need to do this changes to support ULPI PHY: in `tinyusb.c` (current version) or in `tinyusb_ext_phy.c` (this PR)

usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
Dazza0: @roma-jam 

In theory, the P4 should still make a call to `usb_phy.h` API. In fact, USB Host, Device, and USJ should all be making `usb_phy.h` calls on initialization to specify which controller they are, and which PHY they are using.

Future HS capable targets may have multiple possible PHYs (e.g., UTMI vs ULPI), thus we'll need to add a `usb_phy_...` call back in for those targets.

Furthermore, even if the P4 HS only has one possible PHY it can use (i.e., UTMI), it's still good practice to call `usb_phy_new()` even though are no other possible choices for PHY. Because `usb_phy_new()` not only acts as PHY selection, but also PHY initialization. For example, what if before the device stack was initialize, the power management/sleep services ran and turned off the USB PHY clock to save power. Having a `usb_phy_new()` call to turn it back on would be helpful.

usb/esp_tinyusb/include_private/tinyusb_ext_phy.h
Dazza0: The changes will need to be in the `usb_phy.h` API. The `usb_new_phy()` function will need to...

- Select which controller to connect to a PHY (e.g., USB HS, USB FS, USJ)
- Select which PHY interface to use for this controller (FSLS vs UTMI vs ULPI). Note that each controller may only support a subset of these interfaces.
- Whether this interface routes to an internal PHY, or an external PHY. If eternal, users should provide the list of GPIOs that map to each signal of the interface.

The current `usb_phy_config_t` doesn't support the API model described above (thus the need to refactor `usb_phy.h` API.

usb/esp_tinyusb/tinyusb_ext_phy.c
```diff
@@ -0,0 +1,46 @@
...
+
+    if (config->external_phy) {
+        // External PHY IOs config
+        usb_phy_ext_io_conf_t ext_io_conf = { 0 };
+        ext_io_conf.vp_io_num = USBPHY_VP_NUM;
+        ext_io_conf.vm_io_num = USBPHY_VM_NUM;
+        ext_io_conf.rcv_io_num = USBPHY_RCV_NUM;
+        ext_io_conf.oen_io_num = USBPHY_OEN_NUM;
+        ext_io_conf.vpo_io_num = USBPHY_VPO_NUM;
+        ext_io_conf.vmo_io_num = USBPHY_VMO_NUM;
```
Dazza0, line 27: I've already deprecated these GPIO numbers (see description of MR !28293 for more details).

Basically, with an external FSLS PHY, those signals can be routed to any GPIO via the GPIO matrix. Thus the current use of the `USBPHY_..._NUM` macros are forcing users to use an arbitrary set of GPIOs, whereas the cleaner way to do it is to have the users provide the list of GPIOs they want to use via a configuration structure.

roma-jam:
Not relevant anymore, as the changes were made in ESP-IDF (adding usb_phy.c and usb_pins.h).

PR #299, opened 2024-01-22
led_strip: Add support for WS2812B-V5  (IEC-84)

bryghtlabs-richard:
WS2812B-V5 requires >280 microsecond reset time.

When led_strip with 50 microsecond reset time is used with WS2812B-V5 LEDs, most will handle reset correctly, but about 1/3 of certain transmission patterns would cause 1-2 LEDs in a long strip to fail to reset, leading to duplicated colors further down the led strip. Increase the reset time to 280 microseconds, which is compatible with led pixels requiring 50+ microsecond reset time as well.

image

### DIFF
--- a/led_strip/idf_component.yml
+++ b/led_strip/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.5.2"
+version: "2.5.3"
 description: Driver for Addressable LED Strip (WS2812, etc)
 url: https://github.com/espressif/idf-extra-components/tree/master/led_strip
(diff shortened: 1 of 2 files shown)


suda-morris:
@bryghtlabs-richard  Thanks for the improvement!

Could you please help bump up the component to "2.5.3"? https://github.com/espressif/idf-extra-components/blob/master/led_strip/idf_component.yml#L1

igrr (approved):

PR #287, opened 2023-12-27
feat(esp_qemu_rgb_panel): implement 16bpp QEMU RGB panel (IEC-77)

horw:
# Checklist

- [ ] Component contains License
- [ ] Component contains README.md
- [ ] Component contains idf_component.yml file with `url` field defined
- [ ] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [ ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description
_Please describe your change here_

PR #300, opened 2024-01-22
esp_lcd_qemu_rgb minor fixes

igrr:
This PR has small fixes for esp_lcd_qemu_rgb:

- Fixed compatibility with IDF v5.3-dev. In IDF commit ad84b61e490cf1671984e7a52997d5011c5eb7be (MR 28412), esp_lcd_panel_rgb.h is made available only for chip targets with LCD_CAM peripheral. Fix the example project to not include this header, since it's not needed anyway
- Switched the example to use the 16-bit color format introduced in #287. This makes it one less setting to keep in mind, and closer to what most people are going to use.
- Updated the changelog.


horw (approved):
I have tested, and the changes work correctly for me! LGFM

suda-morris (approved):
LGTM

PR #313, opened 2024-03-12
fix(usb) compile tinyusb_net.c when RNDIS is enabled

voedipus:
# Checklist

- [ ] CI passing

# Change description
tinyusb_net.c was not compiled when RNDIS support is enabled


tore-espressif:
@voedipus Does the RNDIS configuration work with your device?

we did not test RNDIS, only NCM. That is why why enabled NCM only

voedipus:
@tore-espressif  Actually it is not working with this MR, but I was able to get RNDIS working with couple of other fixes. I will add changes soon

voedipus:
@tore-espressif  It works now, checked on Linux with esp idf 5.1.3 and Windows 10 (RNDIS)

usb/esp_tinyusb/Kconfig
```diff
@@ -259,12 +259,15 @@ menu "TinyUSB Stack"
...
-
             config TINYUSB_NET_MODE_NCM
                 bool "NCM"
 
+            config TINYUSB_NET_MODE_RNDIS
+                bool "RNDIS"
+
+            config TINYUSB_NET_MODE_ECM
+                bool "ECM"
+
```
tore-espressif, line 270: Can we keep it tha way it was? ECM+RNDIS together? There are 2 reasons for this:

1. The TinyUSB driver has ECM+RNDIS protocol together
2. We usually don't want to enable ECM and disalble RNDIS (or vice versa). RNDIS works with windows, while ECM workds with linux and macOS. Only with both we can support all operating systems

tore-espressif (commented):
Thank you very much for this PR!

I'm still not sure whether we want to claim support form ECM/RNDIS when we already support NCM. I'll check internally and let you know

wuyuanyi135:
> @voedipus Thank you for the PR! Does the RNDIS configuration work with your device?
> 
> we did not test RNDIS, only NCM. That is why why enabled NCM only

Please do include RNDIS support. The upstream is [rewriting](https://github.com/hathach/tinyusb/pull/2227) the NCM driver and some unstablility will be expected. Also, the current NCM driver is buggy and [slower than the RNDIS](https://github.com/hathach/tinyusb/issues/2068). If we would like to enable the USB network feature at this moment I think RNDIS is still the way to go

lijunru-hub:
The current RNDIS driver in TinyUSB does not support FIFO buffering and lacks an effective API to notify the application layer when the previous transmission is complete. Using the more efficient `tinyusb_net_send_async` function often results in cases where the previous RNDIS transmission hasn't finished. On the other hand, the new NCM driver supports software buffering. I believe the RNDIS driver should also support this feature.

Furthermore, supporting RNDIS is necessary because, in network card applications, Windows 10 does not support the NCM driver.

tore-espressif:
> Furthermore, supporting RNDIS is necessary because, in network card applications, Windows 10 does not support the NCM driver.

Here's a step-by-step manual how to install NCM driver on Windows 10 , if anyone is interested

[win10-native-ncm.pdf](https://github.com/user-attachments/files/16905045/win10-native-ncm.pdf)

PR #318, opened 2024-04-09
fix(ccomp_timer): require perfmon only on xtensa

fhrbata:
The perfmon component is supported[1] only on xtensa arch, so it should not be in requires for other archs.

[1] https://github.com/espressif/esp-idf/blob/master/components/perfmon/CMakeLists.txt#L9-L11

Fixes: 5d61b5047f37 ("test_utils: implement performance timer")

### DIFF
--- a/ccomp_timer/CMakeLists.txt
+++ b/ccomp_timer/CMakeLists.txt
@@ -1,3 +1,5 @@
+idf_build_get_property(arch IDF_TARGET_ARCH)
+
 set(srcs "ccomp_timer.c")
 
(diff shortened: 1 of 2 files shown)


igrr (approved):

igrr:
@fhrbata LGTM! Could you please also bump the version in ccomp_timer/idf_component.yml so that this fix gets released?

fhrbata:
> @fhrbata LGTM! Could you please also bump the version in ccomp_timer/idf_component.yml so that this fix gets released?

@igrr I'm sorry, but I somehow missed your comment. It should be done. Thank you

PR #319, opened 2024-04-16
fix(freetype): Remove system lib dependencies (IEC-103)

espressif2022:
# Checklist

- [ ] CI passing

# Change description

1. Remove system lib dependencies: If the package required by `find_package` (such as `BrotliDec`, `HarfBuzz`) exists locally on Linux, the local lib will be used, causing compilation failure.
`-- Found BrotliDec: /usr/include (found version "1.0.9")`
`-- HarfBuzz (required): /usr/lib/x86_64-linux-gnu/libharfbuzz.so`
`/usr/include/stdlib.h:258:25: error: '_Float64x' is not supported on this target`

### DIFF
--- a/freetype/CMakeLists.txt
+++ b/freetype/CMakeLists.txt
@@ -1,7 +1,7 @@
 idf_component_register()
 
+set(CMAKE_SYSTEM_IGNORE_PATH "/usr/include;/usr/lib/x86_64-linux-gnu")
 set(BUILD_SHARED_LIBS OFF)
-option(BUILD_TESTING OFF)
 
 function(install)
 endfunction()
--- a/freetype/idf_component.yml
+++ b/freetype/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.13.0~1"
+version: "2.13.0~2"
 description: freetype C library
 url: https://github.com/espressif/idf-extra-components/tree/master/freetype
 repository: "https://github.com/espressif/idf-extra-components.git"


igrr:
@espressif2022 Could you please update the description, which issue does the MR fix? Is it a Linux-specific issue? I wonder why we didn't encounter it when this component gets built in CI.

espressif2022:
> @espressif2022 Could you please update the description, which issue does the PR fix? Is it a Linux-specific issue? I wonder why we didn't encounter it when this component gets built in CI.

This is related to the system version. For recently installed ubuntu, `find_package` will find the following system library problems, resulting in failed compilation, check that the package does exist locally. 
If it does not exist in the local system, it can be compiled OK. I can't overwrite the option in sub cmake, so I want to ignore the system path here.
`-- Found BrotliDec: /usr/include (found version "1.0.9")`
`-- HarfBuzz (required): /usr/lib/x86_64-linux-gnu/libharfbuzz.so`

igrr:
I see, thanks for explaining. I can reproduce the issue on macOS as well now, provided I install these libraries. I suppose we will need a more general solution since we can't possibly know all the possible include paths on all the systems. 

I'm okay with merging this workaround first, though.

igrr (approved):

igrr:
@espressif2022 I've opened https://github.com/espressif/idf-extra-components/pull/320 as a slightly more future-proof solution to the external dependencies issue.

Issue #324, opened 2024-05-01
CoAP: Support for Q-Block options (RFC9177) is always disabled


sfzhi:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

coap

### ESP-IDF version.

5.2.1

### Development Kit.

Seeed XIAO ESP32C3

### Used Component version.

4.3.4~1

### More Information.

The [component page](https://components.espressif.com/components/espressif/coap) lists RFC9177 as supported, but in reality that is not the case. The corresponding upstream libcoap version does support it, but that support is always disabled in the ESP-IDF component. I can see the relevant macro (`COAP_Q_BLOCK_SUPPORT`) is defined in `port/include/coap_config.h`, but the corresponding configuration option is not present in the `Kconfig` file, so the functionality cannot be enabled (at least not without using some dirty tricks to inject the macro when building the component).

PR #325, opened 2024-05-06
fix(coap): Add config option to enable/disable Q-Block option

hmalpani:
Add config option to enable/disable Q-Block option

Fixes: https://github.com/espressif/idf-extra-components/issues/324

### DIFF
--- a/coap/Kconfig
+++ b/coap/Kconfig
@@ -102,6 +102,12 @@ menu "CoAP Configuration"
 
             If this option is disabled, redundant CoAP Observe Persist code is removed.
 
+    config COAP_Q_BLOCK
+        bool "Enable Q-Block (RFC9177) support in CoAP"
+        default n
+        help
+            Enable Q-Block (RFC9177) fast block transfers using NON (instead of CON) support for CoAP
+
     config COAP_WEBSOCKETS
         bool "Enable WebSockets support within CoAP"
         default n
--- a/coap/idf_component.yml
+++ b/coap/idf_component.yml
@@ -1,4 +1,4 @@
-version: "4.3.4~1"
+version: "4.3.4~2"
 description: Constrained Application Protocol (CoAP) C Library
 url: https://github.com/espressif/idf-extra-components/tree/master/coap
 dependencies:


mahavirj:
@mrdeep1 

- Can you please take a look at this PR?
- When can we expect new bugfix release of libcoap containing the fix for `CVE-2024-0962`?

coap/Kconfig
```diff
@@ -102,6 +102,12 @@ menu "CoAP Configuration"
 
             If this option is disabled, redundant CoAP Observe Persist code is removed.
 
+    config COAP_Q_BLOCK
+        bool "Enable block-wise transfer in CoAP"
```
mrdeep1, line 106: bool "Enable Q-Block (RFC9177) support in CoAP"

Note: change as regular block transfers (RFC7959) is always enabled

coap/Kconfig
```diff
@@ -102,6 +102,12 @@ menu "CoAP Configuration"
...
+        default n
+        help
+            Enable Block-Wise Transfers support for CoAP
```
mrdeep1, line 109: Enable Q-Block (RFC9177) fast block transfers using NON (instead of CON) support for CoAP

mrdeep1 (changes requested):
Q-Block supports fast block based transfer / recovery when using NON PDU type (instead of CON) (RFC9177),  This is in addition to block transfer CON support in RFC7959 which is always enabled in CoAP.

mrdeep1:
> When can we expect new bugfix release of libcoap containing the fix for CVE-2024-0962?

There is a tagged libcoap release that has this fix in - which I can use to update idf-extra-components.  In the near term, we are planning on getting out a Release Candidate which covers this CVE with other updates as well which I will be putting into idf-extra-components.

Do you want me to update the code with the tagged fix?

mahavirj:
> > When can we expect new bugfix release of libcoap containing the fix for CVE-2024-0962?
> 
> There is a tagged libcoap release that has this fix in - which I can use to update idf-extra-components. In the near term, we are planning on getting out a Release Candidate which covers this CVE with other updates as well which I will be putting into idf-extra-components.
> 
> Do you want me to update the code with the tagged fix?

I thought upgrade to official `4.3.5` would be a cleaner solution. It will also help to align our component version.

At the moment, I see only `4.3.4a` tag here https://github.com/obgm/libcoap/tags. If we plan to upgrade to this tag, then I am not sure what will be the corresponding IDF component version.

mrdeep1:
Yes, `4.3.4a` contains the CVE fix.  I had assumed that `coap/idf_component.yml` version would become "4.3.4~2", but I guess something else might not like `4.3.4a`.

Certainly` 4.3.5` would be cleaner.

coap/Kconfig
mrdeep1: Change is good

coap/Kconfig
mrdeep1: Change is good

mahavirj (approved):

PR #329, opened 2024-05-17
fix(onewire): raise recovery time to support more sensor on longer wire (IEC-111)

muggenhor:
# Change description

Specifically this is the time that's necessary to be able to perform a device search on a bus consisting of >= 6 devices on >= 10m of wire at 5V with a 4.7 kΩ pull-up resistor.


muggenhor:
I'm not signing a contract with a company I don't know (the CLA). I also don't believe this signal change in number is copyrightable (not an "original work"). So feel free to regard this PR as a bug report with suggested fix direction instead.

suda-morris (approved):
LGTM

suda-morris:
I will bump up the version of onewire component later

muggenhor:
Thanks for merging this! :)

PR #320, opened 2024-04-19
FreeType: improve library dependency handling, add example

igrr:
- ~~Updated Zlib library to be compatible with CMake `FindPackage` module. Now, third party libraries which call`find_package(ZLIB)` will find `espressif/zlib` component automatically, provided that it has been added as a dependency into the project.~~
  - ~~We should do the same in other common components, as well.~~
  - ~~Updated `freetype` component to have a dependency on `zlib`~~
  - Will do the above in another PR

- Configured `freetype` to not look for Brotli, Harfbuzz, BZ2, libpng, zlib libraries in the system (https://github.com/espressif/idf-extra-components/issues/317)
- Added a simple example to `freetype` component to validate its operation, added pytest script

PR #332, opened 2024-05-20
thorvg: add component of thorvg (IEC-115)

espressif2022:
# Checklist

- [x] CI passing

# Change description
1: Add component of [thorvg](https://github.com/thorvg/thorvg).
2: Example is [here](https://github.com/espressif2022/esp-bsp/tree/example/add_thorvg/examples/display_thorvg) .
3: The structure refers to this [PR](https://github.com/lvgl/lvgl/pull/6081) of LVGL.


suda-morris:
@espressif2022 instead of copying the thorvg source code, can we use git submodule?

thorvg/thorvg/add_lvgl_if.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
```
suda-morris, line 2: this is not part of the thorvg library I think

suda-morris:
can we also provide an example for this?

thorvg/idf_component.yml
```diff
@@ -0,0 +1,8 @@
+version: "0.11.9"
+description: graphics library
```
igrr, line 2:
```suggestion
description: "ThorVG is an open-source graphics library designed for creating vector-based scenes and animations"
```

we can use this sentence from ThorVG readme.

igrr:
LGTM overall, aside from the points mentioned above.

Please also add an SBOM file, there is an example in `fmt` component.

igrr:
Regarding an example, I agree it would be useful. You can do something very simple which only shows the API usage. For instance, see this other PR where I'm adding a FreeType example: https://github.com/espressif/idf-extra-components/pull/320.

espressif2022:
> @espressif2022 instead of copying the thorvg source code, can we use git submodule?

ok, modify this in my fork repository, and then use submodule here.

espressif2022:
> Regarding an example, I agree it would be useful. You can do something very simple which only shows the API usage. For instance, see this other PR where I'm adding a FreeType example: #320.

Okay, I'll add an example referring to this.

igrr:
@espressif2022 regarding this comment,

> ok, modify this in my fork repository, and then use submodule here.

What changes did you have to make to the upstream project, could you please show a diff? Thorvg can usually be ported to different platforms without modifying its source code, so I wonder if we could do the same here.

espressif2022:
> @espressif2022 regarding this comment,
> 
> > ok, modify this in my fork repository, and then use submodule here.
> 
> What changes did you have to make to the upstream project, could you please show a diff? Thorvg can usually be ported to different platforms without modifying its source code, so I wonder if we could do the same here.

Mainly cmake compilation changes, because thorvg is built with meson and should be incompatible with our ESP-IDF.
So here we need to use the original thorvg repository?

igrr:
Yeah, using the original repo is the best option. This will make releaseing new thorvg versions simpler, since we will only need to update the submodule pointer.

Can you describe what these "cmake compilation changes" were? As I understand, you have added component CMakeLists.txt, which is outside of thorvg directory. So is there anything extra that needs to be done inside thorvg itself?

espressif2022:
> Yeah, using the original repo is the best option. This will make releasing new thorvg versions simpler, since we will only need to update the submodule pointer.
> 
> Can you describe what these "cmake compilation changes" were? As I understand, you have added component CMakeLists.txt, which is outside of thorvg directory. So is there anything extra that needs to be done inside thorvg itself?

The internal build system uses Meson: [meson.build](https://github.com/thorvg/thorvg/blob/92e1dd67477495c5b7a38759bd286e58b4c5b3c1/meson.build#L1). 
I see that lvgl's third-party library directly changes meson to cmake, which may be simpler.
Is it possible to be compatible with meson here? 

Additionally, some internal function names are causing compilation issues(**_my self-test seems to be fine_**), as mentioned here: [Issue 2062](https://github.com/thorvg/thorvg/issues/2062).

espressif2022:
> 2\. ExternalProject_Add

I tried submodule + ExternalProject_Add and the compilation was ok.
Let me improve it，there are some configs to see how to pass in the thorvg configuration.

espressif2022:
> Both options are okay:
> 
> 1. Add the original repo as a submodule, then add component CMakeLists.txt file as you already did in this PR. Since the CMakeLists.txt file is outside of the thorvg submodule directory, you don't need to modify the submodule this way.
> 2. Add the original repo as a submodule, then add component CMakeLists.txt file, then use ExternalProject_Add to perform the build of thorvg using `meson`.
> 
> Let me know if you meet any issue or need any help.

I tried submodule + ExternalProject_Add and the compilation was ok.
Let me improve it，there are some configs to see how to pass in the thorvg configuration.

igrr:
I think you can use CMake [`file(WRITE...)` command](https://cmake.org/cmake/help/v3.13/command/file.html#write) to generate a [Meson cross-compilation config file](https://mesonbuild.com/Cross-compilation.html#cross-compilation) inside the build directory, then pass the path of this file to Meson.

espressif2022:
> I think you can use CMake [`file(WRITE...)` command](https://cmake.org/cmake/help/v3.13/command/file.html#write) to generate a [Meson cross-compilation config file](https://mesonbuild.com/Cross-compilation.html#cross-compilation) inside the build directory, then pass the path of this file to Meson.

I tested to use a fixed [config file](https://github.com/espressif2022/idf-extra-components_fork/blob/9b98900ad2414615dd2f49c82f04fd2ffb786f4f/thorvg/cross_file.txt#L2) .

`CONFIGURE_COMMAND meson setup ${THORVG_BUILD_DIR} ${THORVG_SUBDIR} --cross-file ${CROSS_FILE} -Dbindings=capi -Ddefault_library=static`

I encountered a **strdup** error, despite having included **cstring** in the header file. 
I'm not sure if it's an issue with our environment's standard library, I see the cross file example has [this.](https://github.com/thorvg/thorvg/blob/5c8b68c2206d99d922dc7fd02e294eac4b9f0877/cross/android_x86_64.txt#L13)

```
../../components/espressif__thorvg/thorvg/src/renderer/tvgText.h:65:32: error: 'strdup' was not declared in this scope; did you mean 'strcmp'?
   65 |         if (utf8) this->utf8 = strdup(utf8);
```

igrr:
Can you try adding -D_GNU_SOURCE to the compilation flags of thorvg? (`c_args` option in meson)
I think this should make strdup visible: https://github.com/espressif/newlib-esp32/blob/7cdcb0b0848a5d909e906834fe81903ba9466c90/newlib/libc/include/string.h#L83-L85

espressif2022:
> @espressif2022 instead of copying the thorvg source code, can we use git submodule?

replaced now.

espressif2022:
> can we also provide an example for this?

Added, riscv32 in P4 and xtensa in s3 are tested.

espressif2022:
> Can you try adding -D_GNU_SOURCE to the compilation flags of thorvg? (`c_args` option in meson) I think this should make strdup visible: https://github.com/espressif/newlib-esp32/blob/7cdcb0b0848a5d909e906834fe81903ba9466c90/newlib/libc/include/string.h#L83-L85

Both riscv32 on P4 and xtensa on s3 have been tested.

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,151 @@
...
+
+static void init_filesystem(void);
+static void capi_shape_example(void);
+static void capi_lottie_example(void);
+
+#define LOTTIE_SIZE_HOR     300
+#define LOTTIE_SIZE_VER     300
+
+void app_main(void)
+{
```
suda-morris, line 26: should we initialize the LCD panel as well?

espressif2022: > should we initialize the LCD panel as well?

OK, I ran it on P4 and S3 yesterday. I didn't put it here, I'll add it.

thorvg/examples/thorvg-example/sdkconfig.defaults
```diff
@@ -0,0 +1,17 @@
+# This file was generated using idf.py save-defconfig. It can be edited manually.
+# Espressif IoT Development Framework (ESP-IDF) Project Minimal Configuration
+#
+CONFIG_IDF_TARGET="esp32p4"
```
suda-morris, line 4: this doesn't have to be in the sdkconfig.defaults

thorvg/examples/thorvg-example/sdkconfig.defaults
```diff
@@ -0,0 +1,17 @@
...
+
+CONFIG_ESPTOOLPY_FLASHMODE_QIO=y
+CONFIG_ESPTOOLPY_FLASHSIZE_4MB=y
+CONFIG_PARTITION_TABLE_CUSTOM=y
+CONFIG_COMPILER_OPTIMIZATION_PERF=y
+CONFIG_COMPILER_CXX_EXCEPTIONS=y
+CONFIG_SPIRAM=y
+CONFIG_SPIRAM_SPEED_200M=y
```
suda-morris, line 12: they're target specific, add them in the sdkconfig.defaults.esp32p4 and sdkconfig.defaults.esp32s3

thorvg/examples/thorvg-example/sdkconfig.bsp.esp32_s3_lcd_ev_board
```diff
@@ -0,0 +1,15 @@
...
+CONFIG_IDF_TARGET="esp32s3"
+
+CONFIG_ESPTOOLPY_FLASHMODE_QIO=y
+CONFIG_ESPTOOLPY_FLASHSIZE_4MB=y
+CONFIG_PARTITION_TABLE_CUSTOM=y
+CONFIG_COMPILER_OPTIMIZATION_PERF=y
+CONFIG_COMPILER_CXX_EXCEPTIONS=y
+CONFIG_SPIRAM=y
+CONFIG_SPIRAM_MODE_OCT=y
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=50000
```
suda-morris, line 13: duplicated with sdkconfig.defaults

thorvg/examples/thorvg-example/sdkconfig.bsp.esp32_p4_function_ev_board
```diff
@@ -0,0 +1,17 @@
+# This file was generated using idf.py save-defconfig. It can be edited manually.
+# Espressif IoT Development Framework (ESP-IDF) Project Minimal Configuration
+#
+CONFIG_IDF_TARGET="esp32p4"
+
```
suda-morris, line 5: I wonder if we can rename the `sdkconfig.bsp.esp32_p4_function_ev_board` to `sdkconfig.defaults.esp32p4` because the default configs list here are not specific the **p4 function ev board**, other esp32p4 board can also use it.

thorvg/examples/thorvg-example/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# For more information about build system see
+# https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/build-system.html
+# The following five lines of boilerplate have to be in your project's
+# CMakeLists in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.16)
+
+# set(COMPONENTS main) # "Trim" the build. Include the minimal set of components; main and anything it depends on.
```
suda-morris, line 7: remove it or uncomment it

thorvg/examples/thorvg-example/main/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+idf_component_register(
+    SRCS "display_main.c"
+    INCLUDE_DIRS "."
+    PRIV_REQUIRES spiffs)
+
+set_source_files_properties(
+    "display_main.c"
+    PROPERTIES COMPILE_OPTIONS
+    "-Wno-strict-prototypes")
```
suda-morris, line 9: I wonder if we can remove this if we upgrade the thorvg to 0.13.6? See: https://github.com/thorvg/thorvg/compare/v0.13.5...v0.13.6 and https://github.com/thorvg/thorvg/commit/946abbddf2b396aea4fccee944e41fd5745db809

thorvg/README.md
```diff
@@ -0,0 +1,9 @@
+# thorvg component
```
suda-morris, line 1:
```suggestion
# ThorVG component
```

thorvg/examples/thorvg-example/main/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+dependencies:
+  espressif/esp32_p4_function_ev_board:
```
suda-morris, line 2: can we use esp32_p4_function_ev_board_noglib instead? because this example doesn't need a LVGL port

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,173 @@
...
+    uint32_t *cavas_buf_888 = NULL;
+    uint16_t *cavas_buf_565 = NULL;
+
+    cavas_buf_888 = heap_caps_malloc(LOTTIE_SIZE_HOR * LOTTIE_SIZE_VER * sizeof(uint32_t), MALLOC_CAP_SPIRAM);
+    assert(cavas_buf_888);
+
+    cavas_buf_565 = heap_caps_malloc(LOTTIE_SIZE_HOR * LOTTIE_SIZE_VER * sizeof(uint16_t), MALLOC_CAP_SPIRAM);
+    assert(cavas_buf_565);
+
+    assert(tvg_engine_init(TVG_ENGINE_SW, 0) == TVG_RESULT_SUCCESS);
```
suda-morris, line 57: Please don't call a function inside the `assert`. If `NDEBUG` is enabled, `tvg_engine_init` won't even get run

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,173 @@
...
+    if (cavas_buf_888) {
+        free(cavas_buf_888);
+    }
+
+    if (cavas_buf_565) {
+        free(cavas_buf_565);
+    }
+}
+
+static void argb888_to_rgb565(const uint32_t *argb888, uint16_t *rgb565, size_t size)
```
suda-morris, line 159: we can use PPA to convert the ARGB888 to RGB888/565. Not a blocker of this PR.

thorvg/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+endif()
+
+file(APPEND ${TVG_CROSS_CFG} "
+[host_machine]
+system = 'android'
+cpu_family = '${CMAKE_SYSTEM}'
+cpu = 'esp'
+endian = 'little'
+")
+
```
suda-morris, line 79: idf_build_get_property(python PYTHON)

thorvg/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+execute_process(
+    COMMAND pip install meson
```
suda-morris, line 81:
```suggestion
    COMMAND ${python} -m pip install -U meson
```

thorvg/examples/thorvg-example/main/CMakeLists.txt
espressif2022: > I wonder if we can remove this if we upgrade the thorvg to 0.13.6? See: [thorvg/thorvg@v0.13.5...v0.13.6](https://github.com/thorvg/thorvg/compare/v0.13.5...v0.13.6) and [thorvg/thorvg@946abbd](https://github.com/thorvg/thorvg/commit/946abbddf2b396aea4fccee944e41fd5745db809)

Yes, it fixed this warning, I update it now.

thorvg/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+    -Dbindings=capi
+    -Dextra=
+    -Ddefault_library=static
+    -Dthreads=false
+    BUILD_COMMAND ninja -C ${TVG_BUILD_DIR}
+    INSTALL_COMMAND ""
+    BUILD_BYPRODUCTS ${TVG_BUILD_DIR}/src/${TVG_LIB}.a
+)
+
+add_prebuilt_library(${TVG_LIB} ${TVG_BUILD_DIR}/src/${TVG_LIB}.a PRIV_REQUIRES esp_rom)
```
suda-morris, line 105: esp_rom should be common dependency, do we really need to specify it again?

suda-morris (changes requested):
LGTM overall

thorvg/examples/thorvg-example/main/CMakeLists.txt
suda-morris: when you upgrade the thorvg, don't forget to update the sbom as well

thorvg/CMakeLists.txt
espressif2022: When I compile with **xtensa**, it prompts **undefined reference to `__wrap_longjmp'**. 
I added the rom dependency here and it worked.

igrr: This line looks correct to me. Prebuilt libraries by default do not have all the common dependencies.

thorvg/sbom_thorvg.yml
```diff
@@ -0,0 +1,7 @@
+name: thorvg
+version: 0.13.5
```
suda-morris, line 2: 0.13.6

thorvg/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+
+if(NOT MESON_LINKER_FLAGS STREQUAL "''")
+    file(APPEND ${TVG_CROSS_CFG} "
+cpp_link_args = [${MESON_LINKER_FLAGS}]
+")
+endif()
+
+file(APPEND ${TVG_CROSS_CFG} "
+[host_machine]
+system = 'android'
```
suda-morris, line 75: Interesting, why set the host to android?

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,221 @@
...
+        tvg_res = tvg_canvas_draw(canvas);
+        ESP_GOTO_ON_FALSE(tvg_res == TVG_RESULT_SUCCESS, ESP_ERR_INVALID_STATE, err, TAG, "tvg_canvas_draw failed");
+
+        tvg_res = tvg_canvas_sync(canvas);
+        ESP_GOTO_ON_FALSE(tvg_res == TVG_RESULT_SUCCESS, ESP_ERR_INVALID_STATE, err, TAG, "tvg_canvas_sync failed");
+
+        argb888_to_rgb565_ppa(ppa_handle, canvas_buf_888, canvas_buf_565);
+        esp_lcd_panel_draw_bitmap(lcd_panel->panel, 0, 0, LOTTIE_SIZE_HOR, LOTTIE_SIZE_VER, canvas_buf_565);
+
+        vTaskDelay(pdMS_TO_TICKS(1));
```
suda-morris, line 162: so the RTOS tick rate will affect the animation frame rate. It's not easy for the user to tune. maybe we can use esp_timer to create a periodic job?

Not a blocker of this PR BTW.

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,221 @@
...
+    Tvg_Result tvg_res = TVG_RESULT_SUCCESS;
+    Tvg_Result tvg_engine = TVG_RESULT_INSUFFICIENT_CONDITION;
+
+    static uint32_t reac_color = 0x00;
+
+    char *src = "/spiffs/lolo_walk.json";
+
+    uint32_t *canvas_buf_888 = NULL;
+    uint16_t *canvas_buf_565 = NULL;
+    ppa_client_handle_t ppa_handle = NULL;
```
suda-morris, line 69: the PPA client is not a resource should be created and free in a loop, we can create it in the app_main or in the begining of cpi_loop_task

suda-morris (approved):

thorvg/examples/thorvg-example/main/idf_component.yml
```diff
@@ -0,0 +1,8 @@
+dependencies:
+  esp32_p4_function_ev_board_noglib:
+    version: ">=1.2.0"
+    rules:
+      - if: "target == esp32p4"
```
igrr, line 5: What happens when this example is used for a different target?

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,221 @@
...
+
+static esp_err_t capi_create_lottie(bsp_lcd_handles_t *lcd_panel)
+{
+    esp_err_t ret = ESP_OK;
+    Tvg_Result tvg_res = TVG_RESULT_SUCCESS;
+    Tvg_Result tvg_engine = TVG_RESULT_INSUFFICIENT_CONDITION;
+
+    static uint32_t reac_color = 0x00;
+
+    char *src = "/spiffs/lolo_walk.json";
```
igrr, line 65:
```suggestion
    const char *src = "/spiffs/lolo_walk.json";
```

should cause a -Wwrite-strings warning otherwise.

thorvg/CMakeLists.txt
espressif2022: > Interesting, why set the host to android?

```
if get_option('threads') == true and host_machine.system() != 'windows' and host_machine.system() != 'android'
thread_dep = meson.get_compiler('cpp').find_library('pthread')
thorvg_lib_dep += [thread_dep]
endif
```

To solve the `thread_dep` error, but later added `-Dthreads=false`, I think it is better to write `freertos` here now.

thorvg/examples/thorvg-example/main/display_main.c
espressif2022: > so the RTOS tick rate will affect the animation frame rate. It's not easy for the user to tune. maybe we can use esp_timer to create a periodic job?
> 
> Not a blocker of this PR BTW.

OK, I'll update it.

thorvg/examples/thorvg-example/main/display_main.c
espressif2022: > the PPA client is not a resource should be created and free in a loop, we can create it in the app_main or in the begining of cpi_loop_task

ok.

thorvg/CMakeLists.txt
igrr: We do provide pthreads library, so enabling thread support should work fine. What was the exact error you got here?

thorvg/examples/thorvg-example/main/idf_component.yml
espressif2022: > What happens when this example is used for a different target?

It will compile and report an error.
Because `PPA` has been added here, and only the BSP of `esp32_p4_function_ev_board` has the `bsp_display_new_with_handles` function, so the example only supports P4.

thorvg/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+        if(NOT ${index} EQUAL ${last_index})
+            string(APPEND MESON_LINKER_FLAGS ", ")
+        endif()
+    endforeach()
+else()
+    string(APPEND MESON_LINKER_FLAGS "''")
+endif()
+
+string(LENGTH "${CMAKE_CXX_COMPILER}" path_length)
+math(EXPR extract_length "${path_length} - 3")
```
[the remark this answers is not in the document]
igrr: I think you can use the `_CMAKE_TOOLCHAIN_PREFIX` global variable defined in IDF: https://github.com/espressif/esp-idf/blob/c7bbfaee2592fe28dea7700f50cee7c062a51595/tools/cmake/toolchain-esp32.cmake#L8

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,221 @@
...
+        .in.pic_h = LOTTIE_SIZE_VER,
+        .in.block_w = LOTTIE_SIZE_HOR,
+        .in.block_h = LOTTIE_SIZE_VER,
+        .in.block_offset_x = 0,
+        .in.block_offset_y = 0,
+        .in.srm_cm = PPA_SRM_COLOR_MODE_ARGB8888,
+
+        .out.buffer = out,
+        .out.buffer_size = LOTTIE_SIZE_HOR * LOTTIE_SIZE_VER * sizeof(uint16_t),
+            .out.pic_w = LOTTIE_SIZE_HOR,
```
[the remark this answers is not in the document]
espressif2022: > Please install the pre-commit hook and let it format this file, this should fix the failing CI job.

There's pre_commit locally, but commit doesn't seem to be triggered. I'll fix this.

thorvg/CMakeLists.txt
espressif2022: > We do provide pthreads library, so enabling thread support should work fine. What was the exact error you got here?

```
Project name: thorvg
Project version: 0.13.6
C++ compiler for the host machine: /home/xuxin/.espressif/tools/riscv32-esp-elf/esp-13.2.0_20240530/riscv32-esp-elf/bin/riscv32-esp-elf-g++ (gcc 13.2.0 "riscv32-esp-elf-g++ (crosstool-NG esp-13.2.0_20240530) 13.2.0")
C++ linker for the host machine: /home/xuxin/.espressif/tools/riscv32-esp-elf/esp-13.2.0_20240530/riscv32-esp-elf/bin/riscv32-esp-elf-g++ ld.bfd 13.2.0
C++ compiler for the build machine: ccache c++ (gcc 11.4.0 "c++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0")
C++ linker for the build machine: c++ ld.bfd 2.38
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: Generic
Host machine cpu: esp
Target machine cpu family: Generic
Target machine cpu: esp
Configuring config.h using configuration

../../../../thorvg/src/meson.build:47:43: ERROR: C++ shared or static library 'pthread' not found

A full log can be found at /home/xuxin/esp_work/idf-extra-components_fork/thorvg/examples/thorvg-example/build/thorvg_build/meson-logs/meson-log.txt
```

igrr: I see, this should be possible to solve by adding pthread component library to the linker flags.

We can keep this as a follow-up, I think it's okay to not fix this in the current MR. (Provided that thorvg is used from one FreeRTOS task only.)

thorvg/CMakeLists.txt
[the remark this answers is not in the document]
espressif2022: Now use `_CMAKE_TOOLCHAIN_PREFIX` + `tool_name` to replace` _CMAKE_TOOLCHAIN_PREFIX` +` g++`.
Both window and linux, xtensa and riscv32 are supported.

thorvg/examples/thorvg-example/main/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+idf_component_register(
+    SRCS "display_main.c"
+    INCLUDE_DIRS "."
+    PRIV_REQUIRES esp_lcd esp_timer esp_driver_ppa)
+
+idf_component_get_property(board_lib espressif__esp32_p4_function_ev_board_noglib COMPONENT_LIB)
```
suda-morris, line 6: @espzav will you fix the build warning?

espzav: What warning is there? I will try it on the board tomorrow.

thorvg/examples/thorvg-example/main/display_main.c
```diff
@@ -0,0 +1,315 @@
...
+
+    uint32_t *canvas_buf_888 = NULL;
+    uint16_t *canvas_buf_565 = NULL;
+
+    Tvg_Animation *animation = NULL;
+    Tvg_Canvas *canvas = NULL;
+    esp_timer_handle_t play_timer = NULL;
+
+    play_tick_new(&play_timer);
+    ESP_GOTO_ON_FALSE(play_timer, ESP_ERR_INVALID_STATE, err, TAG, "Error create play_timer");
```
suda-morris, line 103: because there're ESP_ERROR_CHECK in the `play_tick_new`, so when that function returns successfully, the `play_timer` must be non-null.

thorvg/examples/thorvg-example/main/CMakeLists.txt
espressif2022: > What warning is there? I will try it on the board tomorrow.

@espzav , this warning.
```
espressif__esp32_p4_function_ev_board_noglib/esp32_p4_function_ev_board.c:33:31: warning: 'tp' defined but not used [-Wunused-variable]
   33 | static esp_lcd_touch_handle_t tp;   // LCD touch handle
```

thorvg/examples/thorvg-example/spiffs_content/lolo_walk.json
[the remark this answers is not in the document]
espressif2022: Delete the useless json now.

PR #338, opened 2024-06-06
feat(coap): libcoap 4.3.5rc1 code (IEC-119)

mrdeep1:
# Checklist

- [ ] CI passing

# Change description
Add in support for libcoap version 4.3.5rc1.


coap/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "4.3.4~3"
+version: "4.3.4~4"
```
hmalpani, line 1: Thanks @mrdeep1 for the PR

I think we should change it to `4.3.5-rc1`. I think component manager will accept that version. Using 4.3.5-rc1 will make it consistent with upstream.

mrdeep1: What goes in `coap/sbom_libcoap.yml` for the version?
````
name: libcoap
version: 4.3.4
cpe: cpe:2.3:a:libcoap:libcoap:{}:*:*:*:*:*:*:*
supplier: 'Organization: libcoap <https://libcoap.net/>'
description: A CoAP (RFC 7252) implementation in C
url: https://github.com/obgm/libcoap
hash: 0392d022173516b6d33a95b8285ba3cbc8767bdf
````

Comment (mrdeep1) [reply coap/idf_component.yml]: Updated `coap/idf_component.yml `only to see what happens.

Comment (hmalpani) [reply coap/idf_component.yml]: I think we will need to update the **version** in sbom_libcoap.yml file as well.

Comment (hmalpani) [reply coap/idf_component.yml]: BTW, what is the timeline for **v4.3.5** release?

Comment (mrdeep1) [reply coap/idf_component.yml]: All depends on what shaking of the v4.3.5rc1 tree finds.  Should be mid July latest I guess.

**Comment by hmalpani** [comment]
Footprint impact:
| Example        | New Firmware Size | Old Firmware Size | Difference |
|--------------------|-----------------------------|----------------------------|----------------|
| coap_client   | 956056                      | 951620                     | 4436          |
| coap_server | 964168                      | 959084                     | 5084           |

**Comment by mahavirj** [comment]
> Footprint impact:
> Example 	New Firmware Size 	Old Firmware Size 	Difference
> coap_client 	956056 	951620 	4436
> coap_server 	964168 	959084 	5084

@mrdeep1 Any thoughts on the code size impact after this change?

File: coap/port/include/coap_config.h [hunk 2]
```diff
@@ -58,6 +58,10 @@
 #define COAP_IPV6_SUPPORT 1
 #endif /* CONFIG_LWIP_IPV6 */
 
+#ifdef CONFIG_ASYNC_SUPPORT
```
hmalpani, line 61:
```suggestion
#ifdef CONFIG_COAP_ASYNC_SUPPORT
```

mrdeep1: Good catch.  Fixed.

mrdeep1:
> Any thoughts on the code size impact after this change?

Nothing in particular stands out.  I would not have thought the extra code for multi-thread support would have added a lot as it is mainly no-op type code if not enabled but does have some wrapper functions.

Is this a significant jump?

It is possible to set the default for support TCP sessions to "n" which will save some code space.

I have updated the PR with your fix and bumped the code version to 4.3.5rc2.

mahavirj:
>  Nothing in particular stands out. I would not have thought the extra code for multi-thread support would have added a lot as it is mainly no-op type code if not enabled but does have some wrapper functions.
Is this a significant jump?

It should be okay, just wanted to check if its due to any of the newly added config options. But I checked, its in general because of the baseline update. I will proceed for the merge here.

mahavirj (approved):

Issue #340, opened 2024-06-13
[led_strip]  After initializing the WS2812 with c6, calling led_strip_clear will cause the LEDs to turn green. (IEC-122)


lijunru-hub:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

led_strip

### ESP-IDF version.

IDF release/v5.2

### Development Kit.

ESP32-C6-Devkit-1 v1.2

### Used Component version.

2.4.3

### More Information.

In the example **led_strip_spi_ws2812**, after powering on and initializing the led_strip, immediately calling **led_strip_clear** will cause the LEDs to turn green.

```c
led_strip_handle_t configure_led(void)
{
    // LED strip general initialization, according to your led board design
    led_strip_config_t strip_config = {
        .strip_gpio_num = LED_STRIP_BLINK_GPIO,   // The GPIO that connected to the LED strip's data line
        .max_leds = LED_STRIP_LED_NUMBERS,        // The number of LEDs in the strip,
        .led_pixel_format = LED_PIXEL_FORMAT_GRB, // Pixel format of your LED strip
        .led_model = LED_MODEL_WS2812,            // LED strip model
        .flags.invert_out = false,                // whether to invert the output signal
    };

    // LED strip backend configuration: SPI
    led_strip_spi_config_t spi_config = {
        .clk_src = SPI_CLK_SRC_DEFAULT, // different clock source can lead to different power consumption
        .flags.with_dma = true,         // Using DMA can improve performance and help drive more LEDs
        .spi_bus = SPI2_HOST,           // SPI bus ID
    };

    // LED Strip object handle
    led_strip_handle_t led_strip;
    ESP_ERROR_CHECK(led_strip_new_spi_device(&strip_config, &spi_config, &led_strip));
    ESP_ERROR_CHECK(led_strip_clear(led_strip));
    ESP_LOGI(TAG, "Created LED strip object with SPI backend");

    return led_strip;
}
```

![1c522d70aef8803085c4a2b011c131a2](https://github.com/espressif/idf-extra-components/assets/68437239/faf8c81f-fece-494e-a09c-1712b486e526)

[...]

PR #342, opened 2024-06-18
fix(led_strip): fix the reset time and resolution bugs (IEC-124)

Kainarx:
This commit add a delay to ensure that the reset time is enough during the initialization. And remove the rigorous judgment of the real spi clk resolution to support different XTAL freq.

https://github.com/espressif/idf-extra-components/issues/340

# Checklist

- [ ] CI passing

### DIFF
--- a/led_strip/idf_component.yml
+++ b/led_strip/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.5.3"
+version: "2.5.4"
 description: Driver for Addressable LED Strip (WS2812, etc)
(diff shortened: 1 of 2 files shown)


suda-morris (changes requested):
The fix LGTM. Please bump up the led_strip component version to 2.5.4 in https://github.com/espressif/idf-extra-components/blob/master/led_strip/idf_component.yml

suda-morris (approved):

PR #346, opened 2024-06-27
fix(coap): correct sbom version as per component release

mahavirj:
# Checklist

- [x] CI passing

# Change description

Align SBOM version for coap component


hmalpani:
Changes looks good to me.
Should we also update the version in idf_component.yml file of the component as these changes won't be pushed to the component registry until the component version is increased in the idf_component.yml file.

mahavirj:
> Changes looks good to me. Should we also update the version in idf_component.yml file of the component as these changes won't be pushed to the component registry until the component version is increased in the idf_component.yml file.

SBOM daily scan happens on the basis of the repository itself. For the official project based scan, I am expecting official coap release soon, so at that time we can bump up the component version.

Issue #345, opened 2024-06-26
HSV计算精度不对


WzxKB:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

bdc_motor

### ESP-IDF version.

V5.2.1

### Development Kit.

esp32s3

### Used Component version.

V2.5.4

### More Information.

在使用led_strip_set_pixel_hsv（）函数的时候，在esp32中，我发现转换的rgb值不对，查阅代码发现是精度问题。
```
uint32_t rgb_max = value;
uint32_t rgb_min = rgb_max * (255 - saturation) / 255.0f;
```
需要改成、
```
uint32_t rgb_max = value * 255 / 100;  // 调整以确保 value 在正确的范围内
uint32_t rgb_min = rgb_max * (100 - saturation) / 100;  // 适用 0-100 范围的 saturation

```
不然计算结果在我的单片机上是错误的，灯颜色不对

Kainarx:
@WzxKB You can rescale **S** and **V** from 0-1 to 0-255 before calling this function, so that the calculated values will be right.

PR #347, opened 2024-07-02
docs(led_strip): change api description of hsv (IEC-131)

Kainarx:
# Checklist

Closes https://github.com/espressif/idf-extra-components/issues/345

- [x] CI passing

# Change description
change api description of hsv in api.md

### DIFF
--- a/led_strip/api.md
+++ b/led_strip/api.md
@@ -10,84 +10,102 @@
 
 ## File include/led_strip.h
 
+
+
+
+
+
 ## Functions
 
 | Type | Name |
 | ---: | :--- |
-|  esp\_err\_t | [**led\_strip\_clear**](#function-led_strip_clear) ([**led\_strip\_handle\_t**](#struct-led_strip_t) strip) <br>_Clear LED strip (turn off all LEDs)_ |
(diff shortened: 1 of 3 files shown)


led_strip/api.md
```diff
@@ -131,8 +131,8 @@ esp_err_t led_strip_set_pixel_hsv (
 - `strip` LED strip
 - `index` index of pixel to set
 - `hue` hue part of color (0 - 360)
-- `saturation` saturation part of color (0 - 255)
-- `value` value part of color (0 - 255)
+- `saturation` saturation part of color (0 - 255, rescaled from 0 - 1)
+- `value` value part of color (0 - 255, rescaled from 0 - 1)
```
suda-morris, line 135: this file actually was auto-generated from the header file.. Use the way suggested in https://github.com/espressif/idf-extra-components/issues/248

suda-morris: so let's update the header file first, and then sync the changes to api.md

suda-morris: can we give the user an example in the API doc, saying if the situation is 50%, how should he set the value here

PR #350, opened 2024-07-03
fix(esp_encrypted_img): correct the dependency for generating encrypted image

mahavirj:
Closes https://github.com/espressif/idf-extra-components/issues/349

# Checklist

- [X] CI passing

# Change description

### DIFF
--- a/esp_encrypted_img/CHANGELOG.md
+++ b/esp_encrypted_img/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.2.1
+
+- Build system: fix the dependency for generating pre encrypted image
+
 ## 2.2.0
 
 ### Enhancements:
--- a/esp_encrypted_img/idf_component.yml
+++ b/esp_encrypted_img/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.2.0"
+version: "2.2.1"
 description: ESP Encrypted Image Abstraction Layer
 url: https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img
 dependencies:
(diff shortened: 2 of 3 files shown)


bonnee:
While you're there, it would be more versatile to have the function create a dynamic target name (e.g. based on input/output file name).
This will enable the possibility to call the function on multiple files (in my use case I need to encrypt multiple partitions)

mahavirj:
@bonnee 

I didn't quite get your requirement, can you please share more information, any pointers to your use-case? Interim, the existing target is used from ESP-IDF example as well, so we can keep that same and maybe consider making provision for new one.

bonnee:
Okay, so I have a project that makes use of a spiffs partition. I want to encrypt both the firmware partition and the spiffs in order to securely flash them on my esp32. With the current function I cannot do that as there would be a duplicate target name. I recreated the function by changing `add_custom_target(encrypt_bin_target, ...` with
```
get_filename_component(name ${tar_file} NAME)
set(target_name archive_${name}_target)
add_custom_target(${target_name}, ...
```
This will create (almost) unique targets for each call and solve my use case.

mahavirj:
@bonnee 

>  I want to encrypt both the firmware partition and the spiffs in order to securely flash them on my esp32.

Just to clarify the pre encrypted OTA scheme only ensures confidentiality over network transport and as such its not related to flash encryption. Please refer this guide: https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-reference/system/esp_https_ota.html#ota-upgrades-with-pre-encrypted-firmware.

bonnee:
Yes I am aware of that. I want to encrypt images in order to send them securely over unsecure channels

hmalpani:
@bonnee 
The script `esp_enc_img_gen.py` present in [tools/](https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img/tools) can be used to create the encrypted image for the spiffs partition.

hmalpani (approved):
I tested the changes and was able to create the encrypted binary for both the application and the spiffs binary. LGTM

PR #354, opened 2024-07-09
ci(gh-actions-ver): updated gh actions latest versions (new nodejs) (IEC-136)

tomassebestik:
This should fix the issue with incompatible version of nodejs (npm) in GitHub actions.

## Related
- https://github.com/espressif/idf-extra-components/actions/runs/9791007440/job/27045936740?pr=348#step:3:16


tomassebestik:
Update external actions version doesn't solve the problem, closing

PR #355, opened 2024-07-10
Update fmt to 11.0.1, add example

igrr:
- Added a simple example to `fmt` component and a test to check that the example works
- Updated `fmt` to 11.0.1

### DIFF
--- a/fmt/LICENSE
+++ b/fmt/LICENSE
@@ -1,0 +1,1 @@
+fmt/LICENSE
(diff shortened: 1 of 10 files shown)


fmt/LICENSE
mahavirj, line 1: Should we add soft link instead to `fmt/fmt/LICENSE` file? (Just to be consistent with other components)

mahavirj (approved):

fmt/LICENSE
igrr: Sorry, I was making this commit on Windows, so couldn't do it using a symlink. I'll fix it up.

igrr: Fixed

PR #356, opened 2024-07-28
Use 32 bit sector id

UnTraDe:
# Checklist

- [x] CI passing

# Change description
The `dhara` library uses the `dhara_sector_t` (`typedef uint32_t dhara_sector_t`) to address sectors internally. This PR changes the component's interface to accept `uint32_t` sector id where applicable.
This change will allow to utilize the entire capacity of NAND chips with sector size > 2^16.


spi_nand_flash/test_app/main/test_spi_nand_flash.c
```diff
@@ -153,14 +155,15 @@ static void do_single_write_test(spi_nand_flash_device_t *flash, uint32_t start_
...
 {
-    uint16_t sector_num, sector_size;
+    uint32_t sector_num;
+    uint16_t sector_size;
     spi_nand_flash_device_t *nand_flash_device_handle;
     spi_device_handle_t spi;
     setup_nand_flash(&nand_flash_device_handle, &spi);
 
     TEST_ESP_OK(spi_nand_flash_get_capacity(nand_flash_device_handle, &sector_num));
     TEST_ESP_OK(spi_nand_flash_get_sector_size(nand_flash_device_handle, &sector_size));
```
igrr, line 165: @UnTraDe Would it make sense to change the type also for `spi_nand_flash_get_sector_size`? I don't know if we will run into a part which has a 64kB sector size in the future. And it seems that there is no extra cost to using a 32-bit variable, since an arguments occupies the whole 32-bit register anyway.

igrr (commented):
Looks good to me, just one suggestion.

spi_nand_flash/test_app/main/test_spi_nand_flash.c
UnTraDe: @igrr Yeah I also thought so at first, but then I saw this part: 
https://github.com/espressif/idf-extra-components/blob/2da4e71f7c5881df286a717e7a4c04fa25f08357/spi_nand_flash/diskio/diskio_nand.c#L95C1-L102C1

I tracked down `WORD` to be defined as 16bit. So it seems to be limited there, and allowing for `uint32_t` in a higher level API might introduce new bugs "silently" by having it truncated later. I also though about going the other way around and change the `page_size`in `spi_nand_flash_device_t` to be `uint16_t` but I haven't fully traced all the places that rely on that.

igrr: Good point! What about expanding the sector size to a 32-bit variable and modifying the code in https://github.com/espressif/idf-extra-components/blob/2da4e71f7c5881df286a717e7a4c04fa25f08357/spi_nand_flash/diskio/diskio_nand.c#L99 to check for overflow before assigning? (And returning an error on overflow.) At least the underlying spi_nand_flash driver will be ready to support larger sector sizes, even if the upper layer FAT integration won't work.

UnTraDe: Yeah sounds good!

igrr (approved):
Changes look good to me!

It would be great if you could clean up the commit history in this branch. (If not, no issue, I can do it for you before merging.)

spi_nand_flash/diskio/diskio_nand.c
```diff
@@ -85,17 +85,23 @@ DRESULT ff_nand_ioctl(BYTE pdrv, BYTE cmd, void *buff)
...
     }
     case GET_SECTOR_SIZE: {
-        uint16_t capacity;
+        uint32_t capacity;
         ESP_GOTO_ON_ERROR(spi_nand_flash_get_sector_size(dev, &capacity),
                           fail, TAG, "get_sector_size failed");
+
+        if (capacity > 0xFFFF) {
+            ESP_LOGE(TAG, "sector size too large");
+            return RES_PARERR;
```
UnTraDe, line 102: @igrr 
Do you think this is the appropriate error to return when overflowing?

https://github.com/espressif/esp-idf/blob/23e4823f17a8349b5e03536ff7653e3e584c9351/components/fatfs/src/diskio.h#L18-L24

UnTraDe:
> Changes look good to me!
> 
> It would be great if you could clean up the commit history in this branch. (If not, no issue, I can do it for you before merging.)

Do you mean to rebase? Or squash?

igrr:
> Do you mean to rebase? Or squash?

If doing an interactive rebase (`git rebase -i` or an equivalent in a GUI tool) I would use "fixup" instead of "squash" since the intermediate commit messages are not so useful. Please also make the final commit message in Conventional Commits format, e.g. `change(spi_nand_flash): use 32-bit sector size and id`.

spi_nand_flash/diskio/diskio_nand.c
igrr: It seems fine, it should prevent FATFS from initializing further.

RathiSonika (approved):
Changes look good to me.  @UnTraDe  thank you for the PR.

UnTraDe:
> If doing an interactive rebase (`git rebase -i` or an equivalent in a GUI tool) I would use "fixup" instead of "squash" since the intermediate commit messages are not so useful. Please also make the final commit message in Conventional Commits format, e.g. `change(spi_nand_flash): use 32-bit sector size and id`.

@igrr Anything else before merging?

PR #359, opened 2024-08-01
change: exclude CVEs that have already been resolved

fhrbata:
These CVEs have not yet been analyzed in the NVD and were identified by the esp-idf-sbom checker during a keyword search in the CVE descriptions. Since the affected components have already been updated to newer versions, these CVEs are no longer relevant to the current version. Therefore, they should be added to the exclude list.

# Checklist

- [ ] CI passing

# Change description

### DIFF
--- a/nghttp/sbom_nghttp2.yml
+++ b/nghttp/sbom_nghttp2.yml
@@ -5,3 +5,6 @@ supplier: 'Organization: nghttp2 <https://nghttp2.org/'
 description: nghttp2 - HTTP/2 C Library and tools
 url: https://github.com/nghttp2/nghttp2
 hash: d13a5758373931064636c1641db6277db45552dc
+cve-exclude-list:
+  - cve: CVE-2024-28182
+    reason: Resolved in version v1.61.0
--- a/coap/sbom_libcoap.yml
+++ b/coap/sbom_libcoap.yml
@@ -5,3 +5,8 @@ supplier: 'Organization: libcoap <https://libcoap.net/>'
 description: A CoAP (RFC 7252) implementation in C
 url: https://github.com/obgm/libcoap
 hash: 3c720f232f61cf097d7cbd5c0bde2b4c681b1c68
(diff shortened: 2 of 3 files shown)


nghttp/sbom_nghttp2.yml
```diff
@@ -5,3 +5,6 @@ supplier: 'Organization: nghttp2 <https://nghttp2.org/'
 description: nghttp2 - HTTP/2 C Library and tools
 url: https://github.com/nghttp2/nghttp2
 hash: d13a5758373931064636c1641db6277db45552dc
+cve-exclude-list:
+  - cve: CVE-2024-28182
+    reason: Resolved in version v1.61.0
```
mahavirj, line 10: I see 2 more advisories from the scan: [CVE-2024-27983](https://nvd.nist.gov/vuln/detail/CVE-2024-27983) and [CVE-2024-25581](https://nvd.nist.gov/vuln/detail/CVE-2024-25581), shall we add those here?

fhrbata: Hi @mahavirj, I have already added them to the global exclude list in
`esp-idf-sbom` since they are not related to nghttp2. I have requested your
review on the esp-idf-sbom merge request.  We can add them here as well, but 
the idea is to place false positives (unrelated to the package) into the global
exclude list in `esp-idf-sbom`. This list should idealy only contain CVEs that
are genuinely related to nghttp2. For idf-extra-components, it likely doesn't
matter whether we exclude them here or in the global exclude list. However, for 
esp-idf, we would need to add the false positive CVEs to each branch
individually instead of just one place in the global exclude list. I'm open to
any suggestions on how to improve this and make it more user-friendly. Thank
you!

mahavirj: Aha, I missed that. Thanks, the current approach looks good.

mahavirj (approved):

Issue #361, opened 2024-08-05
led_strip: avoid dependency on `driver` in recent IDF versions (IEC-147)


igrr:
### Is your feature request related to a problem?

https://github.com/espressif/idf-extra-components/blob/dd2ce03b472504f384fa228ca2938230f2b5bfc2/led_strip/CMakeLists.txt#L22 adds a dependency on `driver` component, which in recent IDF versions pulls in all individual driver components for backward compatibility reasons

### Describe the solution you'd like.

Check IDF version and either add dependency on `driver` (for older versions) or on specific driver component(s) (for newer versions)

PR #344, opened 2024-06-20
feat(led_strip): Support more pixel orders of 3 colors (IEC-126)

Kainarx:
# Checklist

- [ ] Component contains License
- [ ] Component contains README.md
- [ ] Component contains idf_component.yml file with `url` field defined
- [ ] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [ ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description

Closes https://github.com/espressif/idf-extra-components/issues/341

PR #363, opened 2024-08-12
Change/led strip dependency

suda-morris:
Closes https://github.com/espressif/idf-extra-components/issues/361

### DIFF
--- a/led_strip/CMakeLists.txt
+++ b/led_strip/CMakeLists.txt
@@ -1,22 +1,31 @@
 include($ENV{IDF_PATH}/tools/cmake/version.cmake)
 
 set(srcs "src/led_strip_api.c")
+set(public_requires)
 
+# Starting from esp-idf v5.x, the RMT driver is rewritten
 if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.0")
     if(CONFIG_SOC_RMT_SUPPORTED)
         list(APPEND srcs "src/led_strip_rmt_dev.c" "src/led_strip_rmt_encoder.c")
     endif()
-else() 
+else()
     list(APPEND srcs "src/led_strip_rmt_dev_idf4.c")
 endif()
 
-# the SPI backend driver relies on something that was added in IDF 5.1
+# the SPI backend driver relies on some feature that was available in IDF 5.1
 if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.1")
     if(CONFIG_SOC_GPSPI_SUPPORTED)
         list(APPEND srcs "src/led_strip_spi_dev.c")
     endif()
 endif()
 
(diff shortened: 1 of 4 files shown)


led_strip/CMakeLists.txt
```diff
@@ -1,22 +1,32 @@
...
+set(public_requires)
 
+# Starting from esp-idf v5.x, the RMT driver is rewritten
 if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER_EQUAL "5.0")
     if(CONFIG_SOC_RMT_SUPPORTED)
         list(APPEND srcs "src/led_strip_rmt_dev.c" "src/led_strip_rmt_encoder.c")
     endif()
-else() 
+else()
+    message(WARNING "esp-idf v4.x has reached end-of-life. The compatibility with v4.x will be removed in the next major release.")
```
igrr, line 12: I think the warning is probably unnecessary: if the future release removes v4.4 compatibility, the component manager should still be able to install the last compatible release.

led_strip/README.md
```diff
@@ -27,13 +27,9 @@ led_strip_config_t strip_config = {
 };
 
 led_strip_rmt_config_t rmt_config = {
-#if ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 0, 0)
```
igrr, line 30: Does this change mean we are already removing v4.x support? In that case we should update idf_component.yml.

igrr:
@suda-morris thanks for the PR! LGTM aside from two questions.

led_strip/README.md
suda-morris: Not yet, I plan to remove the 4.x support in led_strip version 3 (still in feature collecting stage). I only removed it from the programming guide for now, as "not recommended".

igrr: Okay, then I don't understand this part of the diff. Previously, for IDF versions <5.0, the initializer was
`led_strip_rmt_config_t rmt_config = {.rmt_channel = 0}`. Now, it will be `led_strip_rmt_config_t rmt_config = { .clk_src = RMT_CLK_SRC_DEFAULT, ... }`. Will this work in v4.4?

igrr (approved):

igrr:
@suda-morris could you please also bump the version in idf_component.yml, so that the changes get released?

suda-morris:
> @suda-morris could you please also bump the version in idf_component.yml, so that the changes get released? And update the changelog file.

Hi @igrr Do we want to bump the version every time we made a change? This change doesn't affect the functionality of led_strip in fact, so I want to bump the version in the next time when we have a new feature or a bugfix. e.g. in https://github.com/espressif/idf-extra-components/pull/344

igrr:
I think it does improve developer experience because the time for the full build with IDF v5.3 can now be shorter. So from my perspective, it's worth publishing the release. Besides, there is almost no cost associated with publishing the new version.

If you plan to merge #344 soon then I don't mind bumping the version after that one, either.

Issue #368, opened 2024-08-29
Coremark: a wrong library archive used by linker.lf? (IEC-158)


suda-morris:
### Answers checklist.

- [X] I have read the documentation of the component in question and the issue is not addressed there.
- [X] I have searched the issue tracker for a similar issue and not found a similar issue.

### General issue report

As a managed component, coremark component will be compiled into a static library: `libespressif__coremark.a`.

While in the linker.lf file, we're still using `libcoremark.a` as the archive name.

objdump of the example ELF file, 

```
42006fc4 <core_bench_matrix>:
{
42006fc4:	004136        	entry	a1, 32
    crc = crc16(matrix_test(N, C, A, B, val), crc);
42006fc7:	0222d2        	l32i	a13, a2, 8
42006fca:	0122c2        	l32i	a12, a2, 4
42006fcd:	32b8      	l32i.n	a11, a2, 12
42006fcf:	0022a2        	l32i	a10, a2, 0
42006fd2:	23e380        	sext	a14, a3, 15
42006fd5:	ffc8e5        	call8	42006c64 <matrix_test>
```

The function is still linked to the Flash cache address space.

After change to `archive: libespressif__coremark.a`, we can finally see its in the IRAM now.

```c
4037a244 <core_bench_matrix>:
{
4037a244:	004136        	entry	a1, 32
    crc = crc16(matrix_test(N, C, A, B, val), crc);
4037a247:	0222d2        	l32i	a13, a2, 8
4037a24a:	0122c2        	l32i	a12, a2, 4
4037a24d:	32b8      	l32i.n	a11, a2, 12
4037a24f:	0022a2        	l32i	a10, a2, 0
4037a252:	23e380        	sext	a14, a3, 15
4037a255:	ffc8e5        	call8	40379ee4 <matrix_test>
```

I wish I could see the link errors when I `export LDGEN_CHECK_MAPPING=1`, but no, the build still success. Seems to me like it's a bug in the ldgen.

If we change the linker.lf into this

```
[mapping:coremark]
archive: libcoremark.a
entries:
    core_matrix (noflash)
```
We will see the errors as expected:

```
linker script generation failed for /home/morris/esp/coremark_example/build/esp-idf/esp_system/ld/sections.ld.in
ERROR: 'libcoremark.a:core_matrix None' not found
In fragment 'coremark' defined in '/home/morris/esp/coremark_example/managed_components/espressif__coremark/linker.lf'.
```

PR #383, opened 2024-09-24
fix(coremark): archive in linker.lf should depend on component name

igrr:
Closes https://github.com/espressif/idf-extra-components/issues/368

### DIFF
--- a/coremark/idf_component.yml
+++ b/coremark/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.1.0"
+version: "1.1.0~1"
 description: CoreMark Benchmark
 url: https://github.com/espressif/idf-extra-components/tree/master/coremark
 issues: https://github.com/espressif/idf-extra-components/issues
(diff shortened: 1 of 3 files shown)


coremark/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.1.0"
+version: "1.1.1"
```
igrr, line 1: Or perhaps it should be
```suggestion
version: "1.1.0~1"
```
, not sure. The upstream tag is [v1.01](https://github.com/eembc/coremark/releases/tag/v1.01)...

suda-morris: I vote for `1.1.0~1`

suda-morris (approved):

coremark/idf_component.yml
igrr: I guess too late to roll back to 1.0.1 now, I'll set it to 1.1.0~1.

Issue #386, opened 2024-09-27
CI: migrate clang_tidy job to its own test app, remove "test_app" directory (IEC-170)


igrr:
After https://github.com/espressif/idf-extra-components/pull/381, all the component tests have been migrated from the common `test_app` to component-specific test apps or examples. The only place test_app is used is the clang-tidy job.

Task is to migrate run clang-tidy on a simpler app and remove the original "test_app" entirely.

PR #390, opened 2024-09-27
ci: use a simple test app for clang-tidy, remove the original test_app, readme updates

igrr:
- Created a simple(r) test_app for `clang-tidy` use only
- Removed the original `test_app`
- Updated README: fixed build/test workflow badge URL, rewrote some sections
- Renamed .github/README.md added in #392 to .github/readme_workflows.md. It turns out that `.github/README.md` [has higher priority than the root README.md](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes#about-readmes), so we now have the wrong readme shown on the main page of the repo.

Closes https://github.com/espressif/idf-extra-components/issues/386


.github/clang-tidy/test_app/main/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+idf_component_register(SRCS "test_app.c"
+                    INCLUDE_DIRS ".")
```
tore-espressif, line 2: Nitpick: for `idf.py reconfigure` you don't even need this dummy main component. Sure, it will fail in `idf.py build`, but you will probably never do that...

I'm only comparing with https://github.com/espressif/esp-usb/tree/master/clang_tidy ...

tore-espressif (approved):

.github/clang-tidy/test_app/main/CMakeLists.txt
igrr: Have removed, thanks for noticing!

PR #381, opened 2024-09-18
ci: add component test apps

igrr:
Done in this PR:

- [x] Moved all existing unit tests into per-component test apps — https://github.com/espressif/idf-extra-components/issues/149.
- [x] Added per-component .build-test-rules.yml files
- [x] For components which don't have any unit tests and don't have examples, added a build-only test app
- [x] Added some new tests where it was easy
   - added pytest for coremark example
   - libsodium test now actually compares test output to .exp files
   - added test for libpng
- [x] Build examples and test_apps based on changes in the PR (only relevant components) — https://github.com/espressif/idf-extra-components/issues/16
- [x] Junit reports from examples and test_apps are combined and the results are published
- [x] Removed all target markers (`pytest.mark.supported_targets`, `pytest.mark.esp32`) because they didn't do anything
- [x] Added a pre-commit hook with a few consistency checks
   - every component is added to the upload job
   - every component is mentioned in the issue template — https://github.com/espressif/idf-extra-components/issues/157
   - every .build-test-rules.yml file is mentioned in the idf-build-apps config
      - we can add a "glob" feature to idf-build-apps, then this can be simplified
   - we can add more checks (e.g. changelog, examples, idf_component.yml fields, SBOM files, ...)

Next PR:

- [ ] Migrate clang-tidy job from `test_app`
- [ ] Remove `test_app`
- [ ] Run target test jobs on more chips
- [ ] Re-add test parametrization by chip targets using `pytest.mark.parametrize` over `target` parameter
- [ ] Revisit build-test-rules files, remove unnecessary builds, move some target tests to either linux target or qemu

---

🚧 — means that this is something to be done in future PRs. I will create some issues about this when/if this PR gets merged.

Component | Has example(s) | Example has pytest | Test app added | Test app has tests
-- | -- | -- | -- | --
bdc_motor |   |   | yes | no (build only)
catch2 | yes | yes | no, rely on example |  
cbor | yes | yes | no, rely on example |  
ccomp_timer | no 🚧 |   | yes | yes
coap | yes | yes | no, rely on example |  
coremark | yes | yes, added | no, rely on example |  
dhara | no 🚧 |   | no, rely on spi_nand_flash example | no
eigen | yes | no 🚧 | no, rely on example |  
esp_delta_ota | yes | no 🚧 | no, rely on example |  
esp_encrypted_img | no 🚧 |   | yes | yes
esp_jpeg | yes | no 🚧 | no, rely on example |  
[...]

PR #391, opened 2024-09-30
fix(spi_nand_flash): add dhara to public dependencies (+minor changes)

igrr:
* add dhara to public requirements

   Since 892d994e8b3ee1bdf32ca49513a709c44ee9c12f, spi_nand_flash.h includes dhara/map.h, so dhara should be a public dependency of spi_nand_flash.

   Other changes:
   - move vfs to private requirements (it didn't have to be public)
   - depend specifically on `esp_driver_spi` instead of `driver` when building for IDF >=5.3
* fix memory leaks in test (free temporary buffers) and raise leak threshold to not detect esp_intr_alloc lazy allocation as a leak
* add read/write performance output to the test case
* remove unused sbom_dhara.yml file from spi_nand_flash component
* version changed from 0.3.1 to 0.4.0 since removing `vfs` from the public requirements is technically a breaking change

### DIFF
--- a/spi_nand_flash/CMakeLists.txt
+++ b/spi_nand_flash/CMakeLists.txt
@@ -4,7 +4,17 @@ set(srcs "src/nand.c"
          "vfs/vfs_fat_spinandflash.c"
          "diskio/diskio_nand.c")
 
+set(reqs fatfs)
+
+if("${IDF_VERSION_MAJOR}.${IDF_VERSION_MINOR}" VERSION_GREATER "5.3")
+    list(APPEND reqs esp_driver_spi)
+else()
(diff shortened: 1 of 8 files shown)


igrr:
Failing tests are due to the CI workflow refactoring (#381). When the modified component doesn't have any pytest tests, pytest exits with exit code 5 (no tests found). I'll open another PR to fix this.

RathiSonika (approved):
All the changes look good to me. Just one minor adjustment is needed for the test case. Thank you for the updates!

igrr:
@RathiSonika Thanks for the review and for spotting the issue!

PR #396, opened 2024-10-03
feat(https_delta_ota): Added new function to verify the patch file

hrushikesh430:
Added the new function in esp_delta_ota_patch_gen.py which verifies the newly generated patch file.

# Checklist

# Description
Added a new function in esp_delta_ota_patch_gen.py which verifies the newly generated patch file. The function uses the patch.bin and applies over base_binary.bin and calculates the sha of newly generate new.bin file. Then it compares this sha with the sha of new_binary.bin file. If they are same then it is verifies successfully otherwise invalid.

### DIFF
--- a/esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
+++ b/esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# ESP Delta OTA Ptach Generator Tool. This tool helps in generating the compressed patch file
+# ESP Delta OTA Patch Generator Tool. This tool helps in generating the compressed patch file
 # using BSDiff and Heatshrink algorithms
 #
 # SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
@@ -10,6 +10,9 @@ import argparse
 import os
 import subprocess
 import re
+import tempfile
+import hashlib
+import sys
 
 try:
     import detools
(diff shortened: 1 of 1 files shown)


hrushikesh430:
cc @AdityaHPatwardhan

hrushikesh430:
I have also changes the regex of create_patch function, inside re.search. Could you please verify it, if it is correct. I have changed so that it was giving syntax error for previous expression.

AdityaHPatwardhan (commented):
Some minor comments, otherwise LGTM

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
```diff
@@ -10,6 +10,8 @@
 import os
 import subprocess
 import re
+import tempfile
```
AdityaHPatwardhan, line 13: I think these requirements should also be added to https://github.com/espressif/idf-extra-components/blob/master/esp_delta_ota/examples/https_delta_ota/tools/requirements.txt

hrushikesh430: tempfile and hashlib are pre-installed in python, so no need to add these modules in requirements.txt file.

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
```diff
@@ -19,40 +23,115 @@
 # Magic Byte is created using command: echo -n "esp_delta_ota" | sha256sum
 esp_delta_ota_magic = 0xfccdde10
 
-MAGIC_SIZE = 4
-DIGEST_SIZE = 32
-RESERVED_HEADER = 64 - (MAGIC_SIZE + DIGEST_SIZE)
+MAGIC_SIZE = 4 # This is size of magic byte
+DIGEST_SIZE = 32 # This is SHA256 of the base binary    
+CRC_SIZE = 4 # This is CRC32 of patch file which shall be applied on the base binary
+RESERVED_HEADER = 64 - (MAGIC_SIZE + DIGEST_SIZE + CRC_SIZE) # This is reserved header size
```
AdityaHPatwardhan, line 29: I think we can also add a HEADER_VERSION field here,
This field shall be the version of the patch header.
In the esp_delta_ota component, we shall check this version as follows

if (version > 0) {
   verify crc32 integrity of the patch
}

AdityaHPatwardhan: Can we define a new field called `HEADER_SIZE = 64` and use it instead of the value `64`

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
```diff
@@ -19,40 +23,115 @@
...
+# the hash of new_binary and binary.new, if they are same then verfications get successfull otherwise fails.
+def verify_patch(base_binary: str, patch_to_verify: str, new_binary: str) -> None:
+
+    with open(patch_to_verify, "rb") as original_file:
+        # Read CRC32 from header
+        original_file.seek(MAGIC_SIZE + DIGEST_SIZE)
+        header_crc32 = int.from_bytes(original_file.read(CRC_SIZE), 'little')
+        
+        # Calculate CRC32 of patch content
+        original_file.seek(64)  # Move the file pointer 64 bytes ahead
```
AdityaHPatwardhan, line 76: We can use the HEADER_SIZe field here instead of the value 64

AdityaHPatwardhan (commented):
One suggestion, but otherwise the changes LGTM

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
hrushikesh430: > I think we can also add a HEADER_VERSION field here, This field shall be the version of the patch header. In the esp_delta_ota component, we shall check this version as follows
> 
> if (version > 0) { verify crc32 integrity of the patch }

Yep, we can add but where this field should be added, at the beginning of the header or at last. Because I don't think is there any need of specific locations. What's your opinion?

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
```diff
@@ -19,40 +23,120 @@
...
+                print("Patch file verified successfully")
+            else:
+                print("Failed to verify the patch")
+        else:
+            print("Failed to verify the patch")
+        os.remove("binary.new")
+
+def main() -> None:
+    if len(sys.argv) < 2:
+        print("Usage: python file_name.py create_patch/verify_patch [arguments]")
```
mahavirj, line 118: Please put a name of the utility script itself here `esp_delta_ota_patch_gen.py`

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
```diff
@@ -19,40 +23,120 @@
...
-DIGEST_SIZE = 32
-RESERVED_HEADER = 64 - (MAGIC_SIZE + DIGEST_SIZE)
+MAGIC_SIZE = 4 # This is size of magic byte
+DIGEST_SIZE = 32 # This is SHA256 of the base binary    
+HEADER_VERSION_SIZE = 1 # This is size of header version
+CRC_SIZE = 4 # This is CRC32 of patch file which shall be applied on the base binary
+HEADER_SIZE = 64
+RESERVED_HEADER = HEADER_SIZE - (MAGIC_SIZE + DIGEST_SIZE + HEADER_VERSION_SIZE + CRC_SIZE) # This is reserved header size
+
+HEADER_VERSION = 0 # Define the HEADER_VERSION as 0
```
AdityaHPatwardhan, line 33: Instead of naming this as header version, lets rename this as PATCH_VERSION.

lets define PATCH_VERSION = 1 in this case.

That mean all of the previous patches are considered as 0 and now a patch which has this version field shall be considered as version 1

AdityaHPatwardhan (approved):

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
```diff
@@ -19,40 +23,120 @@
...
     os.system("mv " + patch_file_name + " " + patch_file_without_header)
 
+    with open(patch_file_without_header, "rb") as temp_patch:
+        patch_content = temp_patch.read()
+        crc32 = zlib.crc32(patch_content)
+
     with open(patch_file_name, "wb") as patch_file:
         patch_file.write(esp_delta_ota_magic.to_bytes(MAGIC_SIZE, 'little'))
         patch_file.write(bytes.fromhex(x[1].decode()))
+        patch_file.write(HEADER_VERSION.to_bytes(HEADER_VERSION_SIZE, 'little'))
```
mahavirj, line 64: This now changes the image format.

- Do we need to update the image format diagram in the component?
- Should we update the example to handle the CRC verification part?
- How can we make this change backward compatible? Earlier format won't have the CRC field.

AdityaHPatwardhan:
I think you should also update the patch application code in C to make use of the version.
if version > = 1
    perform crc check
 else
    skip crc check ( older patches with version 0)

AdityaHPatwardhan:
@hrushikesh430 The PR LGTM,
We can merge it

AdityaHPatwardhan (approved):

esp_delta_ota/examples/https_delta_ota/tools/esp_delta_ota_patch_gen.py
hrushikesh430: we have removed the CRC field from the header as it is not possible to calculate the base binary CRC on device, because detools applies the patch in chunk i.e. unable to get the patch size at the beginning.

PR #400, opened 2024-10-10
ci(spi_nand_flash): run example and test on hardware (IEC-180)

igrr:
This PR adds pytest scripts for spi_nand_flash component and extends the test matrix to run these tests on hardware.

CI scripts are updated to allow the test app inside the component to be in `test_app/` subdirectory, previously only `test_apps/` were considered.


igrr:
Looks like the runner is working now, tests passed e.g. in https://github.com/espressif/idf-extra-components/actions/runs/11291400912/job/31405895686.

RathiSonika (approved):
LGTM. Thank you!

PR #426, opened 2024-10-16
update(spi_nand_flash): handle alignment and DMA requirements for work buffers

RathiSonika:
Currently spi_nand_flash component passes pointers to TX and RX buffers to spi_master driver. However, these pointers sometimes are either not in DMA-capable memory or not aligned properly, and as a result spi_master driver allocates a temporary buffer for every transaction and copies the data to/from it. This also applies to cases when the transaction is just a couple of bytes long.

In this PR:
1. Handled alignment and DMA requirements for work buffers
2. For short transactions (4 bytes or less), used txdata/rxdata fields of spi transaction structure instead of pointers.


spi_nand_flash/src/nand.c
```diff
@@ -279,18 +285,23 @@ esp_err_t spi_nand_flash_read_sector(spi_nand_flash_device_t *handle, uint8_t *b
     dhara_error_t err;
     esp_err_t ret = ESP_OK;
 
+    uint8_t *buf = heap_caps_malloc(handle->page_size, MALLOC_CAP_DMA | MALLOC_CAP_8BIT);
```
igrr, line 288: Can we allocate the read buffer in `spi_nand_flash_init_device`, so that we don't have to re-allocate it each time?

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -26,7 +26,17 @@ esp_err_t spi_nand_execute_transaction(spi_device_handle_t device, spi_nand_tran
         .dummy_bits = transaction->dummy_bits
     };
 
-    return spi_device_transmit(device, (spi_transaction_t *) &e);
+    if (transaction->flags == SPI_TRANS_USE_TXDATA) {
```
igrr, line 29: Shall we also assert that the data size does not exceed 4 bytes?

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -26,7 +26,17 @@ esp_err_t spi_nand_execute_transaction(spi_device_handle_t device, spi_nand_tran
...
+        *(uint32_t *)e.base.tx_data = *transaction->mosi_data;
```
igrr, line 30: Recommend an explicit `memcpy`, since we don't know whether `transaction->mosi_data` is 4-byte-aligned. Dereferencing an unaligned pointer with `*(uint32_t*)` may result in an unaligned access exception on some chips.

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -83,7 +95,8 @@ esp_err_t spi_nand_read(spi_device_handle_t device, uint8_t *data, uint16_t colu
         .address = column,
         .miso_len = length,
         .miso_data = data,
-        .dummy_bits = 8
+        .dummy_bits = 8,
+        .flags = SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL,
```
igrr, line 99: Perhaps we can just guard this part with an IDF version check, to avoid raising the minimal supported IDF version?

spi_nand_flash/src/spi_nand_oper.c
RathiSonika: That makes sense. I've updated it.

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -26,7 +27,17 @@
...
 
-    return spi_device_transmit(device, (spi_transaction_t *) &e);
+    if (transaction->flags == SPI_TRANS_USE_TXDATA && transaction->mosi_len <= 4) {
+        memcpy(e.base.tx_data, transaction->mosi_data, transaction->mosi_len);
+    }
+
+    esp_err_t ret = spi_device_transmit(device, (spi_transaction_t *) &e);
+    if (ret == ESP_OK) {
+        if (transaction->flags == SPI_TRANS_USE_RXDATA && transaction->miso_len <= 4) {
+            memcpy(transaction->miso_data, e.base.rx_data, transaction->miso_len);
```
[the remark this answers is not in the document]
RathiSonika: @igrr can I dismiss this alert with won't fix reason?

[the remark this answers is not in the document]
igrr: Yes, you can.

I'll see if we can somehow filter our `memcpy` warnings in clang-tidy, since we don't have `memcpy_s` in newlib yet. I don't want to disable `DeprecatedOrUnsafeBufferHandling` completely because it is still useful for other functions like `strcpy`.

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -26,7 +27,17 @@ esp_err_t spi_nand_execute_transaction(spi_device_handle_t device, spi_nand_tran
         .dummy_bits = transaction->dummy_bits
     };
 
-    return spi_device_transmit(device, (spi_transaction_t *) &e);
+    if (transaction->flags == SPI_TRANS_USE_TXDATA && transaction->mosi_len <= 4) {
```
igrr, line 30: I would recommend something like

```c
if (transaction->flags == SPI_TRANS_USE_TXDATA) {
    assert(transaction->mosi_len <= 4 && "SPI_TRANS_USE_TXDATA used for a long transaction")
```

The point is, this is a sanity check for the way we call `spi_nand_execute_transaction` from other source files. If we accidentally pass length > 4 along with SPI_TRANS_USE_TXDATA, we want to notice this immediately. But the way it is right now, the part inside `if` will be skipped and we will just not pass any data to SPI transaction.

spi_nand_flash/src/nand.c
```diff
@@ -162,9 +166,11 @@ static esp_err_t detect_chip(spi_nand_flash_device_t *dev)
         .address = 0, // This normally selects the manufacturer id. Some chips ignores it, but still expects 8 dummy bits here
         .address_bytes = 1,
         .miso_len = 1,
-        .miso_data = &manufacturer_id
+        .miso_data = &manufacturer_id,
+        .flags = SPI_TRANS_USE_RXDATA,
     };
     spi_nand_execute_transaction(dev->config.device_handle, &t);
+    printf("manufacturer_id: %x\n", manufacturer_id);
```
igrr, line 173: If this is not a debugging leftover, please change to ESP_LOGD

igrr (approved):

spi_nand_flash/src/nand.c
```diff
@@ -281,16 +296,17 @@ esp_err_t spi_nand_flash_read_sector(spi_nand_flash_device_t *handle, uint8_t *b
...
     } else if (err) {
         // This indicates a soft ECC error, we rewrite the sector to recover
-        if (dhara_map_write(&handle->dhara_map, sector_id, buffer, &err)) {
+        if (dhara_map_write(&handle->dhara_map, sector_id, handle->read_buffer, &err)) {
             ret = ESP_ERR_FLASH_BASE + err;
         }
     }
 
     xSemaphoreGive(handle->mutex);
+    memcpy(buffer, handle->read_buffer, handle->page_size);
```
pacucha42, line 309: Proposal: maybe we could introduce a new API expecting DMA buffer as out-parameter? Something like spi_nand_flash_read_sector_dma(..., caller_provided_dma_buffer, ...) to avoid extra memcpy(). In some future update, not in given PR

pacucha42 (approved):
LGTM, approved. Thanks @RathiSonika

spi_nand_flash/src/nand.c
```diff
@@ -242,6 +254,9 @@ esp_err_t spi_nand_flash_init_device(spi_nand_flash_config_t *config, spi_nand_f
     if ((*handle)->work_buffer != NULL) {
         free((*handle)->work_buffer);
     }
+    if ((*handle)->read_buffer != NULL) {
```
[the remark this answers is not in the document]
RathiSonika: Thank you for the review. I'll address these comments in the upcoming PR.

PR #436, opened 2024-10-22
feat(spi_nand_flash): Add Kconfig option to verify write operations

RathiSonika:
# Change description

1. Added an Kconfig option to verify write operations which will be useful when debugging hardware issues or issues with the driver itself.
2. Fixed an issue in the spi_nand_erase_chip() API where the mutex was not released, leading to a deadlock if another API attempted to acquire the mutex after calling spi_nand_erase_chip(). Updated test case for the same.
3. Added `spi_nand_flash_copy_sector()` API in nand.c and test case for the same.


spi_nand_flash/src/nand.c
```diff
@@ -279,6 +279,8 @@ esp_err_t spi_nand_erase_chip(spi_nand_flash_device_t *handle)
                           end, TAG, "");
     }
 
+    xSemaphoreGive(handle->mutex);
+
```
igrr, line 283: Alternatively, we can remove `return ESP_OK;` on the following line. Then we will go through the `end` label, releasing the mutex and returning ESP_OK.

spi_nand_flash/src/nand.c
```diff
@@ -320,6 +339,16 @@ esp_err_t spi_nand_flash_write_sector(spi_nand_flash_device_t *handle, const uin
     if (dhara_map_write(&handle->dhara_map, sector_id, buffer, &err)) {
         ret = ESP_ERR_FLASH_BASE + err;
     }
+#if CONFIG_NAND_FLASH_VERIFY_WRITE
+    if (ret == ESP_OK) {
+        ret = s_verify_write(handle, sector_id, buffer, &err);
+        if (ret == ESP_OK) {
+            ESP_LOGI(TAG, "%s: Write buffer verification succeeded", __func__);
```
igrr, line 346: Logging something on success is probably unnecessary, this will just flood the output. Let's log only on failed verification.

spi_nand_flash/src/nand.c
```diff
@@ -310,6 +312,23 @@ esp_err_t spi_nand_flash_read_sector(spi_nand_flash_device_t *handle, uint8_t *b
...
+static esp_err_t s_verify_write(spi_nand_flash_device_t *handle, dhara_sector_t sector_id, const uint8_t *expected_buffer, dhara_error_t *err)
+{
+    esp_err_t ret = ESP_OK;
+    if (dhara_map_read(&handle->dhara_map, sector_id, handle->read_buffer, err)) {
+        ret = ESP_ERR_FLASH_BASE + *err;
+        ESP_LOGE(TAG, "failed to read nand flash to verify previous write, err: 0x%x", *err);
+        return ret;
+    }
+
+    if (memcmp(handle->read_buffer, expected_buffer, handle->page_size)) {
```
igrr, line 325: To make these errors actually useful for troubleshooting, I think it would be better to log the expected and actual word which doesn't match.

https://github.com/espressif/esp-idf/blob/9106c43accd9f5e75379f62f12597677213f5023/components/spi_flash/esp_flash_api.c#L985-L992

spi_nand_flash/src/nand.c
```diff
@@ -310,6 +312,23 @@ esp_err_t spi_nand_flash_read_sector(spi_nand_flash_device_t *handle, uint8_t *b
...
+        ret = ESP_FAIL;
```
igrr, line 326: Description of spi_nand_flash_write_sector doesn't mention ESP_FAIL now

spi_nand_flash/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "0.4.1"
+version: "0.4.2"
```
igrr, line 1: Since your change is `feat(...)`, you need to increase minor version: https://semver.org/

igrr (commented):
Thanks for the PR @RathiSonika.

I left a few minor comments. The major one is whether we should indeed verify writes above Dhara layer, like you did, or below it, in nand.c.

To me it seems that verifying writes at lower layer achieves the goal of making sure our Flash operations work correctly, while verifying at upper layer doesn't.

We do writes to Flash in dhara_nand_prog, dhara_nand_copy, dhara_nand_mark_bad. If we fail to perform the write in dhara_nand_mark_bad correctly, we might not get the verification error in `spi_nand_flash_write_sector`.

spi_nand_flash/src/dhara_glue.c
```diff
@@ -154,6 +178,13 @@ int dhara_nand_prog(const struct dhara_nand *n, dhara_page_t p, const uint8_t *d
         return -1;
     }
 
+#if CONFIG_NAND_FLASH_VERIFY_WRITE
+    ret = s_verify_write(dev, data, 0, dev->page_size);
+    if (ret != ESP_OK) {
+        ESP_LOGE(TAG, "%s: prog page=%"PRIu32" write verification failed", __func__, p);
+    }
```
igrr, line 185: Technically we also have to verify the used marker here

igrr (approved):

spi_nand_flash/src/dhara_glue.c
igrr: I mean that a few lines above, we write two parts:

```c++
    ESP_GOTO_ON_ERROR(spi_nand_program_load(dev->config.device_handle, data, 0, dev->page_size),
                      fail, TAG, "");
    ESP_GOTO_ON_ERROR(spi_nand_program_load(dev->config.device_handle, (uint8_t *)&used_marker,
                                            dev->page_size + 2, 2),
```

This `s_verify_write` call checks that the first part (from 0 to page_size) is written correctly. We should call s_verify_write second time to check that the used page marker (page_size ... page_size + 2) is also written correctly.

RathiSonika: Oh, my mistake—I misunderstood. Yes, it should verify the markers. I’ll update it. Thank you!

PR #437, opened 2024-10-24
feat(spi_nand_flash): Add data refresh threshold for ECC correction

RathiSonika:
# Change description

Currently, spi_nand_flash_read_sector() rewrites the sector without considering the number of bits corrected by ECC. 

In this PR, a Kconfig option has been added to configure the data refresh threshold at runtime, allowing the decision on whether a data refresh is needed to be based on this threshold and number of bits corrected by ECC


spi_nand_flash/src/nand.c
```diff
@@ -289,6 +291,25 @@ esp_err_t spi_nand_erase_chip(spi_nand_flash_device_t *handle)
...
+    if (handle->ecc_data.ecc_corrected_bits_status == STAT_ECC_1_TO_3_BITS_CORRECTED) {
+        min_bits_corrected = 1;
+    } else if (handle->ecc_data.ecc_corrected_bits_status == STAT_ECC_4_TO_6_BITS_CORRECTED) {
+        min_bits_corrected = 4;
+    } else if (handle->ecc_data.ecc_corrected_bits_status == STAT_ECC_7_8_BITS_CORRECTED) {
+        min_bits_corrected = 7;
+    }
+
+    // if number of corrected bits is greater than refresh threshold then rewite the sector
+    if (min_bits_corrected >= CONFIG_NAND_FLASH_ECC_DATA_REFRESH_THRESHOLD) {
```
igrr, line 307: Seems like this is the only place where this option is used, so it's easy to make it a run-time configurable option (in spi_nand_flash_config_t). Kconfig options should be introduced only when changing something at run time is hard or expensive.

spi_nand_flash/src/nand.h
```diff
@@ -31,6 +47,7 @@ struct spi_nand_flash_device_t {
     uint32_t read_page_delay_us;
     uint32_t erase_block_delay_us;
     uint32_t program_page_delay_us;
+    ecc_data_t ecc_data;
```
igrr, line 50: Perhaps a more descriptive name, such as "last_ecc_data", could be used?

spi_nand_flash/src/dhara_glue.c
```diff
@@ -178,9 +178,20 @@ int dhara_nand_is_free(const struct dhara_nand *n, dhara_page_t p)
     return 0;
 }
 
-static int is_ecc_error(uint8_t status)
+#define PACK_2BITS_STATUS(status, bit1, bit0)         ((((status) & (bit1)) << 1) | ((status) & (bit0)))
+#define PACK_3BITS_STATUS(status, bit2, bit1, bit0)   ((((status) & (bit2)) << 2) | (((status) & (bit1)) << 1) | ((status) & (bit0)))
+
+static ecc_status_t is_ecc_error(uint8_t status, uint8_t ecc_status_reg_len_in_bits)
```
igrr, line 184: "is_..." prefix is usually used for functions which return true/false.

Here it sounds more like "get_ecc_status" or "extract_ecc_status".

spi_nand_flash/src/nand.c
```diff
@@ -145,6 +145,7 @@ static esp_err_t spi_nand_micron_init(spi_nand_flash_device_t *dev)
         .flags = SPI_TRANS_USE_RXDATA,
     };
     spi_nand_execute_transaction(dev->config.device_handle, &t);
+    dev->ecc_data.ecc_status_reg_len_in_bits = 3;
```
igrr, line 148: Do all other vendors use the default (2 bits)?..

spi_nand_flash/src/nand.h
RathiSonika: Since it contains both configuration settings and status, I named it ecc_data. However, can it be ecc_context, ecc_info, ecc_control, ecc_conf_stat?

spi_nand_flash/src/dhara_glue.c
```diff
@@ -235,10 +246,17 @@ int dhara_nand_read(const struct dhara_nand *n, dhara_page_t p, size_t offset, s
...
+    dev->ecc_data.ecc_corrected_bits_status = bits_corrected_status;
+    if (bits_corrected_status) {
         dhara_set_error(err, DHARA_E_ECC);
-        return -1;
+        if (bits_corrected_status == STAT_ECC_MAX) {
+            ESP_LOGE(TAG, "%s: Error while initializing value of ecc_status_reg_len_in_bits", __func__);
+            return -1;
+        } else if (bits_corrected_status == STAT_ECC_NOT_CORRECTED) {
+            ESP_LOGD(TAG, "read ecc error, page=%"PRIu32"", p);
+            return -1;
```
igrr, line 258: According to

https://github.com/dlbeer/dhara/blob/1b166e41b74b4a62ee6001ba5fab7a8805e80ea2/dhara/nand.h#L89-L90

we should set DHARA_E_ECC error if an uncorrectable error occurs. If i'm reading the code right, now we also set it also if e.g. one bit error occurred and was corrected.

spi_nand_flash/src/dhara_glue.c
```diff
@@ -261,10 +279,17 @@ int dhara_nand_copy(const struct dhara_nand *n, dhara_page_t src, dhara_page_t d
...
+    if (bits_corrected_status) {
         dhara_set_error(err, DHARA_E_ECC);
-        return -1;
+        if (bits_corrected_status == STAT_ECC_MAX) {
+            ESP_LOGE(TAG, "%s: Error while initializing value of ecc_status_reg_len_in_bits", __func__);
+            return -1;
+        } else if (bits_corrected_status == STAT_ECC_NOT_CORRECTED) {
+            ESP_LOGD(TAG, "copy, ecc error");
+            return -1;
+        }
```
igrr, line 292: Similar question about setting error status here.

BTW, it seems that simply returning -1 without calling dhara_set_error doesn't do anything because dhara doesn't check the return value:

https://github.com/dlbeer/dhara/blob/1b166e41b74b4a62ee6001ba5fab7a8805e80ea2/dhara/journal.c#L856-L861

igrr (commented):
One comment about error handling, LGTM otherwise!

spi_nand_flash/src/dhara_glue.c
RathiSonika: My understanding of the statement is that "if an uncorrectable error occurs, it should return -1 and set err to E_ECC." However, if there’s an ECC error with corrected bits, we are setting E_ECC in err but not returning -1 and this error is used to decide if a data refresh is needed after successful nand_read.  https://github.com/espressif/idf-extra-components/pull/437/files#diff-41c825f56b23bffd93941478281fd19aba3f9ddedda4b61af5c319827d0bdf1eL306

RathiSonika: FYI, a small update was made to the is_ecc_error() API. It’s not related to the above comment—just a restructuring.

RathiSonika: Updated. Now DHARA_E_ECC error will be set only if uncorrectable error occurs.

PR #439, opened 2024-10-26
esp_jpeg: Add support for default Huffman tables

tore-espressif:
Summary:
This PR introduces an enhancement to the JPEG decoder in the ESP-IDF library to enable decoding of JPEG images that do not include Huffman tables. Specifically, this update enables decoding of non-standard JPEG frames, such as those produced by certain USB cameras (e.g., Logitech C270), which omit Huffman tables to save space and instead rely on predefined quantization and color subsampling settings. This PR adds robustness to the JPEG decoder by expanding support for such non-standard formats.
Key Changes

*    JPEG Decoder Update:
        Modified the JPEG decoder to support images without Huffman tables, allowing the library to apply default Huffman tables when none are found in the input data.
        This change is essential for handling image data from specific USB cameras that omit Huffman tables.
*
    Double Block Size Support (16x8):
        Added handling for double-block size in images (16x8 pixels), a structure unique to frames from certain USB cameras, which diverge from the more common 8x8 block size.
        Improved parsing logic to ensure accurate decompression for such formats while maintaining compatibility with standard JPEG images.

*    Test Case Addition:
        Created a new test case for JPEG decompression without Huffman tables to validate decoding accuracy and RGB888 output correctness.
        The test compares decompressed image data to reference RGB888 arrays, allowing a color tolerance of ±16 for minor discrepancies, as certain decoding processes (e.g., Windows-based decoders) might use slight variations in default tables.
        The test validates the decompressed image dimensions and color data to ensure it falls within acceptable margins of error.

Testing and Validation

*    New Test Coverage: The newly added test case simulates the decoding of a JPEG file without Huffman tables and checks output against expected RGB888 data, allowing for a minor margin of color error.
*    Backward Compatibility: No impact on decoding standard JPEG images, as the decoder falls back on default tables only when Huffman tables are absent in the input.

Motivation and Context

Some USB cameras, such as the Logitech C270, output JPEG frames without Huffman tables, which standard JPEG decoders often cannot process. By introducing support for these cases, this PR enables ESP-IDF to handle a broader range of JPEG image formats, improving its utility in applications that interface with USB cameras or other devices producing such non-standard JPEG frames.


esp_jpeg/tjpgd/tjpgd.c
```diff
@@ -906,7 +975,10 @@ static JRESULT mcu_output (
                         if (ix == 8) {
                             py += 64 - 8;    /* Jump to next block if double block heigt */
                         }
-                        pc += ix & 1;               /* Step forward chroma pointer every two pixels */
+                        /* Step forward chroma pointer every two pixels */
+                        if (ix % 2) {
+                            pc++;
+                        }
```
tore-espressif, line 981: I'm offering a beer to anyone how can explain why test 2 is not passing without this change!

tore-espressif:
@suda-morris @igrr PTAL or please help me find someone for review, thank you

FYI @espzav

esp_jpeg/jpeg_default_huffman_table.c
```diff
@@ -0,0 +1,60 @@
...
+// The *_num_bits array always contains exactly 16 elements.
+// Each element represents the number of Huffman codes of a specific length:
+// - The first element corresponds to codes of length 1 bit,
+// - The second element to codes of length 2 bits, and so forth up to 16 bits.
+//
+// The *_values array has a length equal to the sum of all elements in the *_num_bits array,
+// representing the actual values associated with each Huffman code in order.
+
+// Luminance DC Table
+const unsigned char lum_dc_num_bits[16] = {0, 1, 5, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0};
```
igrr, line 19: Would it be better to namespace these variables (`esp_jpeg_...`)?

esp_jpeg/test_apps/main/test_usb_camera_jpg.h
```diff
@@ -0,0 +1,175 @@
+// JPEG encoded frame 160x120, 2632 bytes, no huffman tables, double block size (16x8 pixels)
+
+/*
+This is raw data from Logitech C270 USB camera.
+It was reconstructed to usb_camera.jpg file using hex_to_jpg.py script
+The it was converted to RGB888 array with jpg_to_rgb888_hex.py
+*/
+
+const unsigned char jpeg_no_huffman[] = {
```
igrr, line 9: Instead of doing conversion from jpg files to C header, you can simply include the jpg file into the binary. PTAL at https://github.com/espressif/idf-extra-components/blob/ca9569400ed009151f3def6117a3c02f85d0e96b/libpng/test_apps/main/CMakeLists.txt#L5 and https://github.com/espressif/idf-extra-components/blob/ca9569400ed009151f3def6117a3c02f85d0e96b/libpng/test_apps/main/test_libpng.c#L12-L13

esp_jpeg/test_apps/sdkconfig.defaults
```diff
@@ -2,3 +2,5 @@
 # Espressif IoT Development Framework (ESP-IDF) 5.4.0 Project Minimal Configuration
 #
 CONFIG_ESP_TASK_WDT_INIT=n
+CONFIG_JD_USE_ROM=n
```
igrr, line 5: Should we also test with CONFIG_JD_USE_ROM=y?

esp_jpeg/Kconfig
```diff
@@ -66,4 +66,15 @@ menu "JPEG Decoder"
         config JD_FASTDECODE_TABLE
             bool "+ Table conversion for huffman decoding (wants 6 << HUFF_BIT bytes of RAM)"
         endchoice
+
+    config JD_DEFAULT_HUFFMAN
```
igrr, line 70: Should we add a line to README.md to mention this new option? It' not easily discoverable since CONFIG_JD_USE_ROM is enabled by default, so CONFIG_JD_DEFAULT_HUFFMAN is not visible in the menu.

igrr (approved):
Overall LGTM, left a few minor comments.

esp_jpeg/tjpgd/tjpgd.c
igrr: I might be missing something, but on first look this appears to be a compiler bug:

```
# a5 is pc, a4 is ix
=> 0x42007aca <+266>:	extui	a5, a4, 0, 1
   0x42007acd <+269>:	addx2	a5, a5, a5
```

`extui` overwrites the value in a5, so `pc` is zero after this instruction. The correct code would look like this:

```
  extui	rd, a4, 0, 1
  addx2	a5, a5, rd
```

where `rd` is another register (not a4 or a5).

esp_jpeg/test_apps/sdkconfig.defaults
tore-espressif: This repository is not setup for multiple configurations (sdkcongi.ci.something) as in esp-idf. I tried to add some minimal code to add this feature but it broke sh2lib examples.
So I think it is out of scope of this PR

tore-espressif:
Thank you for the review @igrr ! I implemented your comments and also tested on esp-idf examples that use esp_jpeg

tore-espressif:
> Thank you for the review @igrr ! I implemented your comments and also tested on esp-idf examples that use esp_jpeg

@igrr Also tested with USB camera. Please let me know if you want to have a second look or I can merge, thanks!

esp_jpeg/tjpgd/tjpgd.c
igrr: I think we can resolve this for now. I am running c-reduce to make a smaller reproducer for the issue before reporting this to toolchain team.

.idf_build_apps.toml
```diff
@@ -1,5 +1,3 @@
-target = "all"
-paths = "."
```
igrr, line 2: Were these changes necessary for some reason? They are part of a commit titled "review changes", so hard to understand why they are removed.

tore-espressif: With this change you can use `idf-build-apps` also locally eg `idf-build-apps find -t esp32s2 -p some_path`

Without the change, the CLI options -t and -p are overriden byt the .toml file

tore-espressif: Originally I added the 
```
target = "all"
paths = "."
```
part because in some older versions of idf-build-apps this was not default behavior

PR #440, opened 2024-10-31
ci(lcd_qemu_rgb_panel): test example in CI

igrr:
This PR adds example test for lcd_qemu_rgb_panel example. It illustrates the approach which other UI-related projects can use in CI with pytest and QEMU.

Closes https://github.com/espressif/idf-extra-components/issues/423.


igrr:
Looks like the latest QEMU releases were built without pixman support, so `screendump` command isn't included.

Will put this on hold until we fix this in the next QEMU release.

PR #441, opened 2024-10-31
feat(freetype): update to 2.13.3

igrr:
This PR updates freetype from 2.13.0 to 2.13.3.

Among other changes, this fixes CMake warning:
```
CMake Deprecation Warning at idf-extra-components/freetype/freetype/CMakeLists.txt:113 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.
  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```


igrr:
@suda-morris @espressif2022 PTAL at this update

suda-morris (approved):

PR #444, opened 2024-11-05
feat(esp_jpeg): Add configuration option for working buffer size

tore-espressif:
# Checklist

- [x] CI passing

# Change description
Add configuration option for working buffer size

### DIFF
--- a/esp_jpeg/include/jpeg_decoder.h
+++ b/esp_jpeg/include/jpeg_decoder.h
@@ -49,6 +49,13 @@ typedef struct esp_jpeg_image_cfg_s {
     } flags;
 
     struct {
+        void *working_buffer;       /*!< If set to NULL, a working buffer will be allocated in esp_jpeg_decode().
+                                         Tjpgd does not use dynamic allocation, se we pass this buffer to Tjpgd that uses it as scratchpad */
+        size_t working_buffer_size; /*!< Size of the working buffer. Must be set it working_buffer != NULL.
+                                         Default size is 3.1kB or 65kB if JD_FASTDECODE == 2 */
+    } advanced;
+
+    struct {
         uint32_t read;  /*!< Internal count of read bytes */
     } priv;
 } esp_jpeg_image_cfg_t;
--- a/esp_jpeg/jpeg_decoder.c
+++ b/esp_jpeg/jpeg_decoder.c
@@ -79,13 +79,21 @@ esp_err_t esp_jpeg_decode(esp_jpeg_image_cfg_t *cfg, esp_jpeg_image_output_t *im
     assert(cfg != NULL);
     assert(img != NULL);
 
-    workbuf = heap_caps_malloc(JPEG_WORK_BUF_SIZE, MALLOC_CAP_DEFAULT);
-    ESP_GOTO_ON_FALSE(workbuf, ESP_ERR_NO_MEM, err, TAG, "no mem for JPEG work buffer");
+    const bool allocate_buffer = (cfg->advanced.working_buffer == NULL);
+    const size_t workbuf_size = allocate_buffer ? JPEG_WORK_BUF_SIZE : cfg->advanced.working_buffer_size;
+    if (allocate_buffer) {
+        workbuf = heap_caps_malloc(JPEG_WORK_BUF_SIZE, MALLOC_CAP_DEFAULT);
+        ESP_GOTO_ON_FALSE(workbuf, ESP_ERR_NO_MEM, err, TAG, "no mem for JPEG work buffer");
+    } else {
+        workbuf = cfg->advanced.working_buffer;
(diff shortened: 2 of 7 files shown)


tore-espressif:
@igrr @espzav One more little feature for esp_jpeg, PTAL!

esp_jpeg/include/jpeg_decoder.h
```diff
@@ -48,6 +48,12 @@ typedef struct esp_jpeg_image_cfg_s {
         uint8_t swap_color_bytes: 1; /*!< Swap first and last color bytes */
     } flags;
 
+    struct {
+        size_t working_buffer_size; /*!< Size of working memory for the decoder. Set to 0 for default value.
+                                         Default size is 3.1kB or 65kB if JD_FASTDECODE == 2
+                                         Tjpgd does not use dynamic allocation, se we pass this buffer to Tjpgd that uses it as scratchpad */
```
igrr, line 54:
```suggestion
                                         Tjpgd does not use dynamic allocation, so we pass this buffer to Tjpgd that uses it as scratchpad */
```

esp_jpeg/jpeg_decoder.c
```diff
@@ -79,13 +79,14 @@ esp_err_t esp_jpeg_decode(esp_jpeg_image_cfg_t *cfg, esp_jpeg_image_output_t *im
     assert(cfg != NULL);
     assert(img != NULL);
 
-    workbuf = heap_caps_malloc(JPEG_WORK_BUF_SIZE, MALLOC_CAP_DEFAULT);
+    const size_t workbuf_size = cfg->advanced.working_buffer_size ? cfg->advanced.working_buffer_size : JPEG_WORK_BUF_SIZE;
+    workbuf = heap_caps_malloc(workbuf_size, MALLOC_CAP_DEFAULT);
```
igrr, line 83: Just an idea, you can also add `alloc_caps` next to `working_buffer_size`, so that the user can easily move the buffer to PSRAM if needed.

Or, `working_buffer_ptr` — which, if set, will be used instead of allocating the new buffer.

igrr (approved):

esp_jpeg/include/jpeg_decoder.h
tore-espressif: I checked memory allocations inside Tjpgd and the required working buffer size is sum of the following:
```
JD_SZBUF bytes for input buffer
256 bytes for de-quantization table
16 bytes for bit distribution table
5 x number_of_code_words for decoding
up to 256 bytes for decoding scratchpad
(number_of_blocks (typical 8x8 or 16x8) + 2) x 64 bytes
```
So it only depends on the JPG image. There is no way to predict the required working buffer size. Image size independent.

Moreover, a typical user has no clue what these values mean (I didn't until few weeks ago :D) 

Taking this into consideration, the best advice is 'If your decoder returns `JDR_MEM1` or `JDR_MEM2` increase the working buffer size 🤷

esp_jpeg/jpeg_decoder.c
tore-espressif: Thank you for the review @igrr , I'll implement it

I was also thinking about refactoring the memory allocation pool in Tjpgd. So we use dynamic allocation instead of this pre-allocated memory pool... maybe in next release :)

esp_jpeg/include/jpeg_decoder.h
igrr: Can we split the process of decoding the image into more steps

1. Load the image header
2. Application calls `get_required_buffer_size(image_handle)` and we calculate the size based on the information above. Application allocates the buffer of the required size
2. Decode the rest of the image with the application-provided buffer.

Similar to how this is done in libpng:
https://github.com/espressif/idf-extra-components/blob/9634f8d6c04c53d82a2036ed6800d357d89c0d5c/libpng/test_apps/main/test_libpng.c#L30-L42

tore-espressif: Yes, that would be the best. It requires more work than this simple hack, though... I'll give it more time

igrr: If the current PR already helps you achieve your goal, I don't mind merging it as is. We can leave this for later.

espzav (approved):
@tore-espressif LGTM
Thank you for changes in esp_jpeg. Do we want still update this component, when we have esp_jpeg_new?

tore-espressif:
> Thank you for changes in esp_jpeg. Do we want still update this component, when we have esp_jpeg_new?

I wanted to use `esp_jpeg_new` first, but it is closed source, so I could not debug and fix my problems there easily.

In case `esp_jpeg_new` does not add HW JPEG decoder support on P4, I'd like to add HW JPEG decoder support to esp_jpeg.

PR #447, opened 2024-11-11
feat(spi_nand_flash): expose lower-level API and make it usable without Dhara

RathiSonika:
spi_nand_flash
└── src
    ├── nand.c                         : Defines all public interfaces; includes definitions for spi_nand_flash.h
    ├── nand_gigadevice.c     : Initializes chip-specific settings for GigaDevice
    ├── nand_micron.c            : Initializes chip-specific settings for Micron
    ├── nand_winbond.c         : Initializes chip-specific settings for Winbond
    ├── nand_alliance.c           : Initializes chip-specific settings for Alliance
    ├── spi_nand_oper.c       : Contains lower-level SPI communication APIs
    ├── nand_impl.c              : Implements lower-level APIs such as nand_erase, nand_mark_bad, nand_prog, etc.
    ├── nand_impl_wrap.c    : Implements wrapper functions for lower-level APIs implemented in nand_impl.c
    └── dhara_glue.c          :  Initializes and defines Dhara-related APIs and registers function pointers in nand.c. Only 
                                                   dhara_glue.c depends on the Dhara library, making it easy to replace with another library if 
                                                   needed, allowing the code to function independently of Dhara.

Refactoring nand.c and the chip-specific settings is possible; however, this PR primarily focuses on exposing lower-level APIs and enabling easy replacement of the Dhara library in the future, if needed. 

# Checklist
 
- [x] Code cleanup
- [x] CI passing
 
# Change description
 
1. Expose lower level API such as nand_erase, nand_mark_bad, nand_prog etc. and make it usable without Dhara
2. Fixed an issue in the nand_mark_bad() API. After erasing a block, it now correctly waits for the erase_block_delay time. Previously, the function was failing to mark the block as bad due to this missing delay.


pacucha42:
Generally LGTM, but why do you place spi_nand_api.h into private includes? Isn't this PR about exposing the APIs?

spi_nand_flash/include/spi_nand_flash.h
```diff
@@ -47,7 +52,7 @@ esp_err_t spi_nand_flash_init_device(spi_nand_flash_config_t *config, spi_nand_f
  * @param sector_id The id of the sector to read.
  * @return ESP_OK on success, or a flash error code if the read failed.
  */
-esp_err_t spi_nand_flash_read_sector(spi_nand_flash_device_t *handle, uint8_t *buffer, dhara_sector_t sector_id);
+esp_err_t spi_nand_flash_read_sector(spi_nand_flash_device_t *handle, uint8_t *buffer, sector_t sector_id);
```
igrr, line 55: One thing that is a bit confusing is that this header includes:
- functions which operate on top of the wear levelling layer (dhara) — like `spi_nand_flash_read_sector`
- functions which operate directly with the NAND storage — `spi_nand_flash_is_bad_block`.

I think it would be better to separate these two. You kind of already do this by having a header file spi_nand_api.h for these lower level functions (which is, a bit confusingly, not part of the public API, despite the file name).

spi_nand_flash/src/spi_nand_api.c
```diff
@@ -0,0 +1,353 @@
+/*
```
igrr, line 1: Not sure if we should keep the name of this file as "spi_nand_api", mostly because it doesn't implement this component's API but rather the internal functions.

spi_nand_flash/priv_include/nand.h
```diff
@@ -0,0 +1,76 @@
...
+    spi_nand_flash_config_t config;
+    spi_nand_chip_t chip;
+    const spi_nand_ops *ops;
+    uint8_t *work_buffer;
+    uint8_t *read_buffer;
+    SemaphoreHandle_t mutex;
+};
+
+esp_err_t register_nand_dev(spi_nand_flash_device_t *handle);
+esp_err_t unregister_nand_dev(spi_nand_flash_device_t *handle);
```
igrr, line 72: Even though these are not public functions, let's namespace them to avoid link time conflicts. E.g. `nand_register_dev`, `nand_unregister_dev` (or whichever prefix you pick for the non-public functions in this component)

spi_nand_flash/src/dhara_glue.c
```diff
@@ -3,329 +3,249 @@
...
-#include "freertos/FreeRTOS.h"
-#include "freertos/task.h"
 #include "spi_nand_oper.h"
-#include "spi_nand_flash.h"
+#include "spi_nand_api.h"
 #include "nand.h"
 
-#define ROM_WAIT_THRESHOLD_US 1000
+static struct dhara_nand dhara_nand[NAND_MAX_DEVICES];
+static struct dhara_map dhara_map[NAND_MAX_DEVICES];
```
igrr, line 21:
```suggestion
static struct dhara_map s_dhara_map[NAND_MAX_DEVICES];
```

spi_nand_flash/src/dhara_glue.c
```diff
@@ -3,329 +3,249 @@
...
+static spi_nand_flash_device_t *s_handle_t[NAND_MAX_DEVICES] = {NULL};
+static uint8_t s_handle_index;
+static _lock_t s_handle_index_lock;
```
igrr, line 24: Can we no longer use the "containerof" trick to get the `spi_nand_flash_device_t` handle from `const struct dhara_nand *n`? As I understand, you need to store the handle index for the duration of dhara call to know which handle the callbacks are related to... Seems like the original "containerof" approach solved that nicely without the extra locks.

igrr (commented):
@RathiSonika LGTM overall, looks like a nice improvement. I left a few comments for discussion.

spi_nand_flash/src/dhara_glue.c
igrr: Oh I see now. That's because you no longer put the dhara_map into spi_nand_flash_device_t. I think you can still do that, via an intermediate holder structure.

This allows you to get rid of all the static members and also NAND_MAX_DEVICES for good measure :)


```c++
struct spi_nand_flash_device_t {
    uint8_t dev_index;
    spi_nand_flash_config_t config;
    spi_nand_chip_t chip;
    const spi_nand_ops *ops;
    void* ops_ctx;   // new member
    uint8_t *work_buffer;
    uint8_t *read_buffer;
    SemaphoreHandle_t mutex;
};

typedef struct {
    struct dhara_nand dhara_info;
    spi_nand_flash_device_t * parent_handle;
} spi_nand_flash_dhara_ctx_t;

// in dhara_init:
static esp_err_t dhara_init(spi_nand_flash_device_t *handle)
{
    // create a holder structure for dhara context
    spi_nand_flash_dhara_ctx_t * dhara_ctx = calloc(1, sizeof(spi_nand_flash_dhara_ctx_t));
    // save the holder inside the device structure
    handle->ops_ctx = dhara_ctx;
    // store the pointer back to device structure in the holder stucture
    dhara_ctx->parent_handle = handle;
    dhara_ctx->dhara_info->log2_page_size = handle->chip.log2_page_size;
    // etc


// in dhara callbacks:
int dhara_nand_is_bad(const struct dhara_nand *n, dhara_block_t b)
{
    // get spi_nand_flash_dhara_ctx_t using containerof
    spi_nand_flash_dhara_ctx_t *dhara_ctx = __containerof(n, spi_nand_flash_dhara_ctx_t, dhara_info);
    // get spi_nand_flash_device_t 
    spi_nand_flash_device_t *dev_handle = dhara_ctx->parent_handle;
    // use dev_handle....

// in dhara_deinit:
// don't forget to delete holder structure (spi_nand_flash_dhara_ctx_t)
```

RathiSonika: I was aiming to eliminate the dependency on Dhara headers, except in dhara_glue. This approach is the most effective way to achieve this, rather than relying on static variables. Thank you.

spi_nand_flash/include/spi_nand_flash.h
RathiSonika: For debugging and diagnostics, I added the spi_nand_flash_is_bad_block() function to get bad block stats. I didn’t want to expose spi_nand_api directly but still needed to access its information, so I created this public API to serve that purpose.

Or perhaps I could create a separate source file for the diagnostic APIs. This file would use spi_nand_api (which I’ll rename), and these APIs could then be used in debug applications.

igrr: Well, you could expose the "lower-level" API as well, just in a different header file and with a different prefix. This could help in writing test cases. This would also resolve the naming issue around spi_nand_api.h.

RathiSonika: Right now I have removed this API from public header. I will add new header while adding diagnostic application in separate PR.

spi_nand_flash/src/spi_nand_api.c
RathiSonika: Renamed as nand_impl.c. Please let me know for any input.

spi_nand_flash/include/spi_nand_flash.h
RathiSonika: Added nand_impl_wrap.h header which will expose lower-level APIs. PTAL.

RathiSonika:
@igrr PTAL. Thank you!

spi_nand_flash/test_app/main/test_spi_nand_flash.c
```diff
@@ -210,3 +211,76 @@ TEST_CASE("copy nand flash sectors", "[spi_nand_flash]")
     }
     deinit_nand_flash(nand_flash_device_handle, spi);
 }
+
+TEST_CASE("verify mark_bad_block works", "[spi_nand_flash]")
+{
+    spi_nand_flash_device_t *nand_flash_device_handle;
+    spi_device_handle_t spi;
+    setup_nand_flash(&nand_flash_device_handle, &spi);
```
igrr, line 219: As far as I understand, setup_nand_flash also initializes Dhara. Since the rest of the test case verifies lower-level APIs, it would make sense to not initialize Dhara and only call `nand_` (rather than `spi_nand_flash_`) prefixed APIs.

spi_nand_flash/test_app/main/test_app_main.c
```diff
@@ -18,7 +18,7 @@ void setUp(void)
 void tearDown(void)
 {
     esp_reent_cleanup();    //clean up some of the newlib's lazy allocations
-    unity_utils_evaluate_leaks_direct(20);
+    unity_utils_evaluate_leaks_direct(200);
```
igrr, line 21: Do you know what part of this change requires higher leak threshold?

spi_nand_flash/src/nand.c
```diff
@@ -3,172 +3,24 @@
...
-        break;
-    default:
-        return ESP_ERR_INVALID_RESPONSE;
-    }
-    return ESP_OK;
-}
+extern esp_err_t spi_nand_winbond_init(spi_nand_flash_device_t *dev);
+extern esp_err_t spi_nand_alliance_init(spi_nand_flash_device_t *dev);
+extern esp_err_t spi_nand_gigadevice_init(spi_nand_flash_device_t *dev);
+extern esp_err_t spi_nand_micron_init(spi_nand_flash_device_t *dev);
```
igrr, line 23: Can we create a private header file and declare these functions there, instead of using an extern declaration?

The advantage of the common header file is that it will prevent accidentally having incompatible declarations.

spi_nand_flash/include/nand_private/nand_impl_wrap.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include <stdint.h>
+#include "esp_err.h"
+#include "spi_nand_flash.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// These APIs provide direct access to lower-level NAND functions, bypassing the Dhara library.
```
igrr, line 17:
```suggestion
// These APIs provide direct access to lower-level NAND functions, bypassing the Dhara library.
// These functions differ from the similarly named `nand_*` functions in that they also take the mutex for the duration of the call.
```

igrr (approved):
Some final comments, LGTM otherwise.

spi_nand_flash/src/nand.c
RathiSonika: Yes, that makes sense.

spi_nand_flash/test_app/main/test_spi_nand_flash.c
igrr: I see, I'm okay merging this as is, we can refactor this in a follow-up MR.

spi_nand_flash/test_app/main/test_app_main.c
RathiSonika: Honestly, I’m not entirely sure which change is causing this leak, and I couldn’t pinpoint it exactly. It’s an 88-byte leak, actually. I will change the threshold to 100. Need to recheck to find the cause of this leak.

RathiSonika: Got it. The memory leak was in dhara_glue.c. Memory was allocated to spi_nand_flash_dhara_priv_data_t but was not freed.

PR #451, opened 2024-12-02
feat(spi_nand_flash): add diagnostics application for NAND flash

RathiSonika:
# Change description

Added a diagnostics application for NAND flash to facilitate debugging and performance analysis.


spi_nand_flash/examples/nand_flash_debug_app/main/Kconfig.projbuild
```diff
@@ -0,0 +1,9 @@
+menu "SPI Example Configuration"
+
+    config EXAMPLE_FORMAT_IF_MOUNT_FAILED
```
pacucha42, line 3: Why do we need this option? This example doesn't (properly) use any file-system to demonstrate the statistics collection

spi_nand_flash/examples/nand_flash_debug_app/sdkconfig.ci
```diff
@@ -0,0 +1 @@
+CONFIG_EXAMPLE_FORMAT_IF_MOUNT_FAILED=y
```
pacucha42, line 1: See above

spi_nand_flash/examples/nand_flash_debug_app/main/Kconfig.projbuild
RathiSonika: Yes, I forgot to remove it. I will update it. Thank you

spi_nand_flash/src/nand_diag_api.c
```diff
@@ -0,0 +1,158 @@
...
+    }
+    free(pattern_buf);
+    free(temp_buf);
+
+    ESP_LOGI(TAG, "Wrote %" PRIu32 " bytes in %" PRId64 " us, avg %.2f kB/s", sector_size * sec_count, write_time, (float)sector_size * sec_count / write_time * 1000);
+    ESP_LOGI(TAG, "Read %" PRIu32 " bytes in %" PRId64 " us, avg %.2f kB/s\n", sector_size * sec_count, read_time, (float)sector_size * sec_count / read_time * 1000);
+    return ret;
+}
+
+static bool s_is_ecc_exceed_threshold(spi_nand_flash_device_t *handle)
```
pacucha42, line 102: 's_' is only for static variables, not functions

spi_nand_flash/src/nand_diag_api.c
```diff
@@ -0,0 +1,158 @@
...
+    uint32_t num_pages = num_blocks * pages_per_block;
+
+    bool is_free = true;
+    for (uint32_t page = 0; page < num_pages; page++) {
+        ret = nand_wrap_is_free(flash, page, &is_free);
+        if (!is_free) {
+            ret = nand_wrap_get_ecc_status(flash, page);
+            if (ret) {
+                ESP_LOGE(TAG, "Failed to read ecc error for page=%" PRIu32 "", page);
+                return ESP_FAIL;
```
pacucha42, line 141: Why don't we return 'ret' as given by nand_wrap_get_ecc_status()? Btw, using `if (ret) ` for esp_err_t looks confusing to me, as it passes for ret != ESP_OK but it's optically saying the other way round.

I just think that examples should be super-clear, sort of model application

RathiSonika: Updated. Thank you for raising this point.

igrr (changes requested):
I think `nand_read_write_sectors_tp` isn't a great "API" now because it returns information by printing it to console. This means that it can only be used in a test/diagnostic application.

If we keep the function as is, I would suggest:
- Move the function definition to the new diagnostics example
- Remove benchmarking part of the test_app — this is now covered by testing the example.

If you would prefer to keep this as an API, I would suggest some changes:
- Combine most of function arguments into a structure. This way, if you need to add yet another argument or a benchmarking option, that won't be a breaking change.
- Instead of printing the results using ESP_LOG, fill in another "result" structure, return the structure to the caller using an output pointer, and let the caller print the results.

RathiSonika:
Thank you for the review. I had considered this point earlier, but I’m not sure how the throughput would be used from the user’s perspective. This API is primarily designed for testing throughput, similar to how the iperf command works—where you provide arguments, and it prints the throughput. That was the idea behind creating this API. Of course, I can update it, but is it really necessary to return the time intervals required for reading and writing sectors?

igrr:
Would it then be enough to define this benchmarking function within the diagnostics application/example? This way you aren't committing it any specific API — you are free to change the example at any time without making it a breaking change. And as you said, it's not clear how the throughput will be used from users' perspective. My guess that having a benchmarking application is enough and the users won't need to integrate this into their own apps.

spi_nand_flash/examples/nand_flash_debug_app/main/spi_nand_flash_debug_app_main.c
```diff
@@ -0,0 +1,187 @@
...
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "esp_system.h"
+#include "soc/spi_pins.h"
+#include "spi_nand_flash.h"
+#include "nand_diag_api.h"
+#include "nand_private/nand_impl_wrap.h"
```
igrr, line 15: Is it okay that the example depends on a private API?

igrr (approved):
LGTM, just one nitpick.

spi_nand_flash/examples/nand_flash_debug_app/main/spi_nand_flash_debug_app_main.c
RathiSonika: Sorry for the delayed response. 
Ideally, adding private APIs in an example isn’t a good practice. However, this is a debug example where we measure throughput at a lower level by bypassing the Dhara library. For this purpose, I’ve used wrapper functions to access the lower-level APIs. In my opinion, it’s acceptable to use this header here, but it should not be used in the application code.

PR #452, opened 2024-12-02
catch2 updates

igrr:
- Fixed linking flags specified in a way incompatible with clang-based toolchain (esp-18.1.2_20240912)
   - Catch2 still doesn't work with clang-based toolchain if exceptions are enabled, but at least the build succeeds. Remaining issue to be fixed on clang toolchain side.
- Upgraded Catch2 from 3.7.0 to 3.7.1


suda-morris:
closed in favor of https://github.com/espressif/idf-extra-components/pull/847

PR #462, opened 2025-01-02
feat(esp_encrypted_img): Added partial download pytest

hrushikesh430:
1. Added the partial download OTA pytest in pre_encrypted_ota pytest
2. Added the conftest.py for the config fixture
3. Modified the config str pattern

Closes IEC-244

# Description
1. Added the pytest for the partial download in pytest_pre_encrypted_ota.py
2. Right now this PR is for testing purpose to check the behaviour of newly modified config str pattern


hrushikesh430:
Need some modification in build configurations, so closing this PR

PR #470, opened 2025-01-28
fix(network_provisioning): Consider WIFI_REMOTE for provisioning (IEC-260)

vikramdattu:
- In cases, such as ESP32-P4, we do not have native Wi-Fi and it uses other
   mechanisms., such as esp-hosted.[ esp-wifi-remote ](https://github.com/espressif/esp-wifi-remote)is used in this case.
- Provisioning component currently check if Wi-Fi is natively
   enabled for it to compile wifi_provisioning sources.
   This causes link errors when enabling the provisiong for ESP32-P4
- Extend the check for Wi-Fi remote as well, so the sources are compiled. (The config option itself comes from esp-wifi-remote)

# Checklist

- [ ] CI passing

# Change description
This change extends the provisioning support for the devices(such as P4+C6 setup using esp_hosted for connectivity) using Wi-Fi remotes.

- Also bumped up the patch version for network_provisioning component

### DIFF
--- a/network_provisioning/CHANGELOG.md
+++ b/network_provisioning/CHANGELOG.md
@@ -1,4 +1,9 @@
-# 17-March-2024
+# 01-April-2025
+
+- Extend provisioning check for `ESP_WIFI_REMOTE_ENABLED` as well along with existing `ESP_WIFI_ENABLED`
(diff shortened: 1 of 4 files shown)


vikramdattu:
@shahpiyushv @igrr @shahpiyushv @wqx6 PTAL

wqx6 (approved):

mahavirj:
Please add a change-log entry to `CHANGELOG.md`

mahavirj (approved):

PR #476, opened 2025-02-20
feat(spi_nand_flash): Add SIO (full-duplex) and DIO mode support

RathiSonika:
# Change description
spi_nand_flash: Added standard (full-duplex), DIO, DOUT spi mode support.
QIO mode support will be added in another PR

### DIFF
--- a/spi_nand_flash/include/spi_nand_flash.h
+++ b/spi_nand_flash/include/spi_nand_flash.h
@@ -26,8 +26,16 @@ typedef struct spi_nand_flash_device_t spi_nand_flash_device_t;
 #include "nand_linux_mmap_emul.h"
 #endif
 
+/** @brief SPI mode used for reading from SPI NAND Flash */
+typedef enum {
+    SPI_NAND_IO_MODE_SIO = 0,
+    SPI_NAND_IO_MODE_DOUT,
+    SPI_NAND_IO_MODE_DIO,
+} spi_nand_flash_io_mode_t;
+
 /** @brief Structure to describe how to configure the nand access layer.
- @note The spi_device_handle_t must be initialized with the flag SPI_DEVICE_HALFDUPLEX
+ @note For DIO and DOUT mode The spi_device_handle_t must be initialized with the flag SPI_DEVICE_HALFDUPLEX
+ SIO mode can be initialized with half-duplex or full-duplex mode
 */
 struct spi_nand_flash_config_t {
 #ifndef CONFIG_IDF_TARGET_LINUX
@@ -37,6 +45,9 @@ struct spi_nand_flash_config_t {
 #endif
     uint8_t gc_factor;                       ///< The gc factor controls the number of blocks to spare block ratio.
     ///< Lower values will reduce the available space but increase performance
+    spi_nand_flash_io_mode_t io_mode;        ///< set io mode for SPI NAND communication
+    uint8_t flags;                           ///< set flag with SPI_DEVICE_HALFDUPLEX for half duplex communcation, 0 for full-duplex.
+    ///< This flag value must match the flag value in the spi_device_interface_config_t structure.
 };
 
 typedef struct spi_nand_flash_config_t spi_nand_flash_config_t;
--- a/spi_nand_flash/priv_include/spi_nand_oper.h
+++ b/spi_nand_flash/priv_include/spi_nand_oper.h
@@ -11,6 +11,7 @@
 #include <stdint.h>
 #include <esp_err.h>
 #include <driver/spi_master.h>
+#include "nand.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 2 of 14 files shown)


spi_nand_flash/Kconfig
```diff
@@ -7,4 +7,18 @@ menu "SPI NAND Flash configuration"
...
+        default SPI_NAND_IO_MODE_SIO
+        help
+            Mode the nand flash chip using read/write the data
+
+        config SPI_NAND_IO_MODE_SIO
+            bool "SIO"
+        config SPI_NAND_IO_MODE_DOUT
+            bool "DOUT"
+        config SPI_NAND_IO_MODE_DIO
+            bool "DIO"
```
igrr, line 22: As it is done in spi_master driver, could we configure the operation mode at run time, rather than via Kconfig?

For example, it will be easier for you to cover all modes in the test app if you don't have to build a separate version of the test app for each mode.

spi_nand_flash/include/spi_nand_flash.h
```diff
@@ -24,6 +24,7 @@ struct spi_nand_flash_config_t {
     spi_device_handle_t device_handle;       ///< SPI Device for this nand chip.
     uint8_t gc_factor;                       ///< The gc factor controls the number of blocks to spare block ratio.
     ///< Lower values will reduce the available space but increase performance
+    uint8_t flags;                          ///< set flag with SPI_DEVICE_HALFDUPLEX for half duplex communcation.
```
igrr, line 27: I wonder if this can be replaced with an enum (SIO/DIO/DOUT/QIO/QOUT) defined in this component, so that there is less dependency on the way SPI master driver defines its flags.

spi_nand_flash/priv_include/spi_nand_oper.h
```diff
@@ -42,6 +43,8 @@ typedef struct spi_nand_transaction_t spi_nand_transaction_t;
 #define CMD_READ_X2         0x3B
 #define CMD_READ_X4         0x6B
 #define CMD_ERASE_BLOCK     0xD8
+#define CMD_READ_DIO        0xBB
+#define CMD_READ_QIO        0xEB
```
pacucha42, line 47: Cosmetics: probably not needed in this PR, QIO is supposed to come in a standalone PR (see the description)

spi_nand_flash/Kconfig
RathiSonika: Updated. Removed Kconfig options and added enum which can be configured at run time.

pacucha42:
Thanks for the review @igrr

I think the following points may possibly done within this PR, as they are easy to fix

> Btw, I've noticed two issues for a follow-up MR:
> 
> 1. Example readme spi_nand_flash/examples/nand_flash/README.md is empty
> 2. If device detection fails, we currently crash on trying to free an uninitialized mutex. The solution could be to either move mutex initialization before probing the device, or to add a check before freeing:

The test can be possibly resolved in a follow-up task, however, it's also quite brief update
> 
> (Would be nice to cover this scenario in a unit test... for example by trying to initialize NAND flash with the wrong pins)

@RathiSonika could you please try to embed the suggested updates into the current PR, please? Thank you

spi_nand_flash/test_app/main/test_spi_nand_flash.c
```diff
@@ -282,5 +286,31 @@ TEST_CASE("verify nand_prog, nand_read, nand_copy, nand_is_free works", "[spi_na
     }
     free(pattern_buf);
     free(temp_buf);
+}
+
+TEST_CASE("verify nand_prog, nand_read, nand_copy, nand_is_free works in SIO and DIO mode", "[spi_nand_flash]")
```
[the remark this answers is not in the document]
RathiSonika: Updated!

igrr (approved):

spi_nand_flash/include/spi_nand_flash.h
```diff
@@ -17,13 +17,21 @@
...
+
 /** @brief Structure to describe how to configure the nand access layer.
  @note The spi_device_handle_t must be initialized with the flag SPI_DEVICE_HALFDUPLEX
 */
 struct spi_nand_flash_config_t {
     spi_device_handle_t device_handle;       ///< SPI Device for this nand chip.
     uint8_t gc_factor;                       ///< The gc factor controls the number of blocks to spare block ratio.
     ///< Lower values will reduce the available space but increase performance
+    spi_nand_flash_io_mode_t io_mode;        ///< set io mode for SPI NAND communication
+    uint8_t flags;                           ///< set flag with SPI_DEVICE_HALFDUPLEX for half duplex communcation.
```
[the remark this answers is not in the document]
RathiSonika: Actually, the half-duplex or full-duplex flag in the spi_nand_flash_device_t handle informs the driver if to use a buffer with an extra byte or not. This is because, in full-duplex mode, the SPI driver cannot pass a dummy byte, so the first byte should be ignored (as per the waveforms in the datasheet). When adding the device to the SPI bus using spi_bus_add_device(), we set SPI_DEVICE_HALFDUPLEX and if the spi_nand_flash_device_t flag is set to full-duplex, it simply adjusts the buffer to include an extra byte for reading data. However, this does not cause corruption or malfunction. So this is not a breaking change.

PR #480, opened 2025-03-03
fix(spi_nand_flash): fix memory alignment issue occurring on esp32c3

RathiSonika:
# Change description

ESP32-C3 is based on the RISC-V architecture and does not allow uint16_t to be memory-aligned like ESP32. As a result, reading bad blocks caused a memory alignment error. This PR fixes the issue.

### DIFF
--- a/spi_nand_flash/src/nand_impl.c
+++ b/spi_nand_flash/src/nand_impl.c
@@ -87,15 +87,12 @@ esp_err_t nand_is_bad(spi_nand_flash_device_t *handle, uint32_t block, bool *is_
     ESP_GOTO_ON_ERROR(read_page_and_wait(handle, first_block_page, NULL), fail, TAG, "");
 
     // Read the first 2 bytes on the OOB of the first page in the block. This should be 0xFFFF for a good block
-    ESP_GOTO_ON_ERROR(spi_nand_read(handle->config.device_handle, (uint8_t *) &bad_block_indicator, handle->chip.page_size, 2),
+    ESP_GOTO_ON_ERROR(spi_nand_read(handle->config.device_handle, (uint8_t *) handle->read_buffer, handle->chip.page_size, 2),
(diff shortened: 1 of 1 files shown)


spi_nand_flash/src/nand_impl.c
```diff
@@ -232,16 +229,13 @@ esp_err_t nand_is_free(spi_nand_flash_device_t *handle, uint32_t page, bool *is_
     uint16_t used_marker;
 
     ESP_GOTO_ON_ERROR(read_page_and_wait(handle, page, NULL), fail, TAG, "");
-    ESP_GOTO_ON_ERROR(spi_nand_read(handle->config.device_handle, (uint8_t *)&used_marker,
+    ESP_GOTO_ON_ERROR(spi_nand_read(handle->config.device_handle, (uint8_t *)handle->read_buffer,
                                     handle->chip.page_size + 2, 2),
                       fail, TAG, "");
 
+    used_marker = *(uint16_t *)handle->read_buffer;
```
igrr, line 236:
```suggestion
    memcpy(&used_marker, handle->read_buffer, sizeof(used_marker));
```
read_buffer is probably going to be 4-byte aligned, but the compiler doesn't know this. So it's probably better to use an explicit memcpy.

(Same in the other case you have fixed.)

igrr (approved):
LGTM, thanks for the fix! Just one suggestion about memcpy instead of a pointer cast.

By the way, it would be nice to test the component also on ESP32-P4.
P4 has two-level caches, which sometimes put additional requirements upon alignment of data buffers.

spi_nand_flash/src/nand_impl.c
RathiSonika: Updated. Thank you!

RathiSonika:
Sure, I'll test this component on ESP32-P4.

PR #487, opened 2025-03-21
change(freetype): update version number in correct file (IEC-284)

Ashish285:
# Change description
Reverts version upgrade in `sbom_component.yml` and increments in `idf_component.yml`

This should fix the vulnerability scan

### DIFF
--- a/freetype/idf_component.yml
+++ b/freetype/idf_component.yml
@@ -1,4 +1,4 @@
-version: "2.13.3"
+version: "2.13.3~1"
 description: freetype C library
 url: https://github.com/espressif/idf-extra-components/tree/master/freetype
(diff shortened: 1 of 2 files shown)


fhrbata:
@Ashish285 yeah, I just noticed this because of the failed scan job https://github.com/espressif/idf-extra-components/actions/runs/13981959950/job/39148871238. I wanted to create a PR for this, but you were faster. Anwyay the `~` is not allowed in the AV pair in the CPE. It has to be quoted if it need to be used. Unfortunately the validation of CPE is currently very basic, but we have the full validation already implemented in esp-idf-sbom, so I will fix this. Thank you!

PR #489, opened 2025-03-26
feat(spi_nand_flash): Add QIO mode support

RathiSonika:
# Change description
spi_nand_flash: Added QIO and QOUT spi mode support

Performance Comparison across all SPI modes on 40M SPI bus frequency:

```
SIO:
Wrote 131072 bytes in 111909 us, avg 1171.24 kB/s
Read 131072 bytes in 41334 us, avg 3171.05 kB/s

DIO:
Wrote 131072 bytes in 115245 us, avg 1137.33 kB/s
Read 131072 bytes in 27383 us, avg 4786.62 kB/s

DOUT:
Wrote 131072 bytes in 115222 us, avg 1137.56 kB/s
Read 131072 bytes in 27414 us, avg 4781.21 kB/s

QIO:
Wrote 131072 bytes in 90107 us, avg 1454.63 kB/s
Read 131072 bytes in 20829 us, avg 6292.76 kB/s

QOUT:
Wrote 131072 bytes in 90098 us, avg 1454.77 kB/s
Read 131072 bytes in 20848 us, avg 6287.03 kB/s
```

### DIFF
--- a/spi_nand_flash/priv_include/nand.h
+++ b/spi_nand_flash/priv_include/nand.h
@@ -53,6 +53,8 @@ typedef struct {
     uint32_t erase_block_delay_us;
     uint32_t program_page_delay_us;
     ecc_data_t ecc_data;
+    uint8_t has_quad_enable_bit;    // 1 if the chip supports enabling QIO/QOUT mode using bit 0 of 0xB0 (REG_CONFIG), 0 otherwise
+    uint8_t quad_enable_bit_pos;
 } spi_nand_chip_t;
 
 typedef struct {
--- a/spi_nand_flash/priv_include/spi_nand_oper.h
+++ b/spi_nand_flash/priv_include/spi_nand_oper.h
@@ -44,6 +44,7 @@ typedef struct spi_nand_transaction_t spi_nand_transaction_t;
 #define CMD_READ_X4         0x6B
 #define CMD_ERASE_BLOCK     0xD8
 #define CMD_READ_DIO        0xBB
+#define CMD_READ_QIO        0xEB
 
 #define REG_PROTECT         0xA0
 #define REG_CONFIG          0xB0
--- a/spi_nand_flash/src/nand.c
+++ b/spi_nand_flash/src/nand.c
@@ -84,6 +84,24 @@ static esp_err_t detect_chip(spi_nand_flash_device_t *dev)
     }
 }
 
+static esp_err_t enable_quad_io_mode(spi_nand_flash_device_t *dev)
+{
+    uint8_t io_config;
+    esp_err_t ret = spi_nand_read_register(dev, REG_CONFIG, &io_config);
+    if (ret != ESP_OK) {
+        return ret;
+    }
+
+    io_config |= (1 << dev->chip.quad_enable_bit_pos);
+    ESP_LOGD(TAG, "%s: quad config register value: 0x%x", __func__, io_config);
+
+    if (io_config != 0x00) {
+        ret = spi_nand_write_register(dev, REG_CONFIG, io_config);
+    }
+
+    return ret;
+}
+
 static esp_err_t unprotect_chip(spi_nand_flash_device_t *dev)
 {
     uint8_t status;
@@ -132,6 +150,11 @@ esp_err_t spi_nand_flash_init_device(spi_nand_flash_config_t *config, spi_nand_f
     ESP_GOTO_ON_ERROR(detect_chip(*handle), fail, TAG, "Failed to detect nand chip");
 #endif
     ESP_GOTO_ON_ERROR(unprotect_chip(*handle), fail, TAG, "Failed to clear protection register");
+
+    if (((*handle)->config.io_mode ==  SPI_NAND_IO_MODE_QOUT || (*handle)->config.io_mode ==  SPI_NAND_IO_MODE_QIO)
+            && (*handle)->chip.has_quad_enable_bit) {
+        ESP_GOTO_ON_ERROR(enable_quad_io_mode(*handle), fail, TAG, "Failed to enable quad mode");
+    }
 
     (*handle)->chip.page_size = 1 << (*handle)->chip.log2_page_size;
     (*handle)->chip.block_size = (1 << (*handle)->chip.log2_ppb) * (*handle)->chip.page_size;
--- a/spi_nand_flash/src/spi_nand_oper.c
+++ b/spi_nand_flash/src/spi_nand_oper.c
@@ -99,6 +99,34 @@ esp_err_t spi_nand_read_page(spi_nand_flash_device_t *handle, uint32_t page)
     return spi_nand_execute_transaction(handle, &t);
 }
 
+static esp_err_t spi_nand_quad_read(spi_nand_flash_device_t *handle, uint8_t *data, uint16_t column, uint16_t length)
+{
+    uint32_t spi_flags = SPI_TRANS_MODE_QIO;
+    uint8_t cmd = CMD_READ_X4;
+    uint8_t dummy_bits = 8;
+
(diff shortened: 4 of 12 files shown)


spi_nand_flash/src/nand.c
```diff
@@ -84,6 +84,24 @@ static esp_err_t detect_chip(spi_nand_flash_device_t *dev)
...
+static esp_err_t enable_quad_io_mode(spi_nand_flash_device_t *dev)
+{
+    uint8_t io_config;
+    esp_err_t ret = spi_nand_read_register(dev, REG_CONFIG, &io_config);
+    if (ret != ESP_OK) {
+        return ret;
+    }
+
+    io_config |= QEbit;
+    ESP_LOGD(TAG, "%s: quad config register value: %d\n", __func__, io_config);
```
igrr, line 96:
```suggestion
    ESP_LOGD(TAG, "%s: quad config register value: %d", __func__, io_config);
```

or even `%x`, register values are more readable in hex?

spi_nand_flash/priv_include/spi_nand_oper.h
```diff
@@ -57,6 +58,8 @@ typedef struct spi_nand_transaction_t spi_nand_transaction_t;
 #define STAT_ECC1           1 << 5
 #define STAT_ECC2           1 << 6
 
+#define QEbit               1 << 0
```
igrr, line 61: Question, is this bit always the same for NAND flash chips?

For NOR flash chips, the specific bit we need to set differs depending on the vendor and chip model.

spi_nand_flash/priv_include/nand.h
```diff
@@ -53,6 +53,7 @@ typedef struct {
     uint32_t erase_block_delay_us;
     uint32_t program_page_delay_us;
     ecc_data_t ecc_data;
+    uint8_t is_quad_enable_bit;
```
igrr, line 56:
```suggestion
    uint8_t has_quad_enable_bit;    // 1 if the chip supports enabling QIO/QOUT mode using bit 0 of 0xB0 (REG_CONFIG), 0 otherwise
```
Just a suggestion to make it easier to understand what this field means.

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -185,12 +215,20 @@ esp_err_t spi_nand_program_execute(spi_nand_flash_device_t *handle, uint32_t pag
 
 esp_err_t spi_nand_program_load(spi_nand_flash_device_t *handle, const uint8_t *data, uint16_t column, uint16_t length)
 {
+    uint8_t cmd = CMD_PROGRAM_LOAD;
+    uint32_t spi_flags = 0;
+    if (handle->config.io_mode == SPI_NAND_IO_MODE_QOUT || handle->config.io_mode == SPI_NAND_IO_MODE_QIO) {
+        cmd = CMD_PROGRAM_LOAD_X4;
```
igrr, line 221: Do I understand it correctly that writes, unlike reads, don't support transferring address on multiple data lines?

spi_nand_flash/test_app/main/test_spi_nand_flash.c
```diff
@@ -68,7 +68,7 @@ static void setup_chip(spi_device_handle_t *spi, uint8_t flags)
     setup_bus(HOST_ID);
 
     spi_device_interface_config_t devcfg = {
-        .clock_speed_hz = 40 * 1000 * 1000,
+        .clock_speed_hz = 20 * 1000 * 1000,
```
igrr, line 71: Curious, is this change related to the addition of QIO mode? Does QIO mode not work at 40 MHz?

igrr (commented):
LGTM, just a few minor comments. Could you please add some performance numbers (comparing QIO and DIO, for instance) to the MR description?

spi_nand_flash/priv_include/spi_nand_oper.h
RathiSonika: I referred to the datasheets of the NAND flash chips we currently support, considering only the vendor. For Alliance and GigaDevice, the bit is the same, while for the other two vendors, it is a reserved bit. I wasn’t aware of the NOR flash configurations. Also, if this varies with the chip model as well, I think it’s better to make this setting chip-specific, so I’ll update it accordingly. Thank you!

spi_nand_flash/src/spi_nand_oper.c
RathiSonika: Yes, this operates like QOUT mode rather than QIO mode. The address will be transmitted using a single line.

spi_nand_flash/test_app/main/test_spi_nand_flash.c
RathiSonika: Well, with 40M, it wasn’t working on my setup, possibly due to hardware limitations. Length of the wires might be causing performance issues at higher frequencies in QIO mode. However, there shouldn’t be an issue with 40M. I’ll revert this change and check if it runs on the runner.

spi_nand_flash/priv_include/spi_nand_oper.h
RathiSonika: Updated!

spi_nand_flash/src/nand.c
```diff
@@ -133,6 +151,11 @@ esp_err_t spi_nand_flash_init_device(spi_nand_flash_config_t *config, spi_nand_f
 #endif
     ESP_GOTO_ON_ERROR(unprotect_chip(*handle), fail, TAG, "Failed to clear protection register");
 
+    if (((*handle)->config.io_mode ==  SPI_NAND_IO_MODE_QOUT || (*handle)->config.io_mode ==  SPI_NAND_IO_MODE_QIO)
+            && (*handle)->chip.has_quad_enable_bit) {
+        ESP_GOTO_ON_ERROR(enable_quad_io_mode(*handle), fail, TAG, "Failed to enable quad mode");
+    }
```
igrr, line 157: Do we need to make sure the quad enable bit is cleared if io_mode is not QIO/QOUT and the chip supports quad I/O?

Imagine a hypothetical scenario:
- Application has been running on a device, which has an SPI NAND flash connected, and has been using QIO mode
- An OTA update is sent to the device. In the new version of the application, the mode has been set to DIO. (Unlikely, but not impossible...)
- After an OTA update, the ESP chip restarts and boots into the new app
- SPI NAND flash doesn't get reset (because it's not powered from the ESP) and is still in QIO mode
- The app boots up again, tries to communicate with the NAND flash chip, and fails, because the chip expects the data to come on 4 lines, while the application is already using DIO.

i.e. something along these lines
```
if chip.has_quad_enable_bit:
   if io_mode is in QOUT, QIO:
      set quad enable bit
   else
      clear quad enable bit
else if io_mode in QOUT, QIO:
  error: quad mode requested but the chip doesn't support it.
```

igrr (approved):
LGTM, hopefully just one last thing about disabling QIO bit...

spi_nand_flash/src/nand.c
RathiSonika: Interesting scenario. 
I tested a case where the NAND flash is configured in DIO mode with the QE bit enabled. There is no performance issue since setting the QE bit is mandatory only for QIO mode. If the bit is set, it does not cause any issues because the master still sends DIO mode commands (read/write) to the SPI NAND flash. However, to avoid any unforeseen consequences, it is safer to disable this bit when not using QIO mode. I will update this accordingly!

igrr: > If the bit is set, it does not cause any issues because the master still sends DIO mode commands (read/write) to the SPI NAND flash

Ah yes, I understand now. I forgot that the commands are different between the modes. All fine then!

PR #493, opened 2025-04-08
Add libjpeg-turbo component (IEC-289)

dmitry1945:
# Checklist

- [ ] CI passing

# Change description
This component ports libjpeg-turbo library into the esp-idf.


libjpeg-turbo/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+
+version: "1.1.0~1"
+description: Jpeg-turbo port to ESP
+url: https://github.com/espressif/idf-extra-components/tree/master/libjpeg-turbo
+dependencies:
+  idf: ">=5.4.0"
```
suda-morris, line 6: should also add the `sbom` manifest file

suda-morris, line 6: why it's only support 5.4? If the user only uses ESP32-S3, should we relax the dependency?

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+    USES_TERMINAL_BUILD TRUE
+    # Arguments to pass to jpeg-turbo CMake invocation:
+    CMAKE_ARGS
+        -DCMAKE_C_FLAGS=${c_flags}
+        -DCMAKE_CXX_FLAGS=${cxx_flags}
+        ${opt_args}
+        -DCMAKE_INSTALL_PREFIX=${BINARY_DIR}/install
+        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+        -DWITH_TURBOJPEG=NO
+        -DWITH_SIMD=NO
```
suda-morris, line 107: do we plan to add our SIMD acceleration?

libjpeg-turbo/idf_component.yml
dmitry1945: Hi @suda-morris I can remove it. Now it is here just to pass the tests.

libjpeg-turbo/CMakeLists.txt
dmitry1945: @suda-morris no, it's parameter from jpeg-turmo cmake. And I pass it as NO to build it.

libjpeg-turbo/idf_component.yml
igrr: Yes, I think you can

libjpeg-turbo/CMakeLists.txt
igrr: I think the question from Morris was more about whether we plan to add our own changes/patches to use PIE instructions to accelerate jpeg-turbo, later.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -55,6 +55,7 @@ body:
         - supertinycron
         - thorvg
         - zlib
+        - libjpeg-turbo
```
igrr, line 58: Since library names are automatically prefixed with `lib` when building, I would suggest naming the component `jpeg-turbo`. The resulting library will be `libjpeg-turbo.a`.

igrr: Ah, the upstream repo is called the same way... Okay, let's follow the upstream naming then.

libjpeg-turbo/.build-test-rules.yml
```diff
@@ -0,0 +1,4 @@
+libjpeg-turbo/examples/hello_jpeg:
+  enable:
+    - if: (IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR >= 4) and (IDF_TARGET in ["esp32", "esp32p4"])
```
igrr, line 3: We don't have the P4 target in CI of idf-extra-components, so the "esp32p4" here doesn't do anything.

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+    set(opt_args -DCMAKE_BUILD_TYPE=Relesee)
+elseif(CONFIG_COMPILER_OPTIMIZATION_PERF)
+    set(opt_args -DCMAKE_BUILD_TYPE=Relesee)
+elseif(CONFIG_COMPILER_OPTIMIZATION_NONE)
+    set(opt_args -DCMAKE_BUILD_TYPE=Debug )
+else()
+    message(FATAL_ERROR "Unsupported optimization level")
+endif()
+
+set(extra_defines -D_DEFAULT_SOURCE)
```
igrr, line 26: Is this really necessary to build libjpeg-turbo?

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+set(c_flags "${includes} ${extra_defines} ")
+set(cxx_flags "${includes} ${extra_defines} ")
+set(common_flags "-ggdb  -ffunction-sections -fdata-sections -fpermissive -lm")
+
+# set(COMPILE_FLAGS "-ggdb  -ffunction-sections -fdata-sections -fpermissive -lm")
+# set(COMPILE_FLAGS "-ggdb  -ffunction-sections -fdata-sections -fpermissive")
```
igrr, line 33: Are these commented out lines necessary? Please remove if not.

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+if(CONFIG_IDF_TARGET_ARCH_XTENSA)
+    set(assert_flags "${assert_flags} -mlongcalls")
+endif()
+
+if(CONFIG_COMPILER_OPTIMIZATION_DEFAULT)
```
igrr, line 39: Looks like this code block is duplicated, you have another `if(CONFIG_COMPILER_OPTIMIZATION_DEFAULT)` to set `opt_args`. Could you please clean it up?

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+    set(opt_flags "-Og ${common_flags} ${assert_flags}")
+    set(opt_args -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_C_FLAGS_RELEASE=${opt_flags}
+    -DCMAKE_CXX_FLAGS_RELEASE=${opt_flags})
+elseif(CONFIG_COMPILER_OPTIMIZATION_SIZE)
+    # We redefine the flags for Release instead.
```
igrr, line 45: The fact that we are redefining is quite obvious from the lines below, the comment should rather help the reader understand _why_ we are doing this.

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+                 -DCMAKE_C_FLAGS_DEBUG=${opt_flags}
+                 -DCMAKE_CXX_FLAGS_DEBUG=${opt_flags})
+else()
+    set(opt_flags "-O0 ${common_flags} ${assert_flags}")
+    set(opt_args -DCMAKE_BUILD_TYPE=Debug
+                 -DCMAKE_C_FLAGS_DEBUG=${opt_flags}
+                 -DCMAKE_CXX_FLAGS_DEBUG=${opt_flags})
+endif()
+
+idf_build_get_property(python PYTHON)
```
igrr, line 67: Unused?

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+# The list here just to make it common with other components
+set(jpegturbo_libraries
+    libjpeg
+)
+
+set(all_libraries)
+set(all_targets)
+
+foreach(libname ${jpegturbo_libraries})
```
igrr, line 82: Why do we need foreach to loop over jpegturbo_libraries when jpegturbo_libraries only contains one library?

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+        -DWITH_ARITH_DEC=YES
+        -DWITH_ARITH_ENC=YES
+        -DWITH_JPEG8=YES
+        -DWITH_JPEG7=YES
+        -DENABLE_SHARED=NO
+        -DENABLE_STATIC=YES        
+)
+
+add_dependencies(${COMPONENT_LIB} jpegturbo_proj)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
igrr, line 118: Should this not be passed to `CMAKE_ARGS` argument of `ExternalProject_Add` above?

libjpeg-turbo/CMakeLists.txt
```diff
@@ -0,0 +1,134 @@
...
+
+# Attach header files to the component library:
+set_target_properties(${COMPONENT_LIB} PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${BINARY_DIR}/install/include)
+
+# Make sure the subproject is built before the component library:
+add_dependencies(${COMPONENT_LIB} jpegturbo_proj)
+
+# Finally, link the interface library to the component library:
+foreach(libjpeg_target ${all_targets})
```
igrr, line 127: Here as well, seems like an unnecessary foreach.

libjpeg-turbo/examples/hello_jpeg/main/idf_component.yml
```diff
@@ -0,0 +1,10 @@
+## IDF Component Manager Manifest File
+dependencies:
+  espressif/libjpeg-turbo: 
+    version: "^1.1.0"
+    # This line define the local path of the eigen component because this 
+    # example is part of the eigen component. This line is optional.
```
igrr, line 6: Seems to be copy-pasted from `eigen`

libjpeg-turbo/idf_component.yml
```diff
@@ -0,0 +1,6 @@
+
+version: "1.1.0~1"
```
igrr, line 2: Let's set the version to match the upstream version of libjpeg-turob we are using.

igrr: By the way, you seem to be using some commit of libjpeg-turbo from their master branch. There are release tags in libjpeg-turbo repo, we use the release commits unless there is a strong reason not to.

dmitry1945:
@suda-morris I have solved all issues. Shall I do something more?

igrr:
I'll squash the commits when merging.

PR #495, opened 2025-04-30
change(nghttp): adds CVE-2025-30194 to exclude list (IEC-291)

Ashish285:
Adds [CVE-2025-30194](https://nvd.nist.gov/vuln/detail/CVE-2025-30194) to CVE exclude list. This CVE doesn't affect `nghttp2`. It is only being reported because `nghttp2` was mentioned in the CVE description.

### DIFF
--- a/nghttp/idf_component.yml
+++ b/nghttp/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.65.0"
+version: "1.65.0~1"
 description: "nghttp2 - HTTP/2 C Library"
 url: https://github.com/espressif/idf-extra-components/tree/master/nghttp
 dependencies:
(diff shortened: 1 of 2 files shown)


fhrbata:
Hello @Ashish285 , thank you for taking care about this CVE, but since it's not related to the nghttp at all, we can remove it globally in https://github.com/espressif/esp-idf-sbom/blob/master/excluded_cves.yaml. Only CVEs really related to the nghttp need to be included directly in the package's exclude list. I will prepare MR for it.

Ashish285:
> Hello @Ashish285 , thank you for taking care about this CVE, but since it's not related to the nghttp at all, we can remove it globally in https://github.com/espressif/esp-idf-sbom/blob/master/excluded_cves.yaml. Only CVEs really related to the nghttp need to be included directly in the package's exclude list. I will prepare MR for it.

Ok, thanks for the info. In this case we can close this PR.

PR #498, opened 2025-05-05
feat(esp_jpeg): Added option to get image size without decoding it

tore-espressif:
# Change description
Added new public API to get image size without decoding it.

This is very useful for scenarios when we have a JPEG encoded image, but we do not know what resolution does it have, thus we don't know how big output buffer we need for decoding. This scenario is added to test-cases.

This is the last missing feature, before we can integrate this component to esp32-camera, in https://github.com/espressif/esp32-camera/pull/740

### DIFF
--- a/esp_jpeg/jpeg_decoder.c
+++ b/esp_jpeg/jpeg_decoder.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -65,6 +65,7 @@ static uint8_t jpeg_get_color_bytes(esp_jpeg_image_format_t format);
 
 static unsigned int jpeg_decode_in_cb(JDEC *jd, uint8_t *buff, unsigned int nbyte);
 static jpeg_decode_out_t jpeg_decode_out_cb(JDEC *jd, void *bitmap, JRECT *rect);
+static inline uint16_t ldb_word(const void *ptr);
 /*******************************************************************************
 * Public API functions
 *******************************************************************************/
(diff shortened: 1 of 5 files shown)


tore-espressif:
cc @igrr since you reviewed some esp_jpeg PRs in the past :)

esp_jpeg/jpeg_decoder.c
```diff
@@ -119,6 +119,50 @@ esp_err_t esp_jpeg_decode(esp_jpeg_image_cfg_t *cfg, esp_jpeg_image_output_t *im
     return ret;
 }
 
+#define LDB_WORD(ptr)       (unsigned short)(((unsigned short)*((unsigned char*)(ptr))<<8)|(unsigned short)*(unsigned char*)((ptr)+1))
```
igrr, line 122: Perhaps you can use a slightly more readable function
```
static inline uint16_t ldb_word(const void* ptr) {
  uint16_t result;
  memcpy(&result, ptr, sizeof(result);
  return result;
}
```

the compiler will optimize out memcpy replacing it with either an unaligned load or a pair of byte loads, depending on whether the CPU supports unaligned loads or not.

igrr (commented):
May I suggest extending `esp_jpeg_image_output_t` with a `buffer_size` field, and instead of esp_jpeg_get_output_size adding something like esp_jpeg_get_image_info, returning esp_jpeg_get_output_size? Then we can get both the buffer size and image dimensions before actually calling the decode function. 

Similar to what libpng allows doing:
https://github.com/espressif/idf-extra-components/blob/c37fc0ae1c77fd0bce944a37d7f7efb1730bb93c/libpng/test_apps/main/test_libpng.c#L30-L38

tore-espressif:
@igrr Thank you for the review, will update soon!

esp_jpeg/jpeg_decoder.c
tore-espressif: @igrr I ended up with the following implementation to preserve the big-endian to little-endian transformation. All other comments have been incorporated!
``` c
static inline uint16_t ldb_word(const void *ptr)
{
    const uint8_t *p = (const uint8_t *)ptr;
    return ((uint16_t)p[0] << 8) | p[1];
}
```

igrr (approved):

PR #503, opened 2025-05-16
feat(esp_encrypted_img): adds support for ECIES encryption scheme

Ashish285:
# Change description

This MR adds support for ECIES based encryption scheme for pre encrypted OTA. 

This scheme provides following benefits over the previous RSA based encryption scheme

1. No symmetric encryption key is transferred over the communication channel
2. The server still needs to maintain just one key per device/batch: an ECC device wrap public key. 

Note that it is highly recommended to enable Secure Boot with this scheme to

1. Restrict Access to HMAC Peripheral or eFuse Keys 
Secure Boot ensures only signed, verified firmware can access the HMAC peripheral and do the ECC key derivation. 
 
2. Ensure PBKDF2 Parameters Are Untampered 
Some PBKDF2 parameters like salt (to derive ECC device wrap private key) and iterations are embedded in the firmware. 
Secure boot would allow them to be not modified by an attacker.


esp_encrypted_img/Kconfig
```diff
@@ -0,0 +1,27 @@
+menu "Pre Encrypted OTA Configuration"
+
+    choice PRE_ENCRYPTED_OTA_SCHEME
+        prompt "Pre-encrypted OTA Scheme"
+        default PRE_ENCRYPTED_OTA_USE_RSA
+        help
+            Select the cryptographic scheme to use for pre-encrypted Over-The-Air updates.
+
+        config PRE_ENCRYPTED_OTA_USE_RSA
+            bool "RSA"
```
mahavirj, line 10:
```suggestion
            bool "RSA-3072 encryption"
```

esp_encrypted_img/Kconfig
```diff
@@ -0,0 +1,27 @@
...
+            bool "ECC"
+            help
+                Use Elliptic Curve Cryptography (ECC) for key agreement.
+                The GCM key will be derived using ECDH with a server public key
+                and a device private key (potentially derived via HMAC).
+
+    endchoice
+    
+    config PRE_ENC_OTA_HMAC_EFUSE_KEY_ID
+        int "HMAC EFUSE KEY ID"
```
mahavirj, line 25: This should have `depends on PRE_ENCRYPTED_OTA_USE_ECIES`

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,218 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
 
 This component can help in integrating pre encrypted firmware in over-the-air updates. Additionally, this component can also be used for other use-cases which requires addition of encryption layer for custom data.
 
-
 ## Image Format
 
-![Image Format](https://raw.githubusercontent.com/espressif/idf-extra-components/master/esp_encrypted_img/image_format.png)
+The ESP Encrypted Image format supports two primary cryptographic schemes: RSA-3072 and ECC-256.
```
mahavirj, line 11: Would ECC be apt here - shall we call it as ECIES-P256?

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,218 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
...
+
+![ECC Image Format](ecc_image_format.png)
+
+The header for an image encrypted using the ECC-256 scheme is structured as follows:
+
+```c
+typedef struct {
+    uint8_t magic[4];             // Magic bytes (e.g., derived from SHA256("esp_encrypted_img_ecc"))
+    uint8_t server_pub_key[64];   // Server's uncompressed ECC public key (P-256, X and Y coordinates)
+    uint8_t kdf_salt[16];         // Salt used for KDF (PBKDF2) to derive AES-GCM key
```
mahavirj, line 49: Recommend using 32-byte salt length

esp_encrypted_img/include/esp_efuse_utilities.h
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * @brief Check if the HMAC key is burnt in efuse.
+ *
+ * @param hmac_key_id[in] The HMAC key ID to check.
+ *
+ * @return
+ *      - true If the HMAC key is burnt.
+ *      - false If the HMAC key is not burnt.
+ */
+bool is_hmac_key_burnt_in_efuse(hmac_key_id_t hmac_key_id);
```
mahavirj, line 20: Please use `esp_encrypted` prefix for consistency. Interim, this header could stay private, and not part of `include` dir?

esp_encrypted_img/include/esp_efuse_utilities.h
```diff
@@ -0,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: 2025-2025 Espressif Systems (Shanghai) CO LTD
```
mahavirj, line 2:
```suggestion
 * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
```

esp_encrypted_img/Kconfig
```diff
@@ -0,0 +1,27 @@
...
+            Select the cryptographic scheme to use for pre-encrypted Over-The-Air updates.
+
+        config PRE_ENCRYPTED_OTA_USE_RSA
+            bool "RSA"
+            help
+                Use RSA for encrypting the GCM key.
+                The device will decrypt the GCM key using its private RSA key.
+
+        config PRE_ENCRYPTED_OTA_USE_ECC
+            bool "ECC"
```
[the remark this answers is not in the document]
mahavirj: This should also select `MBEDTLS_HKDF_C`, otherwise build fails locally on my side?

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,218 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
```
mahavirj, line 6: @Ashish285 Please update example README as well

esp_encrypted_img/Kconfig
```diff
@@ -0,0 +1,27 @@
...
+            help
+                Use Elliptic Curve Cryptography (ECC) for key agreement.
+                The GCM key will be derived using ECDH with a server public key
+                and a device private key (potentially derived via HMAC).
+
+    endchoice
+    
+    config PRE_ENC_OTA_HMAC_EFUSE_KEY_ID
+        int "HMAC EFUSE KEY ID"
+        default 2
```
mahavirj, line 26: Alternatively, we can enforce user to set it explicitly

```
     range -1 5
     default -1
``` 
And then `static_assert` in code to ensure that value is within 0 ...5 range. Please see how its handled for `CONFIG_NVS_SEC_HMAC_EFUSE_KEY_ID` case in IDF.

mahavirj:
@AdityaHPatwardhan @Harshal5 PTAL

esp_encrypted_img/test_apps/main/test_mocks.c
```diff
@@ -0,0 +1,29 @@
...
+uint8_t dummy_pbkdf2_output[32] = {
+    0x83, 0x17, 0x93, 0x66, 0x0d, 0xe4, 0x91, 0x33, 0x66, 0xae, 0x1e, 0x37, 0x9b, 0x2c, 0xeb, 0x43,
+    0x17, 0xc8, 0x87, 0x00, 0xcc, 0x07, 0x91, 0xd9, 0x8e, 0x5a, 0x2a, 0x2d, 0x5c, 0x71, 0xaf, 0x66
+};
+
+bool esp_encrypted_is_hmac_key_burnt_in_efuse(hmac_key_id_t hmac_key_id)
+{
+    // Simulate the behavior of checking if the HMAC key is burnt in efuse
+    // For this example, we'll assume that the key is always burnt
+    return true;
```
Harshal5, line 19: It seems like the project supports running tests on specific markers ([reference](https://github.com/espressif/idf-extra-components/blob/master/spi_nand_flash/test_app/pytest_spi_nand_flash.py#L4). Maybe in a follow up MR we could check if running the test on a target with the tag `nvs_encr_hmac` is possible. This could avoid the HMAC and the HMAC-PBKFD2 simulation.

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -118,13 +147,272 @@ static int decipher_gcm_key(const char *enc_gcm, esp_encrypted_img_t *handle)
...
+        mbedtls_ecp_point_free(server_public_point);
+        free(server_public_key);
+        server_public_key = NULL;
+        return NULL;
+    }
+
+cleanup:
+    mbedtls_ecp_group_free(&grp);
+    if (server_public_key) {
+        memset(server_public_key, 0, len + 1);
```
AdityaHPatwardhan, line 277: Is this memset necessary ? ( Since its a public key)

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -312,18 +618,32 @@ esp_err_t esp_encrypted_img_decrypt_data(esp_decrypt_handle_t ctx, pre_enc_decry
     /* falls through */
     case ESP_PRE_ENC_IMG_READ_GCM:
         if (handle->cache_buf_len == 0 && args->data_in_len - curr_index >= ENC_GCM_KEY_SIZE) {
+#if defined(CONFIG_PRE_ENCRYPTED_OTA_USE_RSA)
```
AdityaHPatwardhan, line 621: Can we combine the two sections and handle the logs accordingly?

esp_encrypted_img/test_apps/main/test.c
```diff
@@ -357,3 +399,60 @@ TEST_CASE("Test canceling decryption frees memory", "[encrypted_img]")
...
+#if defined(CONFIG_PRE_ENCRYPTED_OTA_USE_ECIES)
+
+uint8_t server_pub[SERVER_ECC_KEY_LEN] = {
+    0x6b, 0x1e, 0xda, 0x43, 0xaf, 0xc4, 0x1d, 0x44, 0x28, 0x44, 0x5f, 0x06, 0x12, 0x07, 0xe3, 0x85,
+    0x8b, 0x93, 0x0d, 0x48, 0x09, 0xc3, 0xf2, 0x33, 0x3c, 0x04, 0x26, 0x64, 0xd6, 0x14, 0x1c, 0x62,
+    0x24, 0xd2, 0xf5, 0xc8, 0x2e, 0x10, 0x7c, 0xb5, 0x16, 0x2b, 0x4c, 0xd1, 0xa0, 0x1c, 0x93, 0xe4,
+    0x48, 0x2a, 0x03, 0x5d, 0x5e, 0x98, 0xb2, 0x82, 0xf6, 0x85, 0x70, 0x2a, 0x5e, 0x92, 0xd3, 0x6a,
+};
+
+uint8_t kdf_salt[KDF_SALT_SIZE] = {
```
AdityaHPatwardhan, line 412: Hasn't the KDF salt been changed to 32 bytes?

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,267 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
...
-    char auth[16];
-    char extra_header[88];
-} pre_enc_bin_header;
+    uint8_t magic[4];             // Magic bytes (e.g., derived from SHA256("esp_encrypted_img_rsa"))
+    uint8_t enc_gcm_key[384];     // AES-GCM key encrypted with RSA-3072 public key
+    uint8_t iv[16];               // Initialization Vector for AES-GCM
+    uint8_t bin_size[4];          // Size of the original binary (little-endian)
+    uint8_t auth_tag[16];         // AES-GCM authentication tag
+    uint8_t reserved_rsa[88];     // Reserved for future use
+} esp_enc_img_rsa_header_t;
```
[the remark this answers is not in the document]
Ashish285: This naming is only used inside the README.md. Internally we do not have these structs defined. Instead we use `pre_enc_bin_header` which is agnostic of the underlying algorithm

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -118,13 +147,272 @@ static int decipher_gcm_key(const char *enc_gcm, esp_encrypted_img_t *handle)
...
+
+    // Step 2: Convert output to scalar mod curve order
+    ret = mbedtls_mpi_read_binary(ecc_priv_key, hmac_output, HMAC_OUTPUT_SIZE);
+    if (ret != 0) {
+        ESP_LOGE(TAG, "failed\n  ! mbedtls_mpi_read_binary returned -0x%04x\n", (unsigned int) - ret);
+        goto cleanup;
+    }
+
+    // Reduce mod order to ensure valid private key
+    ret = mbedtls_mpi_mod_mpi(ecc_priv_key, ecc_priv_key, &order);
```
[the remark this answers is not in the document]
Ashish285: I think we can remove this mod operation. Replaced it with the `mbedtls_ecp_check_privkey()`

esp_encrypted_img/test_apps/main/test_mocks.c
Ashish285: Yes, the current mocks are a stopgap measure till we have the new CI targets to run the tests

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,267 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
...
+    space:2 plain_text["Plain Text"]:1
+    space:2 aes_enc["Encrypted by AES-GCM key"]:1
+style app_binary fill:#565
+style aes_enc fill:#565
+```
+
+```c
+typedef struct {
+    uint8_t magic[4];             // Magic bytes (e.g., derived from SHA256("esp_encrypted_img_ecc"))
+    uint8_t server_pub_key[64];   // Server's uncompressed ECC public key (P-256, X and Y coordinates)
```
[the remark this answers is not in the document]
Ashish285: We do have space reserved for longer keys. We need to come up with versioning structure for the header to support newer versions. Accordingly we can then update defined sizes based on the header version

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,267 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
...
+    uint8_t kdf_salt[32];         // Salt used for KDF (HKDF) to derive AES-GCM key
+    uint8_t reserved_key_params[288]; // Reserved to make the key parameters block (server_pub_key, kdf_salt, this field) 384 bytes
```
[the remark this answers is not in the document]
Ashish285: 1. Reduces the changes required to handle new parameters. We can keep the new parameters in the same 384 bytes block and rest of the handling (for encryption parameters) can remain the same.
2. This gives us space to support longer key sizes in the future + additional salt parameters if required.

esp_encrypted_img/README.md
```diff
@@ -6,50 +6,267 @@ This component provides an API interface to decrypt data defined in "ESP Encrypt
...
+    uint8_t iv[16];               // Initialization Vector for AES-GCM
+    uint8_t bin_size[4];          // Size of the original binary (little-endian)
+    uint8_t auth_tag[16];         // AES-GCM authentication tag
+    uint8_t reserved_final_padding[88]; // Reserved padding to make the total header 512 bytes
```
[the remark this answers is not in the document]
Ashish285: Let's discuss how we want to do the header versioning. Maybe we can do that in a separate PR as it will go out of the scope of this PR

esp_encrypted_img/ecc_image_format.png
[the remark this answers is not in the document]
Ashish285: Removing this diagram, switched to mermaid based diagrams which will be easier to maintain

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: Not necessary, but still a good idea to memset all the cryptographic artifacts back to 0.

esp_encrypted_img/README.md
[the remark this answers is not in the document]
Harshal5: >  Internally we do not have these structs defined. Instead we use pre_enc_bin_header which is agnostic of the underlying algorithm

Oh okay!!

esp_encrypted_img/Kconfig
```diff
@@ -0,0 +1,37 @@
...
+            select MBEDTLS_HKDF_C
+            bool "ECIES encryption"
+            help
+                Use Elliptic Curve Cryptography (ECC) for key agreement.
+                The GCM key will be derived using ECDH with a server public key
+                and a device private key (potentially derived via HMAC).
+
+    endchoice
+    
+    config PRE_ENC_OTA_HMAC_EFUSE_KEY_ID
```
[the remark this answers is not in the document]
Ashish285: I had it runtime programmable in initial version but we wanted to have compile time check for the efuse key id. That's why moved it to the Kconfig option. Do you think we should move it to be a runtime parameter?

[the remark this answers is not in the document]
mahavirj: Runtime might be better - also allows to use `esp_decrypt_cfg_t`. You may convert the check to runtime in that case.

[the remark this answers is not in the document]
Ashish285: I moved the configuration to runtime and added in `esp_decrypt_cfg_t`. Please have a look again. Thanks!

mahavirj (approved):

PR #506, opened 2025-05-20
fix(spi_nand_flash): Fix build failure in host tests

RathiSonika:
# Change description

This PR fixes the build failures in host tests caused by implicit function declarations. 
Additionally, compiler warnings observed during host test builds have been addressed and removed to ensure cleaner builds.


.build-test-rules.yml
```diff
@@ -80,6 +80,8 @@ spi_nand_flash/test_app:
       reason: The spi_nand_flash component is compatible with IDF version v5.0 and above, due to a change in the f_mkfs API in versions above v5.0, which is not supported in older IDF versions.
 
 spi_nand_flash/host_test:
+  enable:
+    - if: (IDF_VERSION_MAJOR >= 5 and IDF_VERSION_MINOR >= 3) and (IDF_TARGET == linux)
```
igrr, line 84: Did the app end up getting built in CI? I've looked at https://github.com/espressif/idf-extra-components/actions/runs/15390513434/job/43378314110?pr=506 but found only apps build for chip targets.

pytest.ini
```diff
@@ -4,15 +4,22 @@ python_files = pytest_*.py
 
 # set traceback to "short" to prevent the overwhelming tracebacks
 addopts =
-  -s
+  -s -vv
   --embedded-services esp,idf
   --tb short
+  --ignore-glob */managed_components/*
+  --ignore-glob managed_components/*
```
igrr: Given that you've added these arguments here, could you please remove them from https://github.com/espressif/idf-extra-components/blob/f3e54a9ae25ea970ce493e6235bbf16d9088b821/.github/workflows/build_and_run_apps.yml#L165?

igrr (approved):

.build-test-rules.yml
RathiSonika: Yes, it was successfully built in CI https://github.com/espressif/idf-extra-components/actions/runs/15390513434/job/43378307359?pr=506#step:5:13560

PR #508, opened 2025-05-21
feat(1wire): add config option to enable/disable internal pull-up resistor

suda-morris:
If the external device didn't have a pull-up resistor, enable the internal pull-up resistor can help.

Please note, the internal pull-up resistor may not provide enough current, please use an external resistor when possible.


onewire_bus/include/onewire_types.h
```diff
@@ -31,6 +31,11 @@ typedef struct onewire_device_iter_t *onewire_device_iter_handle_t;
  */
 typedef struct {
     int bus_gpio_num; /*!< GPIO number that used by the 1-Wire bus */
+    struct ow_bus_config_flags {
```
Kainarx, line 34: Is it necessary to abbreviate onewire as ow?

suda-morris:
@mahavirj PTAL

PR #511, opened 2025-06-10
Feat esp encrypted img/adds rsa ds peripheral support

Ashish285:
# Checklist

- [ ] CI passing

# Change description


esp_encrypted_img/Kconfig
```diff
@@ -11,6 +11,32 @@ menu "Pre Encrypted OTA Configuration"
...
+                Use DS Peripheral for RSA OTA Encryption Scheme. The DS peripheral is used 
+                for decrypting the AES-GCM key
+        
+        choice PRE_ENCRYPTED_RSA_PADDING_CHOICE
+            depends on PRE_ENCRYPTED_RSA_USE_DS
+            prompt "RSA Padding scheme"
+            default PRE_ENCRYPTED_RSA_PADDING_V1_5
+
+                config PRE_ENCRYPTED_RSA_PADDING_V1_5
+                    bool "Padding v1.5"
```
mahavirj: We can keep the default padding here to stay compliant with the existing RSA scheme. And remove this config choice.

esp_encrypted_img/include/esp_encrypted_img.h
```diff
@@ -68,6 +68,10 @@ typedef void *esp_decrypt_handle_t;
 
 typedef struct {
 #if defined(CONFIG_PRE_ENCRYPTED_OTA_USE_RSA)
+#if defined(CONFIG_PRE_ENCRYPTED_RSA_USE_DS)
+    void *ds_data;                         /*!< Data context for DS */
+    int rsa_padding;                       /*!< Padding type for RSA decryption */
```
mahavirj: I think we can remove this option

esp_encrypted_img/include/esp_encrypted_img.h
```diff
@@ -68,6 +68,10 @@ typedef void *esp_decrypt_handle_t;
 
 typedef struct {
 #if defined(CONFIG_PRE_ENCRYPTED_OTA_USE_RSA)
+#if defined(CONFIG_PRE_ENCRYPTED_RSA_USE_DS)
+    void *ds_data;                         /*!< Data context for DS */
```
mahavirj, line 72: This could be purely dynamic as well - if `ds_data` is set then use DS peripheral approach else traditional RSA approach. Then we can remove `CONFIG_PRE_ENCRYPTED_RSA_USE_DS`

Ashish285: I am not sure about making it dynamic and creating this fallback. If the choice is made during runtime dynamically, then it might be possible (intentionally or not) to fall back to a less secure option. If the choice is compile time, then we can guarantee the security, and if it fails the OTA will not succeed. Let me know what are your thoughts.

esp_encrypted_img/Kconfig
Ashish285: Removed the config options and made v1.5 padding the default

esp_encrypted_img/include/esp_encrypted_img.h
mahavirj: I think that will be applications responsibility to correctly choose the config scheme. At the moment, the newly added config is default disabled, so unless someone reads about it they will continue to use older scheme.

esp_encrypted_img/idf_component.yml
```diff
@@ -4,3 +4,5 @@ url: https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted
 dependencies:
   idf:
     version: ">=4.4"
+  espressif/esp_secure_cert_mgr:
+    version: '*'
```
mahavirj: Lets set rule to allow only minor or patch version. Major version may break the API.

mahavirj:
Let's bump up the minor component version and add IDF version dependency for newly added feature

esp_encrypted_img/idf_component.yml
Ashish285: Added in version check

Ashish285:
> Let's bump up the minor component version and add IDF version dependency for newly added feature

Sure, updated the component version and added in latest IDF version dependency. Thanks

esp_encrypted_img/idf_component.yml
```diff
@@ -1,6 +1,8 @@
-version: "2.5.1"
+version: "2.6.0"
 description: ESP Encrypted Image Abstraction Layer
 url: https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img
 dependencies:
   idf:
-    version: ">=4.4"
+    version: ">=5.5"
```
mahavirj: This will lock in the component for IDF >= 5.5 release. We should support prior releases and take an exception for newly added feature only

esp_encrypted_img/project_include.cmake
```diff
@@ -23,3 +25,24 @@ function(create_esp_enc_img input_file key_file output_file app)
     add_custom_target(enc_bin_target_${name} ALL DEPENDS ${output_file})
     add_dependencies(enc_bin_target_${name} gen_project_binary)
 endfunction()
+
+# function(create_esp_enc_img_secure_cert_data target_chip input_key input_key_algo)
```
Harshal5, line 29: Probably some cleanup?

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -48,8 +59,11 @@ typedef enum {
 
 struct esp_encrypted_img_handle {
 #if defined(CONFIG_PRE_ENCRYPTED_OTA_USE_RSA)
+#if defined(CONFIG_PRE_ENCRYPTED_RSA_USE_DS)
+#else
```
Harshal5:
```suggestion
#if defined(!CONFIG_PRE_ENCRYPTED_RSA_USE_DS)
```

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -24,14 +24,25 @@
...
+#include "esp_idf_version.h"
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5,5,0)
+#include "rsa_dec_alt.h"
+#else
+#error "DS Peripheral is only supported on IDF >= 5.5"
+#endif /* ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5,5,0) */
+
+#endif /* CONFIG_PRE_ENCRYPTED_RSA_USE_DS */
+
+#include "esp_random.h"
```
Harshal5, line 43: nitpick: Could we move this up under the `#if CONFIG_PRE_ENCRYPTED_RSA_USE_DS ` block?

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -95,6 +109,53 @@ typedef struct esp_encrypted_img_handle esp_encrypted_img_t;
...
+        ESP_LOGE(TAG, "failed\n  ! mbedtls_pk_setup_rsa_alt returned -0x%04x\n", (unsigned int) - ret);
+        goto exit;
+    }
+
+    size_t olen = 0;
+    ret = mbedtls_pk_decrypt(&pk, (const unsigned char *)enc_gcm, ENC_GCM_KEY_SIZE,
+                             (unsigned char *)handle->gcm_key, &olen, GCM_KEY_SIZE,
+                             mbedtls_esp_random, NULL);
+    if (ret != 0) {
+        ESP_LOGE(TAG, "failed\n  ! mbedtls_pk_decrypt returned -0x%04x\n", (unsigned int) - ret);
```
Harshal5, line 138: Just a nitpick, but maybe we could keep the format of logging similar to the below one.

edit: Looks like we already have such logs in the component. We could update the component's log formats some time later as well.

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -417,25 +484,20 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
+    mbedtls_mpi_free(&device_private_mpi);
+    if (ret != 0) {
         ESP_LOGE(TAG, "failed\n  ! mbedtls_ecdh_compute_shared returned -0x%04x\n", (unsigned int) - ret);
-        mbedtls_mpi_free(&shared_secret);
-        mbedtls_mpi_init(&shared_secret);
-        memset(&device_private_mpi, 0, sizeof(mbedtls_mpi));
-        mbedtls_mpi_free(&device_private_mpi);
         goto exit;
     }
-    memset(&device_private_mpi, 0, sizeof(mbedtls_mpi));
```
Harshal5, line 429: Do we need the `memset` before freeing?

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -417,25 +484,20 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
 
-    if ((ret = mbedtls_mpi_write_binary(&shared_secret, shared_secret_bytes, sizeof(shared_secret_bytes))) != 0) {
+    ret = mbedtls_mpi_write_binary(&shared_secret, shared_secret_bytes, sizeof(shared_secret_bytes));
+    mbedtls_mpi_free(&shared_secret);
+    if (ret != 0) {
         ESP_LOGE(TAG, "failed\n  ! mbedtls_mpi_write_binary returned -0x%04x\n", (unsigned int) - ret);
-        mbedtls_mpi_free(&shared_secret); // Free shared_secret on error
         goto exit;
     }
-    mbedtls_platform_zeroize(&shared_secret, sizeof(shared_secret));
```
Harshal5, line 437: Similarly, here.

esp_encrypted_img/project_include.cmake
Ashish285: This is currently commented out because esp_secure_cert_mgr currently doesn't allow creating secure cert partition without the port. Once we add in the functionality we should be able to create the secure partition internally and the user doesn't have to create one by themselves.

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: It is used with [ECIES scheme](https://github.com/Ashish285/idf-extra-components/blob/95069b8ea1660598cd1147cb4254219c380ce544/esp_encrypted_img/src/esp_encrypted_img.c#L311) as well

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: Not really required as `mbedtls_mpi_free` internally does `mbedtls_mpi_zeroize_and_free`

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: https://github.com/espressif/idf-extra-components/pull/511#discussion_r2261793659

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: Yeah, I think the idea with logging here is that in case we have mbedtls error code, then it prints out `failed` along with the mbedtls error code, if not, then it just prints out a generic failure statement. Maybe we can revisit it in future and merge these into one pattern

esp_encrypted_img/project_include.cmake
Harshal5: Oh okay got it. Thanks!

esp_encrypted_img/idf_component.yml
```diff
@@ -1,6 +1,11 @@
...
 url: https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img
 dependencies:
   idf:
-    version: ">=4.4"
+    version: ">=5.0"
+  espressif/esp_secure_cert_mgr:
+    version: ">=2.5.1"
+    rules:
+      - if: "idf_version >= 5.3"
+      - if: target not in ["esp32c61"]
```
mahavirj: Why C61 needs a special handling, could we use `SOC_HMAC_SUPPORTED` capability instead?

esp_encrypted_img/idf_component.yml
```diff
@@ -1,6 +1,11 @@
-version: "2.5.1"
+version: "2.6.0"
```
mahavirj, line 1: Please update ChangeLog file

esp_encrypted_img/examples/pre_encrypted_ota/main/pre_encrypted_ota.c
```diff
@@ -115,6 +118,7 @@ void pre_encrypted_ota_task(void *pvParameter)
     ESP_LOGI(TAG, "Starting Pre Encrypted OTA example");
 
     esp_err_t ota_finish_err = ESP_OK;
+    esp_err_t err = ESP_OK;
```
mahavirj: suggest to start with `ESP_FAIL` as initial value

esp_encrypted_img/idf_component.yml
Ashish285: Hi, I tried to use `SOC_HMAC_SUPPORTED` but it seems like using it like this 
`- if: "$CONFIG{SOC_HMAC_SUPPORTED} == True"` doesn't work as the Kconfig option is not defined in this project.

From the [docs](https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/manifest_file.html#kconfig-options):

`Only Kconfig options defined in the ESP-IDF project or its direct dependency components are supported.`

As this config is not defined for targets which lack HMAC support, the build fails with error

```
CMake Warning at /opt/esp/idf/tools/cmake/build.cmake:647 (message):
  WARNING: The following Kconfig variables were used in "if" clauses, but not
  found in any Kconfig file:

      SOC_HMAC_SUPPORTED, introduced by espressif/esp_secure_cert_mgr, defined in /__w/idf-extra-components/idf-extra-components/esp_encrypted_img/idf_component.yml
```

mahavirj: Try dropping a file like https://github.com/espressif/idf-extra-components/blob/master/bdc_motor/.build-test-rules.yml in `esp_encrypted_img` component

esp_encrypted_img/.build-test-rules.yml
```diff
@@ -0,0 +1,15 @@
...
+
+esp_encrypted_img/test_apps:
+  disable:
+    - if: (IDF_VERSION < "5.3" and CONFIG_NAME == "CONFIG_PRE_ENCRYPTED_RSA_USE_DS") or (IDF_TARGET == "esp32c61")
+      temporary: true
+      reason: IDF Version < 5.3 is not supported yet. Skipping c61 because it does not support HMAC peripheral
+  disable_test:
+    - if: (IDF_VERSION < "5.3" and CONFIG_NAME == "CONFIG_PRE_ENCRYPTED_RSA_USE_DS") or (IDF_TARGET == "esp32c61")
+      temporary: true 
+      reason: IDF Version < 5.3 is not supported yet. Skipping c61 because it does not support HMAC peripheral
```
Harshal5: I think we could remove this disable_test rule as it is similar and gets covered in the disable rule itself.

esp_encrypted_img/idf_component.yml
Ashish285: This can work for the pipeline but the build will fail locally for esp32c61. 

I think I found the problem, `esp_secure_cert_mgr` is set to be [disabled](https://github.com/espressif/esp_secure_cert_mgr/blob/main/Kconfig#L6) for targets not supporting HMAC peripheral but this doesn't include esp32c61. So even if the `SOC_HMAC_SUPPORTED` is used (in idf_component.yml or .build-test-rules.yml), it still tries to build and fails

So,
1. Either we add `esp32c61` to `esp_secure_cer_mgr` disabled list
2. Skip `esp32c61` from here and disable the build

Option 1 sounds better to me, let me know what you think

mahavirj: We should update `esp_secure_cert_mgr` to use `SOC_DIG_SIGN_SUPPORTED` in the config option dependency list

esp_encrypted_img/.build-test-rules.yml
Ashish285: Sure, removed it.

PR #512, opened 2025-06-16
feat(esp_encrpte_img): Added the goto ota_end lablel for failure case

hrushikesh430:
# Description

1. Added the goto ota_end lable, if data is not received completely. So, that it calls esp_https_ota_abort api.
2. Although right now, it ultimately falls into `ota_end` label, but to make code more clear, added it.


hrushikesh430:
CC: @AdityaHPatwardhan

hrushikesh430:
Already solved this issue

PR #515, opened 2025-06-19
feat: Add support for QEMU test runner

Ashish285:
This MR adds support for QEMU runner to run tests in the pipeline

### DIFF
--- a/.github/setup_qemu.sh
+++ b/.github/setup_qemu.sh
@@ -1,0 +1,81 @@
+#!/bin/sh
+set -e
+
+# QEMU version and date string for easy maintenance
+QEMU_VERSION="9.2.2"
+QEMU_DATE="20250228"
+QEMU_RELEASE="esp-develop-${QEMU_VERSION}-${QEMU_DATE}"
+
+# 1. Detect host architecture
+ARCH=$(uname -m)
+case "$ARCH" in
+    x86_64)
+        QEMU_ARCH="x86_64"
+        QEMU_RISCV32_FILE="qemu-riscv32-softmmu-esp_develop_${QEMU_VERSION}_${QEMU_DATE}-x86_64-linux-gnu.tar.xz"
+        QEMU_XTENSA_FILE="qemu-xtensa-softmmu-esp_develop_${QEMU_VERSION}_${QEMU_DATE}-x86_64-linux-gnu.tar.xz"
+        ;;
+    aarch64 | arm64)
+        QEMU_ARCH="aarch64"
+        QEMU_RISCV32_FILE="qemu-riscv32-softmmu-esp_develop_${QEMU_VERSION}_${QEMU_DATE}-aarch64-linux-gnu.tar.xz"
+        QEMU_XTENSA_FILE="qemu-xtensa-softmmu-esp_develop_${QEMU_VERSION}_${QEMU_DATE}-aarch64-linux-gnu.tar.xz"
+        ;;
+    *)
+        echo "Unsupported architecture: $ARCH"
+        exit 1
+        ;;
+esac
+
+# Install some dependencies
+apt-get update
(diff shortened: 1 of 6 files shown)


.github/workflows/build_and_run_apps.yml
```diff
@@ -135,16 +135,27 @@ jobs:
...
+            pytest_args: ""
           - runs-on: "spi_nand_flash"
             marker: "spi_nand_flash"
             target: "esp32"
+            runs_on: [self-hosted, linux, docker, "spi_nand_flash"]
+            pytest_args: ""
+          - runs-on: "qemu"
+            marker: "qemu"
+            target: "esp32s3"
+            runs_on: [self-hosted, linux, docker]
```
tore-espressif: I agree with the co-pilot review that we should try to avoid mixing `runs-on` and `runs_on` (hyphen vs underscore)

.github/setup_qemu.sh
```diff
@@ -0,0 +1,76 @@
...
+        ;;
+    *)
+        echo "Unsupported architecture: $ARCH"
+        exit 1
+        ;;
+esac
+
+# Install some dependencies
+apt update
+apt install -y libgcrypt20 libglib2.0-0 libpixman-1-0 libsdl2-2.0-0 libslirp0
```
tore-espressif: Nitpick: using `apt-get` is preferable over `apt` in scripts

tore-espressif (approved):
@Ashish285 Only nitpick, LGTM!

.github/setup_qemu.sh
Ashish285: Sure, updated to use `apt-get`

.github/workflows/build_and_run_apps.yml
Ashish285: Agree, updated the `runs_on` to be `runner-labels`

.github/setup_qemu.sh
```diff
@@ -0,0 +1,76 @@
...
+esac
+
+# Install some dependencies
+apt-get update
+apt-get install -y libgcrypt20 libglib2.0-0 libpixman-1-0 libsdl2-2.0-0 libslirp0
+
+QEMU_DIR="qemu"
+
+# 2. Download the correct binary
+QEMU_RISCV32_URL="https://github.com/espressif/qemu/releases/download/esp-develop-9.2.2-20250228/$QEMU_RISCV32_FILE"
```
mahavirj: I suggest that we move the release string to some variable at the top: `9.2.2_20250228`. It will be easy to upgrade to latest release in future.

.github/setup_qemu.sh
```diff
@@ -0,0 +1,76 @@
...
+if [ -f "$QEMU_DIR/bin/qemu-system-riscv32" ]; then
+    echo "QEMU RISCV32 installation successful."
+else
+    echo "QEMU RISCV32 installation failed."
+    exit 1
+fi
+# Rename qemu directory to avoid conflicts
+mv "$QEMU_DIR" "${QEMU_DIR}_riscv32"
+
+QEMU_XTENSA_URL="https://github.com/espressif/qemu/releases/download/esp-develop-9.2.2-20250228/$QEMU_XTENSA_FILE"
```
mahavirj: nitpick: another option is that we use latest release URL `releases/latest`  works for GitHub release case

mahavirj:
LGTM!

.github/setup_qemu.sh
Ashish285: Moved to the top as suggested.

.github/setup_qemu.sh
Ashish285: The download file has the same version as the release, so I updated it to get it from one source itself.

PR #516, opened 2025-06-25
feat(esp_encrypted_img): Adds public key export APIs

Ashish285:
## Overview

This pull request adds new public key export APIs to the [esp_encrypted_img](https://github.com/espressif/idf-extra-components/tree/master/esp_encrypted_img) component. These APIs enable programmatic retrieval of public keys for supported cryptographic schemes (RSA-3072 and ECIES-P256) facilitating integration with external tools and workflows that require public key material.

## Reasoning
The primary motivation for this change is to support the "per device unique key" use case. In scenarios where each device is provisioned with a unique key pair, it is essential to be able to export the public key from the device. This allows secure image encryption tailored to each device, improving security and enabling scalable manufacturing and provisioning workflows.


esp_encrypted_img/include/esp_encrypted_img.h
```diff
@@ -181,6 +181,17 @@ esp_err_t esp_encrypted_img_decrypt_abort(esp_decrypt_handle_t ctx);
...
+/**
+ * @brief  Get the public key corresponding to the private key.
+ *         For RSA, the public key is in DER format and corresponds to the private key passed with `esp_encrypted_img_decrypt_start()`.
+ *         For ECIES, the public key is in uncompressed format and is derived from the HMAC key ID passed with `esp_encrypted_img_decrypt_start()`.
+ * 
+ * @param ctx   esp_decrypt_handle_t handle
+ * @param pub_key_len   Pointer to store the length of the public key
+ * @return uint8_t*   Pointer to the public key
+ */
+uint8_t *esp_encrypted_img_get_public_key(esp_decrypt_handle_t ctx, size_t *pub_key_len);
```
mahavirj: - It will be application layer responsibility to free the pointer after usage? If yes, please document.
- Suggest changing API to return `esp_err_t` and then signature to below
- Also suggest using export instead of get in the API name

```
esp_err_t esp_encrypted_img_export_public_key(esp_decrypt_handle_t handle, uint8_t **pub_key, size_t *pub_key_len)
```

esp_encrypted_img/README.md
```diff
@@ -260,6 +258,49 @@ Make sure these configurations are set correctly in your project's configuration
 
     The public key of the ephemeral server key pair used for ECDH is embedded in the image but not saved to a separate file.
 
+## Per Device Unique Key Support
+
+The `esp_encrypted_img` component supports workflows where each device is provisioned with a unique key pair. This is essential for scenarios requiring per-device image encryption and secure provisioning. By exporting the public key from the device or key management system, you can automate image generation and ensure that only the intended device can decrypt its firmware or data.
+
+To generate a per device unique key, use the following approach:
```
mahavirj: Usage of bootloader_random APIs is recommended only if the RF is not yet initialized (e.g., early startup code or factory app not using RF libraries). If RF is initialized then direct usage of `esp_fill_random` is recommended.

We can also link the IDF RNG documentation pointer here.

mahavirj:
@AdityaHPatwardhan PTAL

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -393,8 +487,101 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
+    }
+    *pub_key_len = 0;
+    return err;
+}
+
 #endif /* CONFIG_PRE_ENCRYPTED_OTA_USE_ECIES */
 
+esp_err_t esp_encrypted_img_export_public_key(esp_decrypt_handle_t ctx, uint8_t **pub_key, size_t *pub_key_len)
+{
+    if (ctx == NULL) {
```
mahavirj, line 591: Suggest checking for `pub_key` and `pub_key_len` pointer validity here

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -144,6 +146,98 @@ static int decipher_gcm_key(const char *enc_gcm, esp_encrypted_img_t *handle)
...
+    ret = mbedtls_pk_write_pubkey(&c, *pub_key, &pk);
+    if (ret < 0) {
+        ESP_LOGE(TAG, "Failed to write public key: -0x%04x", (unsigned int) - ret);
+        free(*pub_key);
+        *pub_key = NULL;
+        goto exit;
+    }
+    if (c - *pub_key < 1) {
+        ESP_LOGE(TAG, "Public key buffer is too small");
+        free(*pub_key);
```
mahavirj: Seems a repetitive failure code - might as well handle inside the error label

esp_encrypted_img/README.md
```diff
@@ -260,6 +258,42 @@ Make sure these configurations are set correctly in your project's configuration
...
+### Exporting the Public Key Programmatically
+
+The component provides an API to retrieve the public key associated with the decryption context:
+
+```c
+/**
+ * @brief  Export the public key corresponding to the private key.
+ *         The application should free the memory pointed by `pub_key` after use.
+ *         For RSA, the public key is in DER format and corresponds to the private key passed with `esp_encrypted_img_decrypt_start()`.
+ *         For ECIES, the public key is in uncompressed format and is derived from the HMAC key ID passed with `esp_encrypted_img_decrypt_start()`.
```
mahavirj: Isn't this standard DER format, would it require some additional processing for the conversion?

Ashish285: It was raw format instead of DER. I updated it to also return back with DER formatted key

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -393,8 +485,145 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    esp_encrypted_img_t *handle = (esp_encrypted_img_t *)ctx;
+
+#if defined(CONFIG_PRE_ENCRYPTED_OTA_USE_RSA)
+    // In case of RSA, we return the public key corresponding to the private key
+    // passed with esp_encrypted_img_decrypt_start()
+    esp_err_t err = esp_encrypted_img_export_rsa_pub_key(handle->rsa_pem, handle->rsa_len, pub_key, pub_key_len);
+    return err;
```
Harshal5: nitpick:
```suggestion
    return esp_encrypted_img_export_rsa_pub_key(handle->rsa_pem, handle->rsa_len, pub_key, pub_key_len);
```

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -393,8 +485,145 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
+#elif defined(CONFIG_PRE_ENCRYPTED_OTA_USE_ECIES)
+    // In case of ECIES, we return the public key corresponding to the private key
+    // derived from the HMAC key ID passed with esp_encrypted_img_decrypt_start()
+    esp_err_t err = esp_encrypted_img_export_ecies_pub_key(handle->hmac_key, pub_key, pub_key_len);
+    return err;
```
Harshal5:
```suggestion
    return esp_encrypted_img_export_ecies_pub_key(handle->hmac_key, pub_key, pub_key_len);
```

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -144,6 +147,95 @@ static int decipher_gcm_key(const char *enc_gcm, esp_encrypted_img_t *handle)
...
+    }
+
+    unsigned char *c = *pub_key + max_pub_key_size;
+
+    ret = mbedtls_pk_write_pubkey(&c, *pub_key, &pk);
+    if (ret < 0) {
+        ESP_LOGE(TAG, "Failed to write public key: -0x%04x", (unsigned int) - ret);
+        goto exit;
+    }
+    if (c - *pub_key < 1) {
```
Harshal5: Should the check be < 0 instead of < 1?

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -144,6 +147,95 @@ static int decipher_gcm_key(const char *enc_gcm, esp_encrypted_img_t *handle)
...
+        ESP_LOGE(TAG, "Public key buffer is too small");
+        goto exit;
+    }
+    // Adjust the length of the public key
+    *pub_key_len = max_pub_key_size - (c - *pub_key);
+    if (*pub_key_len > max_pub_key_size) {
+        ESP_LOGE(TAG, "Public key length exceeds allocated buffer size");
+        goto exit;
+    }
```
Harshal5: I think this would be redundant in our case, because we already have the `c - *pub_key < ` check above, so the condition to satisfy `*pub_key_len > max_pub_key_size` should not be possible.

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -393,8 +485,145 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
     return ret;
 }
 
+static esp_err_t esp_encrypted_img_export_ecies_pub_key(hmac_key_id_t hmac_key, uint8_t **pub_key, size_t *pub_key_len)
+{
+    esp_err_t err = ESP_OK;
```
Harshal5: nitpick:
```suggestion
    esp_err_t err = ESP_FAIL;
```
And then we could remove the `err = ESP_FAIL;` assignments in the multiple failure conditions below.

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -393,8 +485,145 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
+    // Set the private key
+    ret = mbedtls_mpi_copy(&ecp_keypair->MBEDTLS_PRIVATE(d), &ecc_priv_key);
+    if (ret != 0) {
+        ESP_LOGE(TAG, "Failed to copy private key: -0x%04x", (unsigned int) - ret);
+        err = ESP_FAIL;
+        goto exit;
+    }
+
+    // Compute the public key from private key
+    ret = mbedtls_ecp_mul(&ecp_keypair->MBEDTLS_PRIVATE(grp), &ecp_keypair->MBEDTLS_PRIVATE(Q), &ecp_keypair->MBEDTLS_PRIVATE(d), &ecp_keypair->MBEDTLS_PRIVATE(grp).G, mbedtls_esp_random, NULL);
```
Harshal5: Could we use `mbedtls_ecp_keypair_calc_public()` instead?

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -393,8 +485,145 @@ static int derive_gcm_key(const char *data, esp_encrypted_img_t *handle)
...
+    }
+
+    ret = mbedtls_pk_write_pubkey_der(&pk_ctx, *pub_key, max_pubkey_len);
+    if (ret < 0) {
+        ESP_LOGE(TAG, "Failed to write public key DER: -0x%04x", (unsigned int) - ret);
+        err = ESP_FAIL;
+        goto exit;
+    }
+
+    *pub_key_len = ret;
```
Harshal5, line 552: Why do we not need the public key length adjustment similar to the way done in case of RSA?

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: Yes

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: Actually for RSA use case as well we can use the same thing, we do not need to do the pointer arithmetic to get the length of the public key

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: Yes, resolved this.

Harshal5 (approved):

esp_encrypted_img/src/esp_encrypted_img.c
```diff
@@ -144,6 +147,91 @@ static int decipher_gcm_key(const char *enc_gcm, esp_encrypted_img_t *handle)
...
+    }
+
+    unsigned char *c = *pub_key + max_pub_key_size;
+
+    ret = mbedtls_pk_write_pubkey(&c, *pub_key, &pk);
+    if (ret < 0) {
+        ESP_LOGE(TAG, "Failed to write public key: -0x%04x", (unsigned int) - ret);
+        goto exit;
+    }
+    if (c - *pub_key < 0) {
```
AdityaHPatwardhan, line 201: If the public key was successfully written above, why would the public key buffer too small here?

esp_encrypted_img/README.md
```diff
@@ -216,11 +216,9 @@ When using an HMAC-derived device key for ECIES-P256, the device hmac key must b
 
     Replace `BLOCK_KEY<N>` with the actual eFuse block you intend to use (e.g., `BLOCK_KEY0`, `BLOCK_KEY1`, etc., up to `BLOCK_KEY5` typically).
 
-2. **Kconfig Configuration for eFuse Block:**
+2. **Code Configuration:**
 
-    The eFuse block chosen in the command above must match the configuration in your project. In the `pre_encrypted_ota` example, this is set using the Kconfig option:
-
-    * `PRE_ENC_OTA_HMAC_EFUSE_KEY_ID`: This integer value (e.g., 0 for `BLOCK_KEY0`, 1 for `BLOCK_KEY1`, etc.) specifies which eFuse key block holds the HMAC key. Ensure this matches the `<N>` used in the `efuse-burn-key` command.
+    The eFuse block chosen in the command above must match the HMAC key ID used in your application code. In the `pre_encrypted_ota` example, the HMAC key ID is defined in the source code and should be updated to match your chosen eFuse block.
```
AdityaHPatwardhan, line 221: can we also mention the variable or configuration option which needs to be set here for the HMAC key id?

AdityaHPatwardhan (approved):
Small nitpicks. Otherwise looks good to me.
Thanks @Ashish285 For improving the esp_encrypted_img feature.

esp_encrypted_img/src/esp_encrypted_img.c
Ashish285: This is just a sanity check to ensure that the write doesn't go past the allocated buffer.

esp_encrypted_img/README.md
Ashish285: Sure, updated the doc

mahavirj:
@Ashish285 

- Please rebase the branch to latest master
- Lets increase the minor version of the component - new feature addition
- Add ChangeLog entry

mahavirj (approved):

PR #518, opened 2025-06-26
Build and deploy API reference && programming guide

suda-morris:
### This is a proposal to build and deploy each component's programming guide with its API reference.

All components need documentation, they should be hosted somewhere. However, ESP Registry currently can't provide a complete documentation host service. This PR proposed a simple and extensible solution, which works well for a **monorepo** likes `idf-extra-components`. Please note, the components inside `idf-extra-components` repo are independent of each other. Thus:
1. The docs (markdown files) are living inside each component (self-contained)
2. Each component can have its own doc extensions and configurations (i.e. each component has a `book.toml`)
3. Doxygen configuration file is also per-component
4. The URL is also unique to each component. There isn't a home page that contains the index of all components, because components are independent.

### Why I choose mdbook as the backend

1. mdbook is written in Rust, and supported by the Rust team. It runs very fast and stable, which is good for CI.
2. the default settings are sufficient for most of the use cases, there's no complicated configurations, only several lines in the book.toml
3. The default theme is clean but professional for API guides

### The flow of the CI pipeline to build and deploy the docs

You can see the flows in the `deploy_gh_pages.yml`. Also the real action: https://github.com/espressif/idf-extra-components/actions/runs/15896958746/job/44830741792?pr=518

1. Search the components in the repo, find the `docs` folder, the `docs` should contain a `book.toml` file (key feature of a mdbook project)
2. Call esp-doxybook to generate a `api.md`
3. Call mdbook to build the docs, the artifacts will be saved in the `my_component/docs/book` folder
4. Copy the contents from  `my_component/docs/book`  to `repo_root/docs_build_output/my_component`
5. Use lychee to check the links (including anchors, because we use markdown anchors to refer to a function) for the `docs_build_output` folder
6. Give the `docs_build_output` a version, currently only `latest` is supported. 
7. Use Github Action to deploy the site.

### Examples

* [led strip](https://suda-morris.github.io/idf-extra-components/latest/led_strip/)
* [ESP Serial Slave Link](https://suda-morris.github.io/idf-extra-components/latest/esp_serial_slave_link/)


.github/build_docs.py
```diff
@@ -0,0 +1,303 @@
...
+    try:
+        # Generate API documentation
+        generate_api_docs(component_docs_path)
+
+        # Build mdbook
+        env = os.environ.copy()
+
+        # Set site-url based on whether this is a PR or merge event
+        if config.pr_number:
+            site_url = f"/idf-extra-components/pr-preview-{config.pr_number}/{component_name}/"
```
igrr, line 136: If I understand it right, you are running the docs build for PRs as well.

When the docs are built for PR, do we somehow preserve the content of the /latest/ directory?

igrr: And if there are multiple PRs in progress, do they not overwrite each others' versions of the docs?

suda-morris: Hi @igrr yes, we only run the docs build for PR at the moment. But there's no deploy for the PR build. And there's also no docs preview. (left for future when we think it's necessary). The "Prepare files for deployment 📁" and "Deploy to Github Pages 🚀" only available for a "Push to Master" event.

igrr (approved):

PR #526, opened 2025-07-04
Feat/add linenoise library

SoucheSouche:
# Checklist

- [ ] CI passing

# Change description
Add linenoise to the list of extra components. The library is used in the console component of esp-idf that is going through refactoring. One part of the refactoring is to move third party libraries outside of the console component. Eventually, the console component itself will be removed from esp-idf, reason why this PR adds the linenoise lib to the idf-extra-components.


suda-morris:
There was a PR about the linenoise library: https://github.com/espressif/esp-idf/pull/10526
WDYT? @SoucheSouche

SoucheSouche:
> There was a PR about the linenoise library: [espressif/esp-idf#10526](https://github.com/espressif/esp-idf/pull/10526) WDYT? @SoucheSouche

Yes, I am aware of it. I took into account Ivan's comment left in the PR when implementing this version of linenoise. The author of the PR you mentioned is aware that I will take over and continue the work on improving the console component (linenoise and more generally the list of idea he listed here https://github.com/espressif/esp-idf/issues/9986)

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+    bool allow_multi_line;
+    bool allow_empty_line;
+    bool allow_dumb_mode;
+    linenoise_completion_callback completion_cb;
+    linenoise_hints_callback hints_cb;
+    linenoise_free_hints_callback free_hints_cb;
+    linenoise_read_bytes_fn read_bytes_fn;
+    linenoise_write_bytes_fn write_bytes_fn;
+    char **history;
+} linenoise_instance_param_t;
```
igrr: For consistency with other components, let's call this "config"

```suggestion
} linenoise_instance_config_t;
```

Or simply `linenoise_config_t`.

If you call the constructor function `linenoise_create` then it's going to be quite clear that it creates a new instance, without having to add the word `instance` everywhere. This probably applies to the handle as well.

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+ * @return Number of bytes written, or -1 on error.
+ */
+typedef ssize_t (*linenoise_write_bytes_fn)(int fd, const void* buf, size_t count);
+
+/**
+ * @brief Structure defining the parameters needed by a linenoise
+ * instance
+ */
+typedef struct linenoise_instance_param {
+    const char *prompt;
```
igrr: Please add Doxygen comments for struct members.

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+ * @param count Number of bytes to write.
+ * @return Number of bytes written, or -1 on error.
+ */
+typedef ssize_t (*linenoise_write_bytes_fn)(int fd, const void* buf, size_t count);
+
+/**
+ * @brief Structure defining the parameters needed by a linenoise
+ * instance
+ */
+typedef struct linenoise_instance_param {
```
igrr: Instance config structure should have a `void* user_ctx` member that will be passed to all the callbacks.

This design in common in libraries which support multiple instances, and allows, for example, for the caller to find the local "state" or "context" structure related to the instance which has triggered the particular callback. This is also useful when wrapping the library in a C++ class, the `user_ctx` is then often the pointer to `this` (class instance).

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+ * @brief Opaque handle to a linenoise instance.
+ */
+typedef struct linenoise_instance *linenoise_handle_t;
+
+/**
+ * @brief Returns the default parameters for creating a linenoise instance.
+ *
+ * @return A populated linenoise_instance_param_t structure with default values.
+ */
+linenoise_instance_param_t linenoise_get_instance_param_default(void);
```
igrr:
```suggestion
void linenoise_get_instance_param_default(linenoise_instance_param_t *out);
```

The caller may have placed the configuration structure in heap or .bss, allowing them to pass the pointer to it helps avoid storing another copy on the stack.

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * @brief Creates a new linenoise instance.
+ *
+ * @param param Pointer to the configuration parameters for the instance.
+ * @return Handle to the created instance, or NULL on failure.
+ */
+linenoise_handle_t linenoise_create_instance(linenoise_instance_param_t *param);
```
igrr:
```suggestion
esp_err_t linenoise_create_instance(const linenoise_instance_param_t *param, linenoise_handle_t* out_handle);
```

1. Configuration structure can be a const pointer
2. You may need to return different types of errors (out of RAM, invalid argument in the config structure, etc), which simply returning NULL doesn't allow. Hence, return esp_err_t and get the instance handle as an output argument.

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * @brief Destroys a linenoise instance and frees associated memory.
+ *
+ * @param handle Handle to the instance to delete.
+ * @return ESP_OK on success, or error code on failure.
+ */
+esp_err_t linenoise_delete_instance(linenoise_handle_t handle);
```
igrr: What should the caller do if the "delete instance" function has returned an error?

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+esp_err_t linenoise_clear_screen(linenoise_handle_t handle);
+
+/**
+ * @brief Sets whether an empty line is allowed to be returned.
+ *
+ * @param handle Handle to the linenoise instance.
+ * @param empty_line true to allow empty lines, false to disallow.
+ * @return ESP_OK on success, or error code on failure.
+ */
+esp_err_t linenoise_set_empty_line(linenoise_handle_t handle, bool empty_line);
```
igrr: Is there a use case for changing these settings (empty_line, multi_line, dumb_mode, max_cmd_line_length) once the instance is already created? If there is no concrete use case, it might be simpler to not implement these getters and setters.

linenoise/linenoise/linenoise.h
```diff
@@ -0,0 +1,335 @@
...
+ * @param max_cmd_line_length Pointer to receive the max length.
+ * @return ESP_OK on success, or error code on failure.
+ */
+esp_err_t linenoise_get_max_cmd_line_length(linenoise_handle_t handle, size_t *max_cmd_line_length);
+
+/*------------------------------------------------------------------------------------------------
+ *OLD LINENOISE API
+ *------------------------------------------------------------------------------------------------*/
+
+typedef struct linenoise_completions linenoiseCompletions;
```
igrr: I'd recommend having the compatibility layer in a different header file from the "new" API.

Components which use the old API, can include, for example, "linenoise/linenoise.h" (as they used to).

Components which use the new API can include a different file, such as "esp_linenoise.h", and only get the new functions, without the old functions polluting the global namespace.

linenoise/idf_component.yml
```diff
@@ -0,0 +1,9 @@
+version: "1.0.0"
```
igrr: Since the original upstream version of linenoise was tagged as 1.0, perhaps we can define this as 2.0.0?

linenoise/sbom_linenoise.yml
```diff
@@ -0,0 +1,3 @@
+name: sbom_linenoise
+description: Line editing C library
+url: https://github.com/antirez/linenoise
```
igrr: The library is not maintained upstream anymore, i think the URL should point to idf-extra-components instead.

Please also check the other fields:

```yml
name: linenoise
version: 2.0.0
cpe: cpe:2.3:a:espressif:linenoise:{}:*:*:*:*:*:*:*
supplier: 'Organization: Espressif Systems'
```

igrr (changes requested):

linenoise/idf_component.yml
suda-morris: or we call this component name "esp_linenoise" then version start from 1.0.0 sounds also good, in case the upstream linenoise library get a chance to upgrade to their 2.0

linenoise/linenoise/linenoise.h
SoucheSouche: It makes sense, I will update the naming conventions.

linenoise/idf_component.yml
igrr: Sounds reasonable. In that case, we can change the API prefix from `linenoise_` to `esp_linenoise`.

linenoise/linenoise/linenoise.h
SoucheSouche: The user might want to repurpose an existing linenoise instance, instead of deleting the instance and creating a new one. In that case it might be useful to keep the getters and setters

linenoise/linenoise/linenoise.h
SoucheSouche: Implemented.

linenoise/linenoise/linenoise.h
SoucheSouche: the only way the function returns an error is if the handle is invalid. I understand the question but then I would also expect that it is also valid for the comment above. So either I update the linenoise_create_instance to return error values and I leave this function as it is now or I update linenoise_delete_instance to abort when the handle is invalid and I keep the function linenoise_create_insance as it is now.

Because I am not sure what the caller can do if linenoise_create_instance returns e.g., no memory error.

linenoise/linenoise/linenoise.h
SoucheSouche: topology updated. I added the new linenoise to esp_linenoise.c/h files

SoucheSouche:
@igrr, I updated the component based on your input. I left a few comments open for discussion.

I added a bunch of tests for the basic functionality of linenoise when escape sequences are supported by the terminal.

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
igrr, line 5: Please convert the above license block to SPDX format. And you can relicense the final library as Apache 2.0.

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+
+#pragma once
+
+#include <sys/types.h>
+#include <unistd.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include "esp_err.h"
+
```
igrr, line 19: Missing C++ guards

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+/**
+ * @brief Function pointer type for writing bytes.
+ *
+ * @param fd File descriptor.
+ * @param buf Buffer containing bytes to write.
+ * @param count Number of bytes to write.
+ * @param user_ctx User-defined context pointer passed from the configuration.
+ * @return Number of bytes written, or -1 on error.
+ */
+typedef ssize_t (*esp_linenoise_write_bytes_fn)(int fd, const void *buf, size_t count, void *user_ctx);
```
igrr: By convention, user_ctx should be the first argument.

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+    int out_fd; /*!< File descriptor to write output to (e.g., STDOUT_FILENO) */
+    bool allow_multi_line; /*!< Whether to allow multi-line input (true to enable) */
+    bool allow_empty_line; /*!< Whether to allow accepting an empty line as valid input */
+    bool allow_dumb_mode; /*!< Whether to allow running in dumb terminal mode (without advanced features) */
+    esp_linenoise_completion_callback completion_cb; /*!< Callback function for handling input completions */
+    esp_linenoise_hints_callback hints_cb; /*!< Callback function to provide input hints (e.g., usage suggestions) */
+    esp_linenoise_free_hints_callback free_hints_cb; /*!< Callback function to free hints returned by `hints_cb` */
+    void *user_ctx; /*!< User context information passed to the callbacks */
+    esp_linenoise_read_bytes_fn read_bytes_fn; /*!< Function used to read bytes from the input stream */
+    esp_linenoise_write_bytes_fn write_bytes_fn; /*!< Function used to write bytes to the output stream */
```
igrr: Why are some callback names end with `_callback` and others with `_fn`? Can we make them consistent?

I would also recommend keeping `_t` suffix for function pointer typedefs.

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+ */
+void esp_linenoise_get_instance_config_default(esp_linenoise_config_t *config);
+
+/**
+ * @brief Creates a new linenoise instance.
+ *
+ * @param param Pointer to the configuration parameters for the instance.
+ * @return Handle to the created instance, or NULL on failure.
+ */
+esp_linenoise_handle_t esp_linenoise_create_instance(esp_linenoise_config_t *param);
```
igrr:
```suggestion
esp_linenoise_handle_t esp_linenoise_create_instance(const esp_linenoise_config_t *param);
```

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+
+/**
+ * @brief Opaque handle to a linenoise instance.
+ */
+typedef struct esp_linenoise_instance *esp_linenoise_handle_t;
+
+/**
+ * @brief Returns the default parameters for creating a linenoise instance.
+ *
+ * @param config esp_linenoise_config_t structure to populate with default values.
```
igrr:
```suggestion
 * @param[out] config esp_linenoise_config_t structure to populate with default values.
```

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,303 @@
...
+ */
+void esp_linenoise_add_completion(esp_linenoise_completions_t *lc, const char *str);
+
+/**
+ * @brief Reads a line of input from the user into a buffer.
+ *
+ * @param handle Handle to the linenoise instance.
+ * @param cmd_line_buffer Buffer to store the input line.
+ * @param cmd_line_length Length of the command line buffer.
+ * @return ESP_OK on success, or error code on failure.
```
igrr: It would be nice to clarify what happens if cmd_line_buffer is not large enough to fit a line of input. Do we return ESP_OK or an error, is the line truncated, etc.

SoucheSouche: Firstly, I added an input validation check on cmd_line_length so that if it exceeds max_cmd_line_length, the function returns with invalid arg error. esp_linenoise_history_add creates a buffer of size max_cmd_line_length so we have to enforce a restriction on cmd_line_length. I added a test to check the return values of `esp_linenoise_get_line`.

Secondly, the current behavior of `esp_linenoise_edit` is to fill the buffer until `cmd_line_length - 1` (last char used for '\0') and skip the new characters until new line is received, then returns. I added a description in the doxygen of the function. I also updated `esp_linenoise_dumb` to behave like `esp_linenoise_edit`. I extended the behavioral tests to make sure this behavior is respected for both dumb mode on and off.

esp_linenoise/linenoise/linenoise.c
```diff
@@ -0,0 +1,413 @@
...
+#include "esp_linenoise_private.h"
+#include "esp_linenoise.h"
+#include "linenoise.h"
+#include "esp_err.h"
+
+static esp_linenoise_instance_t *s_linenoise_instance = NULL;
+static linenoiseCompletionCallback *s_completion_cb = NULL;
+static linenoiseHintsCallback *s_hints_cb = NULL;
+static linenoiseFreeHintsCallback *s_free_hints_cb = NULL;
+static linenoiseReadBytesFn s_read_cb = NULL;
```
ESP-Marius: If we want this linenoise to work as it did before, do we need to default to `read()` here?

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * @brief Function pointer type for reading bytes.
+ *
+ * @param user_ctx User-defined context pointer passed from the configuration.
+ * @param fd File descriptor.
+ * @param buf Buffer to store the read bytes.
+ * @param count Number of bytes to read.
+ * @return Number of bytes read, or -1 on error.
+ */
+typedef ssize_t (*esp_linenoise_read_bytes_t)(void *user_ctx, int fd, void *buf, size_t count);
```
ESP-Marius: Just thinking back to the "how to deinit the console/cancel a read" issue. 

On which level do we plan to handle it now? Do we here expect users to use a read function that they have a way to "cancel". Who should do the cancellation? E.g. will it our repl implementation do it? If so what kind of interface do we imagine here?

esp_linenoise/idf_component.yml
```diff
@@ -0,0 +1,10 @@
+version: "1.0.0"
+description: "ESP Linenoise - Line editing C library"
+url: https://github.com/espressif/idf-extra-components/tree/master/esp_linenoise
+license: Apache-2.0
+dependencies:
+  idf: ">=6.0"
```
ESP-Marius: This library is not very coupled to anything in IDF, so it sounds like it should be able to be used with any IDF version?

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,276 @@
...
+
+#include <sys/types.h>
+#include <unistd.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include "esp_err.h"
+
+/*
+ * IMPORTANT: This library is not thread safe.
```
ESP-Marius, line 21: not thread-safe in what way? multiple threads operating on the same instance?

Or is it not safe even if thread a works on instance A and thread B on instance B? (I assume it is?)

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,276 @@
...
+
+/**
+ * @brief Structure defining the parameters needed by a linenoise
+ * instance
+ */
+typedef struct esp_linenoise_config {
+    const char *prompt; /*!< Prompt string displayed to the user */
+    size_t max_cmd_line_length; /*!< Maximum length (in bytes) of the input command line */
+    int history_max_length; /*!< Maximum number of entries to store in command history */
+    int in_fd; /*!< File descriptor to read input from (e.g., STDIN_FILENO) */
```
ESP-Marius, line 111: If users are going to provide their own filedescriptors, do we need to document what kind of operations we expect them to be able to do? I think I saw that it needs to be able to do both blocking and nonblocking right?

esp_linenoise/host_test/pytest_linenoise.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+@pytest.mark.generic
```
ESP-Marius: Should we test on linux as well?

esp_linenoise/test_apps/main/test_main.c
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void tearDown(void)
+{
+    unity_utils_evaluate_leaks_direct(0);
+}
+
+void app_main(void)
+{
+    printf("Running linenoise component tests\n");
```
ESP-Marius, line 24: maybe I missed it, but i didnt see any tests that acutally tried to use multiple instances with sepearte FDs .Should we add some?

esp_linenoise/include/esp_linenoise.h
```diff
@@ -0,0 +1,276 @@
...
+    int out_fd; /*!< File descriptor to write output to (e.g., STDOUT_FILENO) */
+    bool allow_multi_line; /*!< Whether to allow multi-line input (true to enable) */
+    bool allow_empty_line; /*!< Whether to allow accepting an empty line as valid input */
+    bool allow_dumb_mode; /*!< Whether to allow running in dumb terminal mode (without advanced features) */
+    esp_linenoise_completion_t completion_cb; /*!< Callback function for handling input completions */
+    esp_linenoise_hints_t hints_cb; /*!< Callback function to provide input hints (e.g., usage suggestions) */
+    esp_linenoise_free_hints_t free_hints_cb; /*!< Callback function to free hints returned by `hints_cb` */
+    esp_linenoise_read_bytes_t read_bytes_cb; /*!< Function used to read bytes from the input stream */
```
ESP-Marius, line 119: It feels a bit strange that we are mixing filedescriptors and custom write/reads.

Would it make more sense to just stick to one? Either we pass a `api` struct, and the `fd` is just any backend context. Or we stick to posix?

esp_linenoise/include/esp_linenoise.h
SoucheSouche: I added a few line of comments to better explain.

esp_linenoise/linenoise/linenoise.c
SoucheSouche: fixed.

esp_linenoise/include/esp_linenoise.h
SoucheSouche: I think it makes sense to keep both the custom read / write and the FD in the config structure. The user can choose to provide the FD but no custom read / write in which case, the default posix read / write function will be called with the FD provided by the user. Or the user can provide custom read / write without specifying the FD in which case its custom read / write will be called with default FD (FILENO_STDIN, FILENO_STDOUT). Or the user can use the default posix read / write with the default FD (FILENO_STDIN, FILENO_STDOUT). I think there is a use case for each configuration so I would suggests keeping the config structure as it is.

esp_linenoise/host_test/pytest_linenoise.py
SoucheSouche: The tests are supposed to run on linux only. I added the target 'linux' to the test

esp_linenoise/test_apps/main/test_main.c
SoucheSouche: test added

esp_linenoise/include/esp_linenoise.h
SoucheSouche: I updated the default read function provided by linenoise when the user doesn't provide its own. Now it uses select / read. I also added an abort function that will force select to return and return from the default read function with a buffer containing a new line character, forcing esp_linenoise_get_line to return as if the user entered the new line itself.

I made this feature only available if the user uses esp_linenoise_abort (the way we handled it in the esp-idf console component) so that if the abort is never used, the default read just calls blocking posix read instead and we don't have to include the binary overhead of select and eventfd.

esp_linenoise/include/esp_linenoise.h
SoucheSouche: I think I will do it as follow:
- if the user doesn't provide a custom read, then linenoise will set the read mode to blocking internally. If the user provides its own custom read function, then it will fall under the user responsibility to set the proper read mode,

what do you think ?

SoucheSouche: implemented as explained.

SoucheSouche:
@igrr, could you please take another look, and check the comments I left opened?

PR #527, opened 2025-07-09
Ci/add precommit

Honza0297:
# Checklist

- [ ] Component contains License - I don't add component
- [ ] Component contains README.md - I don't add component
- [ ] Component contains idf_component.yml file with `url` field defined - I don't add component
- [ ] Component was added to [upload job](https://github.com/espressif/idf-extra- components/blob/master/.github/workflows/upload_component.yml#L18) - I don't add component
- [ ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8) - I don't add component
- [ ] _Optional:_ Component contains unit tests - I don't add component
- [ ] CI passing

# Description

I found out some Kconfig and sdkconfig files have issues. I fixed them + added kconfcheck pre-commit hooks to prevent such issues in the future. 

Note to sdkconfig.defaults change: ESP-IDF 4.x is unsupported, no need to have deprecated option there.


coremark/examples/coremark_example/sdkconfig.defaults
```diff
@@ -1,4 +1 @@
-# For IDF v5.x
 CONFIG_ESP_TASK_WDT_INIT=n
-# For IDF v4.x
-CONFIG_ESP_TASK_WDT=n
```
igrr, line 4: However this component does support IDF v4.4:

https://github.com/espressif/idf-extra-components/blob/03d9d6ca4dc071d5b6bbee30265c1a78c6da2389/coremark/idf_component.yml#L7-L8

Honza0297: Ah, true. Thanks for pointing that out!
~~I also removed the hook checking deprecated options as it is not taking the minimal supported IDF version into account.~~
EDIT: After IRL discussion, I bumped up coremark major version and dropped IDF v4.4 there, the rest got only bugfix version bump and no dependency changes.

coremark/idf_component.yml
```diff
@@ -1,8 +1,8 @@
-version: "1.1.0~1"
+version: "2.0.0"
```
igrr: @Honza0297 Could you please keep "1.1.0~2" here?

1.1 is the upstream version, so we try to keep the component version aligned.

Changing the minimal supported IDF version is not a breaking change requiring a major version bump in this case, since users who have have IDF v4.4 and have specified the dependency as `~1.1.0` will get `1.1.0~1` automatically due to the constraint on the IDF version. Users of newer IDF versions will get `1.1.0~2` automatically.

esp_jpeg/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "1.3.0"
+version: "1.3.1"
```
igrr, line 1: Could you please add an entry in https://github.com/espressif/idf-extra-components/blob/55c79124f1973c565cae1683a4274ed87dc8ac6a/esp_jpeg/CHANGELOG.md?

esp_encrypted_img/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "2.5.0"
+version: "2.5.1"
```
igrr, line 1: Likewise, please update https://github.com/espressif/idf-extra-components/blob/55c79124f1973c565cae1683a4274ed87dc8ac6a/esp_encrypted_img/CHANGELOG.md

(sorry that I forgot to mention this earlier)

igrr (approved):

PR #529, opened 2025-07-17
feat: Add a component for parsing external partition tables (IEC-333)

adokitkat:
# Checklist

- [x] CI passing

# Description

This MR adds a component which can parse and generate external partition tables. Right now it only supports MBR. Adding support for another partition table like GPT is possible but not planned for now.

```c
esp_err_t esp_mbr_parse(void* mbr_buf, esp_ext_part_list_t* part_list, esp_mbr_parse_extra_args_t* extra_args);

esp_err_t esp_mbr_generate(mbr_t* mbr, esp_ext_part_list_t* part_list, esp_mbr_generate_extra_args_t* extra_args);
```

Parsing a partition table outputs `esp_ext_part_list_t` structure filled with `esp_ext_part_list_item_t` items.

```c
typedef struct {
    uint32_t data[1];
    esp_ext_part_signature_type_t type;
} esp_ext_part_list_signature_t;

typedef struct {
    uint64_t address; /*!< Start address in bytes */
    uint64_t size; /*!< Size in bytes */
    uint64_t extra; /*!< Extra information (e.g. LittleFS block size stored in CHS hack, etc.) */
    char *label;
    esp_ext_part_flags_t flags; /*!< Flags for the partition */
    uint8_t type; /*!< Known partition type for this component (usually a part of `esp_ext_part_type_known_t`) */
} esp_ext_part_t;

typedef struct esp_ext_part_list_item_ {
    esp_ext_part_t info;
    SLIST_ENTRY(esp_ext_part_list_item_) next;
} esp_ext_part_list_item_t;

typedef struct {
    esp_ext_part_list_signature_t signature; /*!< Disk signature or identifier */
    SLIST_HEAD(esp_ext_part_list_head_, esp_ext_part_list_item_) head; /*!< Head of the partition list */
    esp_ext_part_list_flags_t flags; /*!< Flags for the partition list */
    esp_ext_part_sector_size_t sector_size; /*!< Sector size (storage medium property) */
} esp_ext_part_list_t;
```

When generating MBR, user can decide on sector size (512B, 4KiB) and partition alignment (4KiB, 1MiB). The defaults are 512B sector size and 1MiB partition alignment.

Note: Generating MBR doesn't mean formatting any partition on given sector span, however helper function could be created which would combine this behavior - each partition type has it's own formatting function (FATFS formats differently than LittleFS, etc.).

These are recognized types:

```c
typedef enum {
    ESP_EXT_PART_TYPE_UNKNOWN = 0,
    ESP_EXT_PART_TYPE_FAT12,
    ESP_EXT_PART_TYPE_FAT16, /*!< FAT16 with LBA addressing */
    ESP_EXT_PART_TYPE_FAT32, /*!< FAT32 with LBA addressing */
    ESP_EXT_PART_TYPE_LITTLEFS, /*!< Possibly LittleFS (MBR CHS field => LittleFS block size hack) */
// Note: The following types are not supported, but we can return a type for them
    ESP_EXT_PART_TYPE_LINUX_ANY, /*!< Linux partition (any type) */
    ESP_EXT_PART_TYPE_EXFAT_OR_NTFS, /*!< Not supported, but we can return a type for it */
    ESP_EXT_PART_TYPE_GPT_PROTECTIVE_MBR, /*!< Not supported, but we can return a type for it */
} esp_ext_part_type_known_t;
```

Future TODOs and could haves:
- FATFS and this component is right now not coupled - it would be nice to add a function to FATFS or VFS, which could return FAT type of a partition, so it ca be used in this component. 
- Add support for BDL and generic BDL partition components, also probably PR to esp_littlefs repo
- Support formatting along with partitioning? 
- Maybe add something like `ESP_EXT_PART_TYPE_FAT_ANY` and a function which inspects a FAT partition and decides on a correct type? This would require BDL.


esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,347 @@
...
+    for (int i = 0; i < 4; i++) {
+        partition = (mbr_partition_t*) &mbr->partition_table[i];
+
+        // Check if the partition entry is empty and if so, skip it
+        bool is_empty = true;
+        for (uint8_t i = 0; i < sizeof(mbr_partition_t); i++) {
+            if (((uint8_t *)partition)[i] != 0) {
+                is_empty = false;
+            }
+        }
```
adokitkat: Maybe this would be enough?

```c
if (partition->type == 0x00) {
    is_empty = true;
}
```

What if the partition was there anyway? Some file-systems do not have defined MBR type code. This is also normaly true for LittleFS, however there I reused 0xC3 type code after a discussion with LittleFS maintainers.

esp_ext_part_tables/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
adokitkat, line 1: Not sure if this is the correct license.

adokitkat:
@igrr @pacucha42 please take a look

esp_ext_part_tables/LICENSE
pacucha42: Seems 1:1 with https://www.apache.org/licenses/LICENSE-2.0, so it's the correct one

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,383 @@
...
+{
+    uint32_t lba_start = partition->lba_start;
+    uint32_t lba_end = lba_start - 1 + partition->sector_count;
+
+    switch (item->info.type) {
+    case ESP_EXT_PART_TYPE_FAT12:
+    case ESP_EXT_PART_TYPE_FAT16:
+    case ESP_EXT_PART_TYPE_FAT32:
+        // Set CHS values based on LBA start and end
+        esp_mbr_lba_to_chs_arr(partition->chs_start, lba_start);
```
pacucha42, line 129: Please, provide an explanation why you fill the CHS fields in for LBA file-system types (as discussed 1:1, this is given mostly by the FatFS support of FDISK functionality). Tbh, it would be great to know the reasons behind, otherwise this operation looks useless

esp_ext_part_tables/src/esp_mbr.c
pacucha42: 0x00 says there is no partition defined for given disk space and the space is considered free. I don't see a reason to go behind this model. You report all unprocessable partitions as unknown, so the risk of damaging existing (unknown) data is zero. Based on https://en.wikipedia.org/wiki/Partition_type

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,383 @@
...
+                           esp_mbr_generate_extra_args_t* extra_args)
+{
+    if (mbr == NULL || part_list == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    // Set default arguments for MBR generation
+    esp_mbr_generate_extra_args_t args = {
+        .sector_size = part_list->sector_size != ESP_EXT_PART_SECTOR_SIZE_UNKNOWN ? part_list->sector_size : ESP_EXT_PART_SECTOR_SIZE_512B, // Default sector size
+        .alignment = ESP_EXT_PART_ALIGN_1MiB, // Default alignment
```
pacucha42, line 209: Why do we enforce 1MB partition alignment? ~~Isn't this option a good candidate for a parameter setting?~~ If not, why don't we use 4kB which would fit much better into usual IoT storage device requirements?

Sorry, I have missed ESP_EXT_PART_ALIGN_1MiB is just a default. Other questions are valid. Thanks

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,383 @@
...
+{
+    switch (type) {
+    case ESP_EXT_PART_TYPE_FAT12:
+        return 0x01; // FAT12
+    case ESP_EXT_PART_TYPE_FAT16:
+        return 0x06; // FAT16B with LBA addressing
+    case ESP_EXT_PART_TYPE_FAT32:
+        return 0x0C; // FAT32 with LBA addressing
+    /* 
+    LittleFS is not a standard MBR partition type, but we can use a custom type `0xC3`, which is not usually used nowadays.
```
pacucha42: Thanks for this description. Good example of a good in-code comment, as mentioned before :-)

esp_ext_part_tables/README.md
```diff
@@ -0,0 +1,39 @@
...
+    return err;
+}
+
+esp_ext_part_list_item_t* item;
+item = esp_ext_part_list_item_head(&part_list); // Get the first partition
+
+for (int i = 0; item != NULL; i++) {
+    printf("Partition %d:\n
+        LBA start: %" PRIu64 "\n
+        sector count: %" PRIu64 "\n
```
adokitkat: TODO: Fix, changed to address and size

esp_ext_part_tables/src/esp_ext_part_tables.c
```diff
@@ -0,0 +1,122 @@
...
+    if (part_list == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    esp_ext_part_list_item_t* _item = (esp_ext_part_list_item_t*) malloc(sizeof(esp_ext_part_list_item_t));
+    if (_item == NULL) {
+        return ESP_ERR_NO_MEM;
+    }
+
+    memcpy(_item, item, sizeof(esp_ext_part_list_item_t)); // Copy the item
```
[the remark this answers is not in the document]
adokitkat: Not in ESP-IDF.

esp_ext_part_tables/src/esp_ext_part_tables.c
```diff
@@ -0,0 +1,122 @@
...
+    esp_ext_part_list_item_t* it = NULL;
+    esp_ext_part_list_item_t* tmp = NULL;
+    SLIST_FOREACH_SAFE(it, &part_list->head, next, tmp) {
+        SLIST_REMOVE(&part_list->head, it, esp_ext_part_list_item_, next);
+        if (it->info.label != NULL) {
+            free(it->info.label); // Deep free the label if it was allocated
+        }
+        free(it); // Free the item itself
+    }
+    memset(part_list, 0, sizeof(esp_ext_part_list_t)); // Reset the part_list structure
```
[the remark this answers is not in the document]
adokitkat: Not in ESP-IDF.

esp_ext_part_tables/src/esp_ext_part_tables.c
```diff
@@ -0,0 +1,122 @@
...
+esp_err_t esp_ext_part_list_signature_get(esp_ext_part_list_t* part_list, void* signature)
+{
+    if (part_list == NULL || signature == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    switch (part_list->signature.type) {
+    case ESP_EXT_PART_LIST_SIGNATURE_MBR:
+        uint32_t out = (uint32_t) part_list->signature.data[0];
+        memcpy(signature, &out, sizeof(uint32_t));
```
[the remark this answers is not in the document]
adokitkat: Not in ESP-IDF.

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,362 @@
...
+        sector = MBR_CHS_MAX_SECTOR;
+    }
+
+    uint8_t chs_bytes[3];
+    chs_bytes[0] = head & 0xFF;
+    chs_bytes[1] = ((cylinder >> 2) & 0xC0) | (sector & 0x3F); // high 2 bits of cylinder + 6-bit sector
+    chs_bytes[2] = cylinder & 0xFF;
+
+    if (chs != NULL) {
+        memcpy(chs, chs_bytes, 3);
```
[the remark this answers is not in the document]
adokitkat: Not in ESP-IDF.

esp_ext_part_tables/src/esp_mbr.c
adokitkat: I do it for FAT partitions to stay as close to FATFS library behavior ESP-IDF is using. FATFS library can format partitions and generate MBR as well and it does fill CHS fields. However technically it isn't needed because modern systems should ignore it in favor of LBA.

esp_ext_part_tables/src/esp_mbr.c
adokitkat: SD cards present themselves with minimal read/write size of 512B, which is 1 sector. However internally NAND page sizes can be from 4kB to 64kB. If the accessed data is misaligned on those boundaries, it can lead to slower performance. This sector and page sizes are also different for different storage media.

I used 1MiB as a default because it is dividable by all of those sizes and should be always properly aligned, plus this component is to be used with SD cards or maybe USB drives whose capacities are in from tens to thousands of gigabytes range. If this component is to be used with a media of smaller capacity, 4kiB still can be chosen.

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -0,0 +1,215 @@
...
+    ESP_EXT_PART_LIST_FLAG_NONE = 0,
+    ESP_EXT_PART_LIST_FLAG_READ_ONLY = 1 << 0, /*!< Read-only partition list */
+} esp_ext_part_list_flags_t;
+
+typedef enum {
+    ESP_EXT_PART_LIST_SIGNATURE_MBR, /*!< MBR signature type */
+} esp_ext_part_signature_type_t;
+
+typedef struct {
+    uint32_t data[1];
```
pacucha42, line 66: 🙂 what is the purpose of this declaration, please?

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * @brief Convert bytes to sector count based on the sector size.
+ *
+ * This function performs a ceiling division to ensure that any remaining bytes
+ * that do not fill a complete sector are counted as an additional sector.
+ *
+ * @param total_bytes Total number of bytes.
+ * @param sector_size Size of a single sector.
+ *
+ * @return Total size in bytes.
```
pacucha42: This function apparently returns the number of sectors needed to envelope required byte array - please, fix the description

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -0,0 +1,215 @@
...
+ */
+uint64_t esp_ext_part_sector_count_to_bytes(uint64_t sector_count, esp_ext_part_sector_size_t sector_size);
+
+/**
+ * @brief Deinitialize an external partition list structure and free all resources.
+ *
+ * This function releases all memory and resources associated with the partition list,
```
pacucha42: Nitpicking: grammar, phrasing. Suggestion:

This function releases all the memory and resources associated with the partition list referenced by 'part_list' parameter.

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -0,0 +1,215 @@
...
+    esp_ext_part_list_flags_t flags; /*!< Flags for the partition list */
+    esp_ext_part_sector_size_t sector_size; /*!< Sector size (storage medium property) */
+} esp_ext_part_list_t;
+
+/**
+ * @brief Convert bytes to sector count based on the sector size.
+ *
+ * This function performs a ceiling division to ensure that any remaining bytes
+ * that do not fill a complete sector are counted as an additional sector.
+ *
```
pacucha42, line 96: Missing explanation of retval = 0. See 
`    if (sector_size == ESP_EXT_PART_SECTOR_SIZE_UNKNOWN) {
        return 0; // Avoid division by zero
    }`
in the function body

esp_ext_part_tables/include/esp_ext_part_tables.h
adokitkat: I made the disk signature a separate struct type instead of using uint32_t or uint64_t in case we were to support e.g. GPT partition tables in the future (which is 128 bits / 16 bytes long). This way the signature of the struct doesn't change, only the size of `data` array, i.e. `uint32_t data[4];`.

pacucha42: Ok, I see. Thanks for the explanation

esp_ext_part_tables/src/esp_ext_part_tables.c
```diff
@@ -0,0 +1,159 @@
...
+        }
+        free(it); // Free the item itself
+    }
+    memset(part_list, 0, sizeof(esp_ext_part_list_t)); // Reset the part_list structure
+    return ESP_OK;
+}
+
+esp_err_t esp_ext_part_list_insert(esp_ext_part_list_t* part_list, esp_ext_part_list_item_t* item)
+{
+    if (part_list == NULL) {
```
pacucha42: Once we check 'part_list'for NULL, maybe we should do the same 'item' too?

If you agree, extend also the Doxygen info in the header. Thanks

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -0,0 +1,215 @@
...
+ * @param[out] dst Pointer to the destination partition list structure (must be allocated before but not initialized, i.e. "empty").
+ * @param[in] src Pointer to the source partition list structure to copy from.
+ *
+ * @return
+ *     - ESP_OK: Deep copy was successful.
+ *     - ESP_ERR_INVALID_ARG: dst or src is NULL.
+ *     - ESP_ERR_NO_MEM: Memory allocation failed.
+ */
+esp_err_t esp_ext_part_list_deep_copy(esp_ext_part_list_t* dst, esp_ext_part_list_t* src);
```
pacucha42: Just wondering: what would be use-case for such a function?

esp_ext_part_tables/include/esp_mbr_utils.h
```diff
@@ -0,0 +1,33 @@
...
+#endif
+
+#define MBR_CHS_HEADS 255
+#define MBR_CHS_SECTORS_PER_TRACK 63
+#define MBR_CHS_MAX_CYLINDER 1023
+#define MBR_CHS_MAX_HEAD 254
+#define MBR_CHS_MAX_SECTOR 63
+
+// Helper functions for MBR CHS conversion and LBA alignment
+void esp_mbr_chs_arr_val_set(uint8_t chs[3], uint32_t val);
```
pacucha42, line 30: It seems the helper functions are used only in` esp_mbr.c`. If so, please make them static and move them to the C module.

Made I already commented similar setup elsewhere 🙂 

If those APIs are supposed to stay public, add the Doxygen desc, please

esp_ext_part_tables/src/esp_ext_part_tables.c
```diff
@@ -0,0 +1,159 @@
...
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    esp_ext_part_list_item_t* it = NULL;
+    esp_ext_part_list_item_t* tmp = NULL;
+    SLIST_FOREACH_SAFE(it, &part_list->head, next, tmp) {
+        SLIST_REMOVE(&part_list->head, it, esp_ext_part_list_item_, next);
+        if (it->info.label != NULL) {
```
pacucha42: Nitpicking: free() does the NULL check internally, no need for extra if. Applies to all the code

esp_ext_part_tables/src/esp_ext_part_tables.c
```diff
@@ -0,0 +1,159 @@
...
+esp_err_t esp_ext_part_list_signature_get(esp_ext_part_list_t* part_list, void* signature)
+{
+    if (part_list == NULL || signature == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    switch (part_list->signature.type) {
+    case ESP_EXT_PART_LIST_SIGNATURE_MBR:
+        uint32_t out = (uint32_t) part_list->signature.data[0];
+        memcpy(signature, &out, sizeof(uint32_t));
```
pacucha42, line 132: Is this line necessary?

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,383 @@
...
+
+    mbr_partition_t* partition;
+    for (int i = 0; i < 4; i++) {
+        partition = (mbr_partition_t*) &mbr->partition_table[i];
+
+        // Check if the partition entry is empty and if so, skip it
+        bool is_empty = true;
+        for (uint8_t i = 0; i < sizeof(mbr_partition_t); i++) {
+            if (((uint8_t *)partition)[i] != 0) {
+                is_empty = false;
```
pacucha42: Missing break? Maybe direct return?

esp_ext_part_tables/README.md
adokitkat: fixed

esp_ext_part_tables/src/esp_ext_part_tables.c
adokitkat: I used `memcpy` in case `void* signature` points to to an array or something for some reason and `out` is just a temporary variable so the code is cleaner (will probably be optimized out anyway?).

esp_ext_part_tables/src/esp_mbr.c
adokitkat: Changed to checking 0x00 partition type.

esp_ext_part_tables/src/esp_mbr.c
adokitkat: Changed to 0x00 partition type checking.

esp_ext_part_tables/include/esp_ext_part_tables.h
adokitkat: To be honest I thought I need it for something in an example code but then I resolved the problem another way and the function stayed here. I just thought copying the structure is non-trivial due to pointers and malloc'd memory so I created this function (i.e. the logic is hidden here and if there were changes in the part list struct users wouldn't need to change anything in their code). However right now I am not sure if this functionality needs to be in the public API (if and why it is something users would need often), so I can remove it.

adokitkat: One use case may be if the user wanted to load the MBR and make changes to it, while also keeping the original part list created from the MBR as "a reference" / a backup, so the changes would be made to a copy.

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -0,0 +1,425 @@
...
+
+static void ext_part_list_item_do_extra(esp_ext_part_list_item_t *item, mbr_partition_t *partition)
+{
+    // This function is for any extra actions that might be needed for specific partition types.
+    // It can be used to set flags, perform additional operations or checks if needed.
+
+    switch (item->info.type) { // Parsed type
+    case ESP_EXT_PART_TYPE_LITTLEFS:
+        item->info.flags |= ESP_EXT_PART_FLAG_EXTRA; // Set the extra flag to indicate that this partition has extra information
+        item->info.extra = (uint64_t) esp_mbr_chs_arr_val_get(partition->chs_start); // Put LittleFS block size which was stored in `chs_start` to `extra` field
```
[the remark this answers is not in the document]
adokitkat: extra field is now `uint64_t`

esp_ext_part_tables/examples/basic/README.md
```diff
@@ -0,0 +1,66 @@
...
+This example demonstrates how to use an ESP32 to read and parse the Master Boot Record (MBR) from a (micro)SD card. It shows how to extract and display information about the card's partitions, such as their type, size, and starting sector, using the ESP-IDF framework.
+
+## Requirements
+
+- any ESP32 board which supports SPI with a (micro)SD card slot or breakout board connected
+- formatted (micro)SD card using MBR (not GPT) as a partition table
+
+## Build and Flash
+
+The example runs on any ESP development board with SD card connected via SPI (use `idf.py menuconfig` -> `Example config` to set used GPIO pins). For example, to build and run the code on ESP32-S3, use:
```
pacucha42: Nitpicking: `The example` followed by `For example` sounds a bit odd, maybe you could rephrase the second sentence as `For instance, ...`. Pure cosmetics, can be done in a follow-up

esp_ext_part_tables/examples/basic/main/example_utils.c
```diff
@@ -0,0 +1,124 @@
...
+        return ret;
+    }
+
+    ret = sdmmc_read_sectors(&card, mbr_buffer, 0, 1);
+    if (ret != ESP_OK) {
+        ESP_LOGE(TAG, "Failed to read first sector from SD card.");
+        spi_bus_free(host.slot);
+        return ret;
+    }
+#endif // CONFIG_IDF_TARGET_LINUX
```
pacucha42: Nitpicking: this should likely be `! CONFIG_IDF_TARGET_LINUX`, as the code part is used when not compilimg for Linux. Not important, though

esp_ext_part_tables/examples/basic/main/main.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+void esp_ext_part_tables_mbr_parse_example_task(void *pvParameters)
+{
+    TaskHandle_t main_task_handle;
+    ESP_LOGI(TAG, "Starting MBR parsing example task");
+    esp_err_t err;
+
+    // Allocate memory for the MBR
+    mbr_t *mbr = (mbr_t *) heap_caps_malloc(sizeof(mbr_t), (MALLOC_CAP_DMA | MALLOC_CAP_8BIT));
```
pacucha42, line 47: Just wondering: common `malloc` doesn't work well? Why restricted to DMA/8-bit?

esp_ext_part_tables/examples/basic/main/main.c
```diff
@@ -0,0 +1,204 @@
...
+    vTaskDelete(NULL); // Delete the current task
+}
+
+void app_main(void)
+{
+    ESP_LOGI(TAG, "Example started");
+    TaskHandle_t main_task_handle = xTaskGetCurrentTaskHandle(); // Get the handle of the main task
+
+    // Create new tasks with bigger stack size just in case
+    xTaskCreate(esp_ext_part_tables_mbr_parse_example_task, "esp_ext_part_tables_mbr_parse_example_task", 4096, (void *) main_task_handle, 5, NULL);
```
pacucha42, line 197: Wondering again: why do you run extra tasks once the processing is serialized anyway? Isn't it cheaper to stay in one thread? If not, could you please describe the use you cover here?

esp_ext_part_tables/examples/basic/main/example_utils.c
adokitkat: It has this structure

```c
#if CONFIG_IDF_TARGET_LINUX

#else

#endif // CONFIG_IDF_TARGET_LINUX
```

How should I signify it ends `CONFIG_IDF_TARGET_LINUX`?

esp_ext_part_tables/examples/basic/main/main.c
adokitkat: I used this variant because this buffer is passed to `sdmmc_read_sectors(&card, mbr_buffer, 0, 1);` so it can directly use it. Using malloc is also ok, after all we are reading only one sector. Should I change it to malloc?

esp_ext_part_tables/examples/basic/main/main.c
adokitkat: I wanted it to have an access to a larger stack, but now I am not sure if it is needed as it has been quite some time ago.

esp_ext_part_tables/examples/basic/main/example_utils.c
```diff
@@ -0,0 +1,124 @@
...
+#else
+    // This function loads the first sector (MBR) from the SD card into the provided buffer
+    // It uses SDSPI but can be adapted for SDMMC as well
+    esp_err_t ret = ESP_OK;
+    sdmmc_host_t host = SDSPI_HOST_DEFAULT();
+
+    // For SoCs where the SD power can be supplied both via an internal or external (e.g. on-board LDO) power supply.
+    // When using specific IO pins (which can be used for ultra high-speed SDMMC) to connect to the SD card
+    // and the internal LDO power supply, we need to initialize the power supply first.
+#if CONFIG_EXAMPLE_SD_PWR_CTRL_LDO_INTERNAL_IO
```
pacucha42, line 61: I don't think this part is necessary when running in SDSPI mode - afaik, you use high speed setup only in direct SD mode. Please, double check and possibly remove this unnecessary code. Thanks

esp_ext_part_tables/examples/basic/main/example_utils.c
pacucha42: As indicated 🙂 - `! CONFIG_IDF_TARGET_LINUX`. Small readability update, not important thing

esp_ext_part_tables/examples/basic/main/main.c
pacucha42: Ok, thanks for the explanation. Please, keep it as is

esp_ext_part_tables/examples/basic/main/example_utils.c
adokitkat: It is also present in sdspi example in esp-idf, not only sdmmc example. (I had to add it there 18 months back to make it work with P4.)

esp_ext_part_tables/examples/basic/main/main.c
pacucha42: There is Kconfig option named "Main task stack size" (ESP_MAIN_TASK_STACK_SIZE) in Menuconfig's Component config → ESP System Settings.

It is not critical point, though. I just didn't see any sense in this piece of code. Thanks for the explanation

esp_ext_part_tables/examples/basic/README.md
adokitkat: Rephrased.

pacucha42:
@adokitkat - all my comments seem addressed. LGTM, thanks.

Please, also revisit your future updates/todos in this PR description and create appropriate trackers.

esp_ext_part_tables/examples/basic/main/example_utils.c
pacucha42: Well, ok, but it still doesn't make any sense there

pacucha42: Based on internal discussion: here, the LDO is used only as an external power source at 3.3V and is not actually used for any high-speed support (which is truly not available in SD SPI). Sorry, my bad, keep the code as is. Thanks

pacucha42 (approved):
Approved

PR #530, opened 2025-07-25
feat(esp_cli_commands): add esp_cli_commands component (IEC-338)

SoucheSouche:
# Checklist

- [x] CI passing

# Change description

### DIFF
--- a/.build-test-rules.yml
+++ b/.build-test-rules.yml
@@ -34,6 +34,11 @@ esp_jpeg/examples/get_started:
     - if: IDF_VERSION_MAJOR > 4 and IDF_TARGET in ["esp32", "esp32s2", "esp32s3"]
       reason: Example depends on BSP, which is supported only for IDF >= 5.0 and limited targets
 
+esp_schedule/examples/get_started:
+  enable:
+    - if: ((IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR >= 1) or (IDF_VERSION_MAJOR > 5)) and SOC_WIFI_SUPPORTED == 1
+      reason: Network provisioning component has dependencies IDF >= 5.1; example only supports Wi-Fi enabled targets
+
 catch2/examples/catch2-test:
   enable:
     - if: INCLUDE_DEFAULT == 1 or IDF_TARGET == "linux"
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -39,6 +39,7 @@ body:
         - esp_isotp
         - esp_jpeg
         - esp_lcd_qemu_rgb
+        - esp_schedule
         - esp_serial_slave_link
         - expat
         - fmt
--- a/.github/workflows/upload_component.yml
+++ b/.github/workflows/upload_component.yml
@@ -43,6 +43,7 @@ jobs:
             esp_lcd_qemu_rgb
             esp_linenoise
             esp_jpeg
+            esp_schedule
             esp_serial_slave_link
             expat
             fmt
--- a/.idf_build_apps.toml
+++ b/.idf_build_apps.toml
@@ -14,6 +14,7 @@ manifest_file = [
     "esp_gcov/.build-test-rules.yml",
     "esp_jpeg/.build-test-rules.yml",
     "esp_linenoise/.build-test-rules.yml",
+    "esp_schedule/.build-test-rules.yml",
     "esp_serial_slave_link/.build-test-rules.yml",
     "expat/.build-test-rules.yml",
     "iqmath/.build-test-rules.yml",
--- a/coap/sbom_libcoap.yml
+++ b/coap/sbom_libcoap.yml
@@ -12,3 +12,5 @@ cve-exclude-list:
     reason: Resolved in version 4.3.5-rc1
   - cve: CVE-2024-46304
     reason: Resolved in version 4.3.5-rc3
+  - cve: CVE-2025-50518
+    reason: Not applicable as per comment https://github.com/obgm/libcoap/issues/1724#issuecomment-3296780541
--- a/esp_gcov/CMakeLists.txt
+++ b/esp_gcov/CMakeLists.txt
@@ -59,7 +59,7 @@ if(CONFIG_ESP_GCOV_ENABLE)
 
     # Register component with app_trace dependency
     idf_component_register(SRCS "gcov_rtio.c"
-                          REQUIRES "app_trace"
+                          REQUIRES "esp_trace"
                           INCLUDE_DIRS ".")
 
     # Configure gcov-specific flags
@@ -74,6 +74,6 @@ if(CONFIG_ESP_GCOV_ENABLE)
 
 else()
     # Register empty component when GCOV is disabled
-    idf_component_register(REQUIRES "app_trace")
+    idf_component_register(REQUIRES "esp_trace")
     message(STATUS "GCOV: Component registered but GCOV support is disabled")
 endif()
--- a/esp_gcov/Kconfig
+++ b/esp_gcov/Kconfig
@@ -2,10 +2,10 @@ menu "GNU Code Coverage"
 
     config ESP_GCOV_ENABLE
         bool "GCOV to Host Enable"
-        depends on APPTRACE_ENABLE && !APPTRACE_SV_ENABLE
+        depends on ESP_TRACE_TRANSPORT_APPTRACE && ESP_TRACE_LIB_NONE
         select ESP_DEBUG_STUBS_ENABLE
         select ESP_IPC_ENABLE
-        default y
+        default n
         help
             Enables support for GCOV data transfer to host.
 
--- a/esp_gcov/gcov_rtio.c
+++ b/esp_gcov/gcov_rtio.c
@@ -39,7 +39,7 @@ void gcov_dump_task(void *pvParameter)
         goto gcov_exit;
     }
     ESP_EARLY_LOGV(TAG, "Config apptrace down buf");
-    esp_err_t res = esp_apptrace_down_buffer_config(ESP_APPTRACE_DEST_JTAG, down_buf, ESP_GCOV_DOWN_BUF_SIZE);
+    esp_err_t res = esp_apptrace_down_buffer_config(down_buf, ESP_GCOV_DOWN_BUF_SIZE);
     if (res != ESP_OK) {
         ESP_EARLY_LOGE(TAG, "Failed to config apptrace down buf (%d)!", res);
         dump_result = res;
@@ -51,7 +51,7 @@ void gcov_dump_task(void *pvParameter)
     __gcov_reset();
     free(down_buf);
     ESP_EARLY_LOGV(TAG, "Finish file transfer session");
-    dump_result = esp_apptrace_fstop(ESP_APPTRACE_DEST_JTAG);
+    dump_result = esp_apptrace_fstop();
     if (dump_result != ESP_OK) {
         ESP_EARLY_LOGE(TAG, "Failed to send files transfer stop cmd (%d)!", dump_result);
     }
@@ -117,7 +117,7 @@ void esp_gcov_dump(void)
 {
     ESP_EARLY_LOGV(TAG, "%s", __FUNCTION__);
 
-    while (!esp_apptrace_host_is_connected(ESP_APPTRACE_DEST_JTAG)) {
+    while (!esp_apptrace_host_is_connected()) {
         vTaskDelay(pdMS_TO_TICKS(10));
     }
 
@@ -132,7 +132,7 @@ void esp_gcov_dump(void)
 void *gcov_rtio_fopen(const char *path, const char *mode)
 {
     ESP_EARLY_LOGV(TAG, "%s '%s' '%s'", __FUNCTION__, path, mode);
-    void *f = esp_apptrace_fopen(ESP_APPTRACE_DEST_JTAG, path, mode);
+    void *f = esp_apptrace_fopen(path, mode);
     ESP_EARLY_LOGV(TAG, "%s ret %p", __FUNCTION__, f);
     return f;
 }
@@ -140,13 +140,13 @@ void *gcov_rtio_fopen(const char *path, const char *mode)
 int gcov_rtio_fclose(void *stream)
 {
     ESP_EARLY_LOGV(TAG, "%s", __FUNCTION__);
-    return esp_apptrace_fclose(ESP_APPTRACE_DEST_JTAG, stream);
+    return esp_apptrace_fclose(stream);
 }
 
 size_t gcov_rtio_fread(void *ptr, size_t size, size_t nmemb, void *stream)
 {
     ESP_EARLY_LOGV(TAG, "%s read %u", __FUNCTION__, size * nmemb);
-    size_t sz = esp_apptrace_fread(ESP_APPTRACE_DEST_JTAG, ptr, size, nmemb, stream);
+    size_t sz = esp_apptrace_fread(ptr, size, nmemb, stream);
     ESP_EARLY_LOGV(TAG, "%s actually read %u", __FUNCTION__, sz);
     return sz;
 }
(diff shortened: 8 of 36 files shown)


esp_commands/include/esp_commands.h
```diff
@@ -0,0 +1,275 @@
...
+ * @note see the linker.lf file for more information concerning the section characteristics
+ */
+#define ESP_COMMAND_REGISTER(cmd_name, cmd_group, cmd_help, cmd_func, cmd_func_w_ctx, cmd_func_ctx, cmd_hint_cb, cmd_glossary_cb) \
+    static_assert(((cmd_func) == NULL) != ((cmd_func_w_ctx) == NULL)); \
+    static const esp_command_t cmd_name __attribute__((used, section(".esp_commands"))) = { \
+        .name = #cmd_name, \
+        .group = #cmd_group, \
+        .help = cmd_help, \
+        .func = cmd_func, \
+        .func_w_ctx = cmd_func_w_ctx, \
```
ESP-Marius: I wonder it makes sense to just delete this and make the default one have a context instead

esp_commands/include/esp_commands.h
```diff
@@ -0,0 +1,275 @@
...
+ */
+#define FIELD_ACCESSOR(NAME) get_##NAME
+
+/**
+ * @brief Configuration parameters for esp_commands_manager initialization
+ */
+typedef struct esp_commands_config {
+    size_t max_cmdline_length;  /*!< length of command line buffer, in bytes */
+    size_t max_cmdline_args;    /*!< maximum number of command line arguments to parse */
+    uint32_t heap_alloc_caps;   /*!< where to (e.g. MALLOC_CAP_SPIRAM) allocate heap objects such as cmds used by esp_console */
```
ESP-Marius: I think we could drop this. It was mainly to allow cmds to be placed in SPIRAM, but now they will be mostly in flash.

Seems uncessary to let the user control this directly now. This also removes the public dependency on heap

esp_commands/esp_commands.c
```diff
@@ -0,0 +1,455 @@
...
+
+/** run-time configuration options */
+static esp_commands_config_t s_config = {
+    .heap_alloc_caps = MALLOC_CAP_DEFAULT
+};
+
+/** temporary buffer used for command line parsing */
+static char *s_tmp_line_buf;
+
+esp_err_t esp_commands_init(const esp_commands_config_t *config)
```
ESP-Marius: With the current command implementation, it is not possible to run multiple consoles in parallel is it? (e.g. s_tmp_line_buf is shared)

Is allowing this the plan? I thought we wanted to be able to run multiple repls or similar implementions in parallel (hence the changes to make linenoise capable of this)

esp_commands/include/esp_commands.h
```diff
@@ -0,0 +1,275 @@
...
+ */
+#define ESP_COMMAND_REGISTER(cmd_name, cmd_group, cmd_help, cmd_func, cmd_func_w_ctx, cmd_func_ctx, cmd_hint_cb, cmd_glossary_cb) \
+    static_assert(((cmd_func) == NULL) != ((cmd_func_w_ctx) == NULL)); \
+    static const esp_command_t cmd_name __attribute__((used, section(".esp_commands"))) = { \
+        .name = #cmd_name, \
+        .group = #cmd_group, \
+        .help = cmd_help, \
+        .func = cmd_func, \
+        .func_w_ctx = cmd_func_w_ctx, \
+        .func_ctx = cmd_func_ctx, \
```
ESP-Marius, line 37: I wonder if we still need to provide a way to register commands dynamically. A context will in many cases be something not known at compile time, wont it? :thinking:

esp_commands/include/esp_commands.h
SoucheSouche: agreed, I removed the field from the config structure.

esp_commands/esp_commands.c
SoucheSouche: I updated the component to allocate the temp buffer in the exec command.

esp_commands/include/esp_commands.h
SoucheSouche: updated.

esp_commands/include/esp_commands.h
SoucheSouche: I added the support for dynamic registration.

esp_commands/include/esp_commands.h
```diff
@@ -0,0 +1,303 @@
...
+    esp_command_hint_t hint_cb; /*!< Callback returning the hint for the command */
+    esp_command_glossary_t glossary_cb; /*!< Callback returning the glossary for the command */
+} esp_command_t;
+
+/**
+ * @brief Macro to define a forced inline accessor for a string field of esp_command_t
+ *
+ * @param NAME Field name of the esp_command_t structure
+ */
+#define DEFINE_FIELD_ACCESSOR(NAME) \
```
ESP-Marius: Do you think these are likely to be used by users? if not we can simplify the API interface and drop them (or move to e.g. a helper header to keep the main interface clean?)

SoucheSouche:
@igrr, could you please take a look at the overall look of the component and give feedbacks about the API?

You will find it fairly similar to the idf console `commands.c` since the main functionalities (register / unregister and execute) remain. However the static command registration strategy and command subsets handling are newly added features and I would appreciate your opinion on the way they are implemented in this component.

esp_commands/.build-test-rules.yml
```diff
@@ -0,0 +1,4 @@
+esp_commands/test_apps:
+  enable:
+    - if: (IDF_TARGET == "linux") and (IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR >= 3)
```
ESP-Marius: Since it depends on .lf the placement of sections in memory could potentially fail on chip but not on linux right?

Consider if this is a sufficient reason to also test on target.

esp_commands/README.md
```diff
@@ -0,0 +1,183 @@
...
+```c
+static int my_cmd(void *ctx, int argc, char **argv) {
+    printf("Hello from my_cmd!\n");
+    return 0;
+}
+
+ESP_COMMAND_REGISTER(my_cmd, tools, "Prints hello", my_cmd, NULL, NULL, NULL);
+```
+
+This places the command into the `.esp_commands` section.
```
ESP-Marius: section in flash

esp_commands/include/esp_commands.h
```diff
@@ -0,0 +1,326 @@
...
+#define ESP_COMMAND_STRINGIFY(name) #name
+#define ESP_COMMAND_
+
+/**
+ * @brief Console command main function type with user context
+ *
+ * This function type is used to implement a console command.
+ *
+ * @param context User-defined context passed at invocation
+ * @param fd_out The file descriptor to use to output data
```
ESP-Marius: What does output data mean in this context?

esp_commands/README.md
```diff
@@ -0,0 +1,183 @@
...
+    esp_command_glossary_t glossary_cb;   /*!< Glossary callback */
+} esp_command_t;
+```
+
+### Static Registration
+
+Use the `ESP_COMMAND_REGISTER` macro to register a command at compile time:
+
+```c
+static int my_cmd(void *ctx, int argc, char **argv) {
```
ESP-Marius: what about the `fd`? Is it missing here?

esp_commands/.build-test-rules.yml
SoucheSouche: You are right. I added esp32 to the list of target to test on

esp_commands/include/esp_commands.h
SoucheSouche: I moved the `DEFINE_XX` to a separate header as suggested. I would keep these macros together with the `FIELD_ACCESSOR` one because it makes it easier to use e.g., `esp_commands_create_cmd_set` since this function needs to know on what specific field the set of commands if created.

esp_commands/README.md
```diff
@@ -0,0 +1,183 @@
...
+### Dynamic Registration
+
+Commands can also be registered/unregistered at runtime:
+
+```c
+esp_command_t cmd = {
+    .name = "echo",
+    .group = "utils",
+    .help = "Echoes arguments back",
+    .func = echo_func,
```
ESP-Marius: Do we use a different function from my_cmd for a reason here? Seems clearer if both registration examples use the same one.

esp_cli_commands/include/esp_cli_commands_utils.h
```diff
@@ -0,0 +1,181 @@
...
+typedef ssize_t (*esp_commands_write_t)(int fd, const void *buf, size_t count);
+
+/**
+ * @brief Structure containing dynamic argument necessary for the\
+ * command callback to execute properly
+ */
+typedef struct esp_commands_exec_arg {
+    int out_fd; /*!< file descriptor that the command function has to use to print data in the environment it was called from */
+    esp_commands_write_t write_func; /*!< write function the command function has to use to output datga */
+    void *dynamic_ctx; /*!< dynamic context passed to the command function */
```
ESP-Marius, line 40: What is this for? Seems like it is passing the `cmd_set` to it but i am not sure i completely understand it

esp_commands/include/esp_commands_utils.h
```diff
@@ -0,0 +1,181 @@
...
+ */
+typedef ssize_t (*esp_commands_write_t)(int fd, const void *buf, size_t count);
+
+/**
+ * @brief Structure containing dynamic argument necessary for the\
+ * command callback to execute properly
+ */
+typedef struct esp_commands_exec_arg {
+    int out_fd; /*!< file descriptor that the command function has to use to print data in the environment it was called from */
+    esp_commands_write_t write_func; /*!< write function the command function has to use to output datga */
```
ESP-Marius:
```suggestion
    esp_commands_write_t write_func; /*!< write function the command function has to use to output data */
```

esp_commands/idf_component.yml
```diff
@@ -0,0 +1,9 @@
+version: "1.0.0"
```
ESP-Marius: If we are not 100% sure about the components (API) stability  we could merge it as 0.1.0 instead

esp_commands/include/esp_commands_utils.h
```diff
@@ -0,0 +1,181 @@
...
+ * @param count Number of bytes to write.
+ * @return Number of bytes written, or -1 on error.
+ */
+typedef ssize_t (*esp_commands_write_t)(int fd, const void *buf, size_t count);
+
+/**
+ * @brief Structure containing dynamic argument necessary for the\
+ * command callback to execute properly
+ */
+typedef struct esp_commands_exec_arg {
```
ESP-Marius: I think it would be nice to describe the idea behind this a bit clearer somewhere. 

Mention that if command is used from a different "channel" than where normal print/log goes then printf/log will output to the "wrong" place. 

Without any additional explaination i am not sure everyone will understand the purpose

esp_commands/README.md
```diff
@@ -0,0 +1,183 @@
...
+- Customizable configuration for command parsing and hint display.
+
+---
+
+## Configuration
+
+The component is initialized with a configuration struct:
+
+```c
+esp_commands_config_t config = ESP_COMMANDS_CONFIG_DEFAULT();
```
ESP-Marius: ESP_COMMANDS_CONFIG_DEFAULT: does this still exist?

esp_commands/src/esp_commands.c
```diff
@@ -0,0 +1,720 @@
...
+extern esp_command_t _esp_commands_start;
+extern esp_command_t _esp_commands_end;
+
+typedef struct esp_command_sets {
+    esp_command_set_t static_set;
+    esp_command_set_t dynamic_set;
+} esp_command_sets_t;
+
+/** run-time configuration options */
+static esp_commands_config_t s_config = {
```
ESP-Marius: static data, so is there a small potential for issues here if one thread updates config while another is executing commands?

esp_commands/src/esp_commands.c
```diff
@@ -0,0 +1,720 @@
...
+    } else if (cmd_set_a && !cmd_set_b) {
+        return cmd_set_a;
+    } else if (!cmd_set_a && cmd_set_b) {
+        return cmd_set_b;
+    }
+
+    /* Reaching this point, both cmd_set_a and cmd_set_b are set.
+     * Create a new cmd_set that can host the items from both sets,
+     * assign the items to the new set and free the input sets */
+    esp_command_sets_t *concat_cmd_sets = heap_caps_malloc(sizeof(esp_command_sets_t), s_config.heap_caps_used);
```
ESP-Marius: We defined our own malloc but still use this way of doing it a few places. Maybe ocnsider if the custom malloc function is necessary

esp_commands/test_apps/main/test_esp_commands.c
```diff
@@ -0,0 +1,462 @@
...
+
+    /* call esp_commands_execute to run help command on a registered command with wrong
+     * verbosity syntax */
+    cmd_ret = -1;
+    TEST_ASSERT_EQUAL(ESP_OK, esp_commands_execute(NULL, NULL, "help cmd_a -v=1", &cmd_ret));
+    TEST_ASSERT_EQUAL(1, cmd_ret);
+
+    /* call esp_commands_execute to run help command with too many command names */
+    cmd_ret = -1;
+    TEST_ASSERT_EQUAL(ESP_OK, esp_commands_execute(NULL, NULL, "help cmd_a cmd_b -v 1", &cmd_ret));
```
ESP-Marius: nitpick: In general i think if we have "optional" arguments (which are often null). it might make more sense to place them last.

esp_cli_commands/test_apps/main/test_esp_cli_commands.c
```diff
@@ -0,0 +1,462 @@
...
+    TEST_ASSERT_EQUAL(0, strcmp(dyn_cmd_a_msg_glossary_bis, ctx_a.message));
+
+    const char *dyn_cmd_b_msg_glossary_bis = esp_commands_get_glossary(handle_set, "dyn_cmd_b");
+    TEST_ASSERT_NULL(dyn_cmd_b_msg_glossary_bis);
+
+    TEST_ASSERT_EQUAL(ESP_OK, esp_commands_unregister_cmd("dyn_cmd_a"));
+    TEST_ASSERT_EQUAL(ESP_OK, esp_commands_unregister_cmd("dyn_cmd_b"));
+
+    esp_commands_destroy_cmd_set(&handle_set);
+}
```
ESP-Marius, line 462: Maybe trivial, but we could consider adding a test for the legacy style cmd functions to show that we can still easily use those from a wrapper function

ESP-Marius: That test-case would also add some documention for the legacy behavior for the future

ESP-Marius (commented):
Overall looks good to me @SoucheSouche, just left some minor comments

esp_cli_commands/include/esp_cli_commands_utils.h
SoucheSouche: this structure is passed by the caller of `esp_commands_execute`. It is just forwarded to the callback function associated with the command being called by the user. So dynamic_ctx can be anything the user wants to transfer to the command callback function.

In the case of the help command, since it has been registered by this component, it is natural that the dynamic_ctx is set by esp_commands itself. Since the help command needs to know is the esp_commands_execute function was called with a given `cmd_set` I update the `dynamic_ctx` passed to the help command function to `cmd_set`.

I updated the logic in `esp_commands_execute` to reflect this idea.

esp_cli_commands/test_apps/main/test_esp_cli_commands.c
SoucheSouche: I think this test shouldn't be implemented in this component. technically, we should be oblivious of the IDF console in this component. Outside of your knowledge and mine, this component has not been developed to supplant the command related implementation of the IDF console.

Those tests will be performed within IDF after the IDF console component is updated to use this component in the background.

esp_commands/src/esp_commands.c
SoucheSouche: good point.

esp_commands/src/esp_commands.c
```diff
@@ -0,0 +1,761 @@
...
+{
+    if (!config ||
+            (config->max_cmdline_args == 0) ||
+            (config->max_cmdline_length == 0)) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    /* Initialize the mutex if not initialized. Since this function is not mandatory
+     * to be called for the component to work properly, it is possible that the mutex
+     * was never initialized before. */
```
ESP-Marius: Will the mutex behave as expected if the user doesnt call `esp_commands_update_config` but instead directly call execute (or other functions that can take the mutex)?

SoucheSouche: updated.

esp_cli_commands/src/esp_cli_commands.c
```diff
@@ -0,0 +1,771 @@
...
+        return NULL;
+    }
+
+    esp_cli_command_sets_t *cmd_ptr_sets = esp_cli_commands_malloc(sizeof(esp_cli_command_sets_t));
+    if (!cmd_ptr_sets) {
+        return NULL;
+    }
+
+
+    esp_cli_command_t *static_cmd_ptrs_temp[ESP_CLI_COMMANDS_COUNT];
```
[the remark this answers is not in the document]
SoucheSouche: won't fix.
start and stop symbols are bounds to a contiguous memory region. the subtraction operation is not dangerous.

PR #532, opened 2025-08-01
(feat): add esp idf legacy unit-test-app (IEC-343)

ESP-Marius:
# Checklist

- [ ] CI passing

# Change description
Adds the old legacy unit-test-app as an "example" component, allowing easy integration into any projects that still wants to use it with:

`idf.py create-project-from-example "espressif/unit-test-app:unit-test-app`


ESP-Marius:
@suda-morris PTAL

unit-test-app/examples/unit-test-app/tools/ElfUnitTestParser.py
```diff
@@ -0,0 +1,83 @@
...
+    import CreateSectionTable
+
+
+def get_target_objdump(idf_target: str) -> str:
+    toolchain_for_target = {
+        'esp32': 'xtensa-esp32-elf-',
+        'esp32s2': 'xtensa-esp32s2-elf-',
+        'esp32s3': 'xtensa-esp32s3-elf-',
+        'esp32c2': 'riscv32-esp-elf-',
+        'esp32c3': 'riscv32-esp-elf-',
```
suda-morris, line 24: what about other targets?

ESP-Marius: Seeing it is a legacy app and we dont really recommend using it anymore, Im not sure it make sense to update it for newer targets. (what you see here is simply whjat was in IDF)

ESP-Marius: But I guess we can be nice since this is a simple change, changed the default dict value to riscv toolchain

PR #534, opened 2025-08-05
fix(esp_gcov): ipc enabled by default (IEC-345)

erhankur:
Enable IPC by default from the esp_gcov component to avoid external config dependency in the IDF.

Kconfig file renamed as Kconfig.gcov to allow importing from app_trace/Kconfig file


suda-morris:
@erhankur let's bump the component version to 1.0.1

erhankur:
> @erhankur let's bump the component version to 1.0.1

Done.

suda-morris (approved):

PR #535, opened 2025-08-06
feat(isotp):  Add ISO-TP (ISO 15765-2) Protocol Component with Examples and Testing Infrastructure (IEC-296)

eternal-echo:
# Checklist

## Required Items
- [x] Component contains idf_component.yml with url field
- [ ] Component added to upload job (NEEDS MANUAL ADDITION)
- [ ] Component added to build job (NEEDS MANUAL ADDITION)
- [x] CI passing

## Optional Items
- [x] Component contains unit tests (pytest)
- [x] QEMU testing support
- [x] Multiple examples

# Change Description

Complete ISO-TP protocol implementation for ESP-IDF enabling CAN bus communication with large payloads up to 4095 bytes (12-bit protocol limit).
- [x] Echo example with pytest
- [ ] Utils and OTA examples

## Features
- Full ISO-TP protocol support (single/multi-frame, flow control)
- ESP-IDF TWAI driver integration
- Linux can-utils compatibility  
- Automatic upstream patch management
- QEMU testing support

## Structure
```
isotp/
├── esp_isotp/           # ESP-IDF integration
├── isotp-c/             # Upstream library (submodule)
├── patch/               # Build-time patches
└── examples/
    ├── echo/            ✓ Message echo service
    ├── utils/           ⏳ CLI tools (isotpsend/recv)
    └── ota/             ⏳ Firmware upgrade system
```

## API Example
```c
esp_isotp_handle_t handle = esp_isotp_new(&config);
esp_isotp_send(handle, data, length);
esp_isotp_poll(handle);
int len = esp_isotp_receive(handle, buffer, size);
```

## Dependencies
- ESP-IDF v5.5.0+ (TWAI support)
- isotp-c library (submodule)


suda-morris:
@Dazza0 Please take a look if you're interested in. :smile:

suda-morris:
The upstream code is not actively maintained and may even contain security issues: https://github.com/lishen2/isotp-c/pull/36

I suggest we don't use git submodule but just directly copy the upstream source code and form our component. The proposed file structure could be:

```
esp_isotp/
├── include/      # esp_isotp.h
├── src/             # esp_isotp wrapper
├── isotp-c/      # Upstream library (source code with our patch)
├── test_apps/
└── examples/
```

@eternal-echo

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -39,6 +39,7 @@ body:
         - fmt
         - freetype
         - iqmath
+        - isotp
```
suda-morris: proposed the component name "esp_isotp"

isotp/esp_isotp/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+idf_component_register(SRCS "esp_isotp.c"
+                      INCLUDE_DIRS "."
+                      REQUIRES isotp driver)
```
suda-morris:
```suggestion
                      REQUIRES isotp esp_driver_twai)
```

isotp/esp_isotp/esp_isotp.c
```diff
@@ -0,0 +1,197 @@
...
+#include "freertos/queue.h"
+#include "esp_log.h"
+#include "esp_timer.h"
+#include "esp_twai.h"
+#include "esp_twai_types.h"
+#include "esp_check.h"
+#include "esp_isotp.h"
+#include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
```
suda-morris: libc header file first, then esp header files

isotp/esp_isotp/esp_isotp.c
```diff
@@ -0,0 +1,197 @@
+#include "freertos/FreeRTOS.h"
```
suda-morris: copyright

isotp/esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,105 @@
+#ifndef _ISOTP_ESP_H_
```
suda-morris: use pragma once

suda-morris: copyright

isotp/esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,105 @@
...
+#include "isotp-c/isotp.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief ISO-TP link structure
+ */
+struct esp_isotp_link {
```
suda-morris: can we hide it in the .c file?

isotp/esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,105 @@
...
+};
+typedef struct esp_isotp_config esp_isotp_config_t;
+
+/**
+ * @brief Create a new ISO-TP link
+ *
+ * @param config Pointer to the configuration structure
+ * @return Pointer to the new ISO-TP link, or NULL on error
+ */
+esp_isotp_handle_t esp_isotp_new(const esp_isotp_config_t *config);
```
suda-morris:
```suggestion
esp_isotp_handle_t esp_isotp_new_link(twai_node_handle_t twai_node, const esp_isotp_config_t *config);
```

or


```suggestion
esp_isotp_handle_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_config_t *config);
```

suda-morris: becasue ISO-TP is transport layer protocol, right? maybe use "new_transport" is more clear 🤔

isotp/esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,108 @@
...
+};
+typedef struct esp_isotp_config esp_isotp_config_t;
+
+/**
+ * @brief Create a new ISO-TP link
+ *
+ * @param config Pointer to the configuration structure
+ * @return Pointer to the new ISO-TP link, or NULL on error
+ */
+esp_isotp_handle_t esp_isotp_new_transport(const esp_isotp_config_t *config);
```
suda-morris: to make the ISOTP function looks like a wrapper over the twai node handle, we should highlight it in the first place. Instead of hiding the "twai_node_handle_t twai_node" inside the "esp_isotp_config", let's move it out


```suggestion
esp_isotp_handle_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_config_t *config);
```

suda-morris: and can we return esp_err_t?

isotp/esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * @brief Send data over an ISO-TP link
+ *
+ * @param handle The handle of the ISO-TP link
+ * @param data Pointer to the data to send
+ * @param size Size of the data to send
+ * @return ISOTP_RET_OK on success, ISOTP_RET_ERROR on error
+ */
+int esp_isotp_send(esp_isotp_handle_t handle, const uint8_t *data, uint16_t size);
```
suda-morris: can we return esp_err_t?

isotp/examples/echo/main/isotp_echo_main.c
```diff
@@ -0,0 +1,154 @@
...
+static esp_err_t isotp_echo_init(void)
+{
+    twai_onchip_node_config_t twai_cfg = {
+        .io_cfg = {
+            .tx = CONFIG_EXAMPLE_TX_GPIO_NUM,
+            .rx = CONFIG_EXAMPLE_RX_GPIO_NUM,
+            .quanta_clk_out = -1,
+            .bus_off_indicator = -1,
+        },
+        .clk_src = 0,
```
suda-morris: if we want to clear it to zero, we don't have to do in this way explicitly.


```suggestion
```

isotp/examples/echo/main/isotp_echo_main.c
```diff
@@ -0,0 +1,154 @@
...
+    esp_err_t ret = isotp_echo_init();
+    ESP_GOTO_ON_ERROR(ret, err, TAG, "Failed to initialize ISO-TP echo: %s", esp_err_to_name(ret));
+
+    while (1) {
+        vTaskDelay(pdMS_TO_TICKS(10000));
+    }
+
+    ESP_LOGI(TAG, "ISO-TP Echo Demo finished");
+
+err:
```
suda-morris: we don't do complicated error handling in our example, we recommend to panic on any error. Suggest using "assert" and "ESP_ERROR_CHECK".

esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * @brief Poll the ISO-TP link to process messages
+ *
+ * @param handle The handle of the ISO-TP link
+ * @return
+ *     - ESP_OK: Success
+ *     - ESP_ERR_INVALID_ARG: Invalid argument
+ */
+esp_err_t esp_isotp_poll(esp_isotp_handle_t handle);
```
Dazza0: I get that we've gone for a polling approach here because the underlying isotp-c library is also build around polling. However, polling might become problematic due to the following reasons:

- Default IDF FreeRTOS tick rate is 100Hz, meaning the minimum FreeRTOS delay is 10ms. Thus, if a series of CAN frame specifies all have a separation time (ST) of 1ms, using `vTaskDelay(1)` to poll ISOTP will result in a practical separation time of 10ms, leading to the total delivery to take 10x longer. Users could increase FreeRTOS tick rate to 1000Hz, but that comes at the cost of 10x FreeRTOS overhead.
- We could poll using a timer with higher resolution (like `esp_timer()`), but that also comes at the cost of increased overhead.

Fortunately, the entire isotp-c state is accessible in the `IsoTpLink` object. Thus, it might be possible for `esp_isotp` to offer an event driven interface. It could be something like:

- Let the user register a semaphore/task handle in `esp_isotp_new_transport()` to block on events. When the user unblocks on an event, they must call `esp_isotp_poll()`.
- Internally in `esp_isotp`, whenever an `isotp_...()` API call is made, we check certain members of the `IsoTpLink` to know if we need to propagate an event. If so, give to the semaphore/task notification. Examples of some members check are...
    - `send_st_min` which indicates the separation time in ms before the next CAN frame transmission can start. If we need < 10ms resolution, could possible set an `esp_timer` to propagate the event on timeout.
    - `send_timer_bs` after a CAN frame transmission. Indicates the timeout for the recipient to reply with a flow control frame. Again, could use `esp_timer` to trigger a call to the poll function when that timeout elapses.

esp_isotp/esp_isotp.c
```diff
@@ -0,0 +1,237 @@
...
+
+    return ESP_OK;
+}
+
+esp_err_t esp_isotp_send(esp_isotp_handle_t handle, const uint8_t *data, uint16_t size)
+{
+    ESP_RETURN_ON_FALSE(handle != NULL, ESP_ERR_INVALID_ARG, TAG, "Invalid handle");
+    ESP_RETURN_ON_FALSE(data != NULL, ESP_ERR_INVALID_ARG, TAG, "Invalid data");
+
+    int ret = isotp_send(&handle->link, data, size);
```
Dazza0: Is the `esp_isotp` restricted to single threaded calling contexts? For example, if an IDF application spins up multiple tasks, and those tasks both call `esp_isotp_send()` simultaneously, do we need to make sure that access to the `handle->link` data structure is thread safe?

esp_isotp/esp_isotp.c
```diff
@@ -0,0 +1,237 @@
...
+    ESP_RETURN_ON_FALSE(handle != NULL, ESP_ERR_INVALID_ARG, TAG, "Invalid handle");
+
+    // Clear callback
+    twai_event_callbacks_t cbs = { .on_rx_done = NULL };
+    ret = twai_node_register_event_callbacks(handle->twai_node, &cbs, NULL);
+    ESP_GOTO_ON_ERROR(ret, clean, TAG, "Failed to unregister callbacks: %s", esp_err_to_name(ret));
+
+    // Disable TWAI node
+    ret = twai_node_disable(handle->twai_node);
+    ESP_GOTO_ON_ERROR(ret, clean, TAG, "Failed to disable TWAI node: %s", esp_err_to_name(ret));
```
Dazza0: Might be a better idea to `twai_node_disable()` before deregistering the node's callbacks. It could be very possible that `twai_node_register_event_callbacks()` is called right when a TWAI ISR occurs.  `twai_node_disable()` basically halts the underlying peripheral, thus preventing any further events.

Dazza0 (commented):
@eternal-echo @suda-morris Left some comments (mainly about polling API design), but mostly LGTM otherwise.

esp_isotp/esp_isotp.c
eternal-echo: Yes, I should have diasble first and then set the callback. Thanks for pointing that out

esp_isotp/esp_isotp.c
eternal-echo: Thank you very much for your suggestions, I will fix this thread safety issue right away. My plan is to add an interface to obtain mutex in isotp-c.

The suggestions about esp_timer are also very accurate. I need to study it carefully. If I use this way, I can completely hide the poll part's running startup from the user. Just start a timer in the backend.

esp_isotp/examples/echo/pytest_isotp_echo.py
```diff
@@ -0,0 +1,200 @@
...
+@pytest.mark.parametrize("category", ['basic', 'robustness'])
+def test_echo_basic_and_robustness(dut: Dut, category: str) -> None:
+    """
+    Parameterized test for basic echo functionality and robustness edge cases.
+    """
+    test_cases = ALL_TEST_CASES[category]
+    run_test_batch(dut, test_cases)
+
+@pytest.mark.generic
+@pytest.mark.qemu
```
suda-morris: Please update this file https://github.com/espressif/idf-extra-components/blob/master/.github/setup_qemu.sh#L6 to use the latest release qemu version (https://github.com/espressif/qemu/releases/tag/esp-develop-9.2.2-20250817), which contains the twai peripheral support. @eternal-echo

esp_isotp/docs/src/api.md
```diff
@@ -0,0 +1,183 @@
+# API Reference
+
+<div class="warning">
+
+This file will be automatically generated by esp-doxybook in the future.
+
+Currently maintained manually.
```
suda-morris: This file is automatically generated and overwritten during CI compilation of the document, and we don't need to fill in the content manually here

suda-morris: See it's also empty in other component: https://github.com/espressif/idf-extra-components/blob/master/led_strip/docs/src/api.md

esp_isotp/docs/src/index.md
```diff
@@ -0,0 +1,280 @@
...
+    .tx_id = 0x7E0,           // CAN ID for sending frames
+    .rx_id = 0x7E8,           // CAN ID for receiving frames  
+    .tx_buffer_size = 4096,   // Max message size to send (bytes)
+    .rx_buffer_size = 4096,   // Max message size to receive (bytes)
+    .rx_queue_size = 16,      // TWAI frame buffer depth
+};
+
+// Create ISO-TP transport
+esp_isotp_handle_t isotp_handle;
+ESP_ERROR_CHECK(esp_isotp_new_transport(&config, twai_node, &isotp_handle));
```
suda-morris: Please let the `twai_node` be the first parameter

esp_isotp/docs/src/index.md
```diff
@@ -0,0 +1,280 @@
...
+---
+
+**⚠️ Critical Requirements:**
+- Must call `esp_isotp_poll()` every 1-10ms for proper operation
+- Without regular polling, multi-frame messages will fail
+- All send/receive operations are non-blocking
+
+## How ISO-TP Works
+
+### Single Frame Transmission (≤7 bytes)
```
suda-morris: Better we should mention it, currently twai-fd is not supported yet in this version

esp_isotp/docs/src/index.md
```diff
@@ -0,0 +1,280 @@
...
+- Data is sent immediately in one TWAI frame
+- No flow control required - most efficient for small payloads
+
+### Multi-Frame Transmission (>7 bytes)
+
+For larger messages, ISO-TP automatically segments the data:
+
+#### Step 1: First Frame (FF)
+```
+Sender: [PCI:FF | Length_High | Length_Low | Data[0-5]]
```
suda-morris: question, does this exceed 8 bytes what one twai frame can accommodate?

esp_isotp/docs/src/index.md
```diff
@@ -0,0 +1,280 @@
...
+- **Prevents buffer overflow**: Receiver controls transmission rate
+- **Adapts to processing speed**: STmin allows receiver processing time  
+- **Reliable transmission**: Timeout detection and error handling
+- **Network efficiency**: Optimizes bandwidth usage
+
+## Configuration
+
+### Basic Configuration
+
+The `esp_isotp_config_t` structure controls transport behavior:
```
suda-morris: you can reference the type using markdown anchor, e.g. `[led_strip_new_rmt_device](api.md#function-led_strip_new_rmt_device)`

suda-morris: this way we can provide a clickable link

esp_isotp/docs/src/index.md
eternal-echo: image

I'm sorry I didn't make it clear, this FF and high bit length together make up the first byte, I'm going to see if there is a standard English document, to write well

esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,186 @@
...
+ *     vTaskDelay(pdMS_TO_TICKS(5));
+ * }
+ * ```
+ */
+
+#include "esp_err.h"
+#include "esp_twai.h"
+#include "esp_twai_types.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/queue.h"
```
suda-morris: they're not consumed by the header file, so please include them in the esp_isotp.c instead.

esp_isotp/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Apply patch to esp_isotp-c submodule
+find_package(Git REQUIRED)
```
suda-morris: is it still needed?

esp_isotp/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
...
+
+idf_component_register(
+    SRCS "esp_isotp.c" "isotp-c/isotp.c"
+    INCLUDE_DIRS "." "isotp-c"
+    REQUIRES esp_timer esp_driver_twai
```
suda-morris: esp_timer should be a private require, not public one

esp_isotp/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Key Features
+
+- **Automatic segmentation** for messages >7 bytes with flow control
+- **Non-blocking API** with ISR-based frame processing
+- **Multi-instance support** for concurrent communication channels
+- **Automotive compliance** (UDS, OBD-II compatible)
+- **Robust error handling** with timeout and sequence validation
+
+> **Note**: TWAI-FD is not supported in this version.
```
suda-morris:
```suggestion
> [!NOTE]
> TWAI-FD (Flexible Data-rate)  is not supported in this version.
```

esp_isotp/README.md
```diff
@@ -0,0 +1,83 @@
...
+        vTaskDelay(pdMS_TO_TICKS(5));
+    }
+}
+```
+
+## Examples
+
+| Example | Description |
+|---------|-------------|
+| [echo](examples/echo/) | Basic echo server demonstrating ISO-TP communication |
```
suda-morris: to be added in the next PR.

esp_isotp/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Documentation
+
+- **[Programming Guide](https://espressif.github.io/idf-extra-components/latest/esp_isotp/index.html)** - Complete implementation guide with examples
+- **[Protocol Details](https://espressif.github.io/idf-extra-components/latest/esp_isotp/protocol.html)** - ISO-TP technical specifications
+- **[API Reference](https://espressif.github.io/idf-extra-components/latest/esp_isotp/api.html)** - Function documentation
```
suda-morris: to be added in the next PR

esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,186 @@
...
+/**
+ * @brief Configuration structure for creating a new ISO-TP link
+ */
+struct esp_isotp_config {
+    uint32_t tx_id;                  /*!< TWAI ID for transmitting ISO-TP frames */
+    uint32_t rx_id;                  /*!< TWAI ID for receiving ISO-TP frames */
+    uint32_t tx_buffer_size;         /*!< Size of the transmit buffer (max message size to send) */
+    uint32_t rx_buffer_size;         /*!< Size of the receive buffer (max message size to receive) */
+};
+typedef struct esp_isotp_config esp_isotp_config_t;
```
suda-morris:
```suggestion
typedef struct {
    uint32_t tx_id;                  /*!< TWAI ID for transmitting ISO-TP frames */
    uint32_t rx_id;                  /*!< TWAI ID for receiving ISO-TP frames */
    uint32_t tx_buffer_size;         /*!< Size of the transmit buffer (max message size to send) */
    uint32_t rx_buffer_size;         /*!< Size of the receive buffer (max message size to receive) */
} esp_isotp_config_t;
```

esp_isotp/esp_isotp.h
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * @param handle ISO-TP handle
+ * @param callback Function to call when transmission is complete
+ * @param user_arg User argument to pass to callback
+ * @return
+ *     - ESP_OK: Success
+ *     - ESP_ERR_INVALID_ARG: Invalid arguments
+ */
+esp_err_t esp_isotp_set_tx_done_callback(esp_isotp_handle_t handle,
+        void (*callback)(void *link, uint32_t tx_size, void *user_arg),
```
suda-morris: please typedef the callback function.

suda-morris: void* -> esp_isotp_handle_t

esp_isotp/esp_isotp.c
```diff
@@ -0,0 +1,231 @@
...
+    esp_log_writev(ESP_LOG_DEBUG, "isotp_c", message, args);
+    va_end(args);
+}
+
+esp_err_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_config_t *config, esp_isotp_handle_t *out_handle)
+{
+    esp_err_t ret = ESP_OK;
+    ESP_RETURN_ON_FALSE(config != NULL, ESP_ERR_INVALID_ARG, TAG, "Invalid configuration");
+    ESP_RETURN_ON_FALSE(twai_node != NULL, ESP_ERR_INVALID_ARG, TAG, "Invalid TWAI node");
+    ESP_RETURN_ON_FALSE(out_handle != NULL, ESP_ERR_INVALID_ARG, TAG, "Invalid output handle");
```
suda-morris: merge them

esp_isotp/esp_isotp.c
```diff
@@ -0,0 +1,231 @@
...
+            isotp_on_can_message(&link_handle->link, frame_data, rx_frame.buffer_len);
+        }
+    }
+    return false;
+}
+
+uint32_t isotp_user_get_us(void)
+{
+    return (uint32_t)esp_timer_get_time();
+}
```
suda-morris:
```suggestion
}

```

suda-morris (approved):

PR #536, opened 2025-08-11
fix(esp_gcov): rename Kconfig.gcov as Kconfig

erhankur:
Kconfig.gcov is renamed as Gcov options will be under component's own section instead of app_trace menuconfig.


erhankur:
@suda-morris PTAL

suda-morris (approved):
Please bump the version

PR #541, opened 2025-08-19
Feat/add esp CLI component

SoucheSouche:
# Checklist

- [ ] CI passing

# Change description


esp_cli/src/esp_cli.c
```diff
@@ -0,0 +1,214 @@
...
+            (config->max_cmd_line_size == 0)) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    esp_repl_instance_t *instance = malloc(sizeof(esp_repl_instance_t));
+    if (!instance) {
+        return ESP_ERR_NO_MEM;
+    }
+
+    instance->config = *config;
```
o-marshmallow, line 79: We must check that `linenoise_handle` is not NULL since it's not optional

esp_repl/esp_repl.c
```diff
@@ -0,0 +1,214 @@
...
+
+    /* update the state to force the while loop in esp_repl to return */
+    state->state = ESP_REPL_STATE_STOPPED;
+
+    /* Call the abort function from esp_linenoise to force esp_linenoise_get_line to return.
+    * This function is expected to return ESP_OK only if the user has registered a custom
+    * read to the esp_linenoise instance and if the abort function succeeds.
+    * ESP_ERR_INVALID_STATE is expected to be returned by esp_linenoise_abort if it is called
+    * when the user has registered a custom read to the esp_linenoise instance. From the point
+    * of view of esp_repl_stop, this return value if indicating that the user will have to take
```
o-marshmallow:
```suggestion
    * of view of esp_repl_stop, this return value is indicating that the user will have to take
```

o-marshmallow: Moreover, this comment looks important to me, should it also be mentioned in the public header too?

esp_repl/esp_repl.c
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /* Call the on_stop callback to let the user unblock esp_linenoise
+    * if a custom read is provided */
+    if (config->on_stop.func != NULL) {
+        config->on_stop.func(config->on_stop.ctx, handle);
+    }
+
+    /* Wait for esp_repl() to finish and signal completion */
+    xSemaphoreTake(state->mux, portMAX_DELAY);
```
o-marshmallow: This means that we cannot call `esp_repl_stop` from the same Task that calls `esp_repl` (for example if there is an `exit` command that it meant to kill the REPL). This should be stated in the public header.

esp_repl/esp_repl.c
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        /* at this point, if the command is NULL, skip the executing part */
+        if (read_ret != ESP_OK) {
+            continue;
+        }
+
+        /* try to run the command */
+        int cmd_func_ret;
+        const esp_err_t exec_ret = esp_commands_execute(c_set, -1, cmd_line, &cmd_func_ret);
```
o-marshmallow: Is it safe if `c_set` is NULL? It was not tested in the `init` function

esp_repl/include/esp_repl.h
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * @brief Function prototype called before executing a command.
+ *
+ * @param ctx User-defined context pointer.
+ * @param buf Buffer containing the command.
+ * @param reader_ret_val Return value from the reader function.
+ *
+ * @return ESP_OK to continue execution, error code to abort.
+ */
+typedef esp_err_t (*esp_repl_pre_executor_fn)(void *ctx, char *buf, const esp_err_t reader_ret_val);
```
o-marshmallow: The `const` on `esp_err_t` here isn't necessary since it will be passed by copy anyway (`int`).
However, a `const` on `buf` seems necessary to me to prevent having the executor modify the command, except if that's is a feature/expected behavior.

esp_repl/include/esp_repl.h
```diff
@@ -0,0 +1,156 @@
...
+ * @brief Function prototype called after executing a command.
+ *
+ * @param ctx User-defined context pointer.
+ * @param buf Command that was executed.
+ * @param executor_ret_val Return value from the executor function.
+ * @param cmd_ret_val Command-specific return value.
+ *
+ * @return ESP_OK on success, error code otherwise.
+ */
+typedef esp_err_t (*esp_repl_post_executor_fn)(void *ctx, const char *buf, const esp_err_t executor_ret_val, const int cmd_ret_val);
```
o-marshmallow: Same remarks for `const esp_err_t` and `const int`

esp_repl/include/esp_repl.h
```diff
@@ -0,0 +1,156 @@
...
+    void *ctx;                /**!< Context passed to the exit function */
+} esp_repl_on_exit_t;
+
+/**
+ * @brief Configuration structure to initialize a REPL instance.
+ */
+typedef struct esp_repl_config {
+    esp_linenoise_handle_t linenoise_handle;    /**!< Handle to the esp_linenoise instance */
+    esp_command_set_handle_t command_set_handle;   /**!< Handle to a set of commands */
+    size_t max_cmd_line_size;                   /**!< Maximum allowed command line size */
```
o-marshmallow: Does it include the NULL-byte terminator?

esp_repl/esp_repl.c
```diff
@@ -0,0 +1,214 @@
...
+    ESP_REPL_STATE_STOPPED
+} esp_repl_state_e;
+
+typedef struct esp_repl_state {
+    esp_repl_state_e state;
+    SemaphoreHandle_t mux;
+} esp_repl_state_t;
+
+typedef struct esp_repl_instance {
+    esp_repl_handle_t self;
```
o-marshmallow: I am not sure to understand the purpose of this field, I see that it is used to check the instance, but does it only serves the purpose of integrity check? 
In which case it would be possible that `self` is different than the pointer within the API?

esp_cli/src/esp_cli.c
SoucheSouche: it is already checked at the beginning of the function

esp_repl/esp_repl.c
SoucheSouche: I updated the comment so it should be clearer hopefully. i also added a note in the doxygen comment of the function.

esp_repl/esp_repl.c
SoucheSouche: yes, NULL command set is interpreted as no command set by the esp_commands component.

esp_repl/esp_repl.c
SoucheSouche: I removed the field. It is doesn't bring much to the sanity check and uses unnecessary space.

esp_repl/esp_repl.c
SoucheSouche: I updated the logic to skip the mutex taking if esp_repl_stop is called from the task running esp_repl. I added a simple test to make sure esp_repl does return properly in this scenario.

esp_repl/esp_repl.c
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        /* at this point, if the command is NULL, skip the executing part */
+        if (read_ret != ESP_OK) {
+            continue;
+        }
+
+#if CONFIG_ESP_REPL_HAS_QUIT_CMD
+        /* evaluate the command name */
+        if (strncmp("quit", cmd_line, 4) == 0) {
```
o-marshmallow: You can merge both `if` statements:
```
        if (strncmp("quit", cmd_line, 4) == 0 && esp_repl_stop(handle) == ESP_OK) { ...
```

esp_repl/README.md
```diff
@@ -0,0 +1,118 @@
...
+    vTaskDelete(NULL);
+}
+
+void app_main(void)
+{
+    esp_err_t ret;
+    esp_repl_handle_t repl = NULL;
+
+    // Initialize esp_linenoise (mandatory)
+    esp_linenoise_handle_t esp_linenoise_hdl = esp_linenoise_create();
```
ESP-Marius: if using linenoise is mandatory, but the setup does not configure anything: what is the benefit of making the user init and passing the handle vs just doing it internally inside repl_create?

esp_repl/idf_component.yml
```diff
@@ -0,0 +1,14 @@
+version: "1.0.0"
+description: "esp_repl - Read Eval Print Loop component"
+url: https://github.com/espressif/idf-extra-components/tree/master/esp_repl
+dependencies:
+  SoucheSouche/esp_linenoise:
+    version: "*"
```
ESP-Marius: For the release do you want to limit the versions here?

esp_repl/esp_repl.c
SoucheSouche: I will leave it as it is because it increases readability this way.

PR #546, opened 2025-08-25
refactor(tjpgd): simplify the example to decouple it from bsp packages

suda-morris:
This PR is to fix a build issue we encounter in https://github.com/espressif/idf-extra-components/actions/runs/17204187200/job/48801105823

By decoupling the example from the BSP packages.

### DIFF
--- a/esp_jpeg/examples/get_started/README.md
+++ b/esp_jpeg/examples/get_started/README.md
@@ -1,8 +1,10 @@
 # LCD tjpgd example
 
+## Overview
+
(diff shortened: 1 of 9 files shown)


mahavirj:
Shall we bump up the component version?

suda-morris:
> Shall we bump up the component version?

I want to wait some time to collect more feature request or bugfix. Actually we're here not fixing a bug in the component itself. Thanks for the remind.

PR #547, opened 2025-08-26
Fix/upgrade idf build apps

suda-morris:
Use the latest idf-build-apps.


.github/workflows/build_and_run_apps.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           git config --global --add safe.directory $build_dir
           cd -
       - name: Install dependencies
-        run: pip install 'idf-build-apps>=2.4,<2.5'
+        run: pip install 'idf-build-apps~=2.12'
```
tore-espressif, line 37: Same is on line 92, then it should work

suda-morris: good catch

PR #548, opened 2025-08-26
fix(jpeg-turbo): version 3.1.1 && don't build cli tools

suda-morris:
This PR aims to remove the build warnings e.g.: https://github.com/espressif/idf-extra-components/actions/runs/17233944936/job/48894577947

Because it is also trying to build the CLI tools and tests, which is not necessary for us.

We're using the commit a little bit ahead of the upstream tag 3.1.1 because we want to include this commit:
https://github.com/libjpeg-turbo/libjpeg-turbo/commit/942ac87e479c48199d48781e9a5b2cda3b910d7b

jpeg-turbo is a pure C project, so removed the compile flags that are only for C++: `-fpermissive`


igrr (approved):
Thanks for the fix and the cleanups!

PR #563, opened 2025-09-05
feat(isotp): Add thread safety, callbacks, and extended ID support (IEC-370)

eternal-echo:
## Overview

This PR enhances the `esp_isotp` component with thread safety, event-driven callbacks, and extended feature support, transforming it from a basic polling implementation to a production-ready ISO-TP communication library.

### Thread Safety & ISR Support
- **ISR-safe send functions**: `esp_isotp_send()` and `esp_isotp_send_with_id()` can be called from interrupt context
- **Spinlock protection**: Core state protected with `portMUX` locks to prevent race conditions
- **TX frame buffer pool**: Pre-allocated memory pool eliminates stack corruption in async TWAI operations

### Event-Driven Callbacks  
- **Asynchronous notifications**: Optional `rx_callback` and `tx_callback` in configuration
- **Non-blocking operation**: Eliminates need for continuous polling in callback mode

### Extended Features
- **29-bit ID support**: Added `use_extended_id` configuration option
- **Dynamic TX ID**: New `esp_isotp_send_with_id()` function for runtime ID override
- **Enhanced error handling**: More specific `esp_err_t` return codes

### New Configuration Options
```c
esp_isotp_config_t config = {
    .use_extended_id = false,
    .tx_frame_pool_size = 8,
    .rx_callback = my_rx_callback,
    .tx_callback = my_tx_callback,
    .callback_arg = user_data
};
```

### New Functions
- `esp_err_t esp_isotp_send_with_id(esp_isotp_handle_t handle, uint32_t id, const uint8_t *data, uint32_t size)`

## Breaking Changes

None. This enhancement maintains full backward compatibility.

## Files Modified

- `src/esp_isotp.c`: Core implementation with memory management and thread safety
- `inc/esp_isotp.h`: API enhancements and documentation
- `README.md`: Updated documentation with usage examples

## Notes

- `esp_isotp_poll()` must still be called periodically for multi-frame message handling
- RX callbacks execute in ISR context - keep processing minimal
- TX frame pool prevents memory corruption in async operations

## Checklist

- [x] Component contains unit tests (pytest)

### DIFF
--- a/esp_isotp/inc/esp_isotp.h
+++ b/esp_isotp/inc/esp_isotp.h
@@ -34,14 +34,52 @@ extern "C" {
 typedef struct esp_isotp_link_t *esp_isotp_handle_t;
 
 /**
+ * @brief ISO-TP receive callback function type
+ *
+ * Called when a complete message has been received successfully.
+ *
+ * @warning This callback executes in the same context as isotp_on_can_message(),
+ *          which is typically ISR context. Keep callback execution time minimal
+ *          and avoid blocking operations.
+ * @note The data pointer is only valid during the callback execution.
+ *       Copy data immediately if needed beyond the callback scope.
+ *
+ * @param[in] handle ISO-TP handle that received the message
+ * @param[in] data Pointer to received data (valid only during callback)
+ * @param[in] size Size of received data in bytes
+ * @param[in] user_arg User argument provided during configuration
+ */
+typedef void (*esp_isotp_rx_callback_t)(esp_isotp_handle_t handle, const uint8_t *data, uint32_t size, void *user_arg);
+
+/**
+ * @brief ISO-TP transmit callback function type
+ *
+ * Called when a complete message has been transmitted successfully.
+ *
+ * @note Execution context depends on message type:
+ *       - Single-frame: Called immediately from isotp_send() in caller's context
+ *       - Multi-frame: Called from isotp_poll() in task context
+ * @note Keep callback execution time minimal to avoid affecting system performance.
+ *
+ * @param[in] handle ISO-TP handle that transmitted the message
+ * @param[in] tx_size Size of transmitted data in bytes
+ * @param[in] user_arg User argument provided during configuration
+ */
+typedef void (*esp_isotp_tx_callback_t)(esp_isotp_handle_t handle, uint32_t tx_size, void *user_arg);
+
+/**
  * @brief Configuration structure for creating a new ISO-TP link
  */
 typedef struct {
-    uint32_t tx_id;                  /*!< TWAI ID for transmitting ISO-TP frames (11-bit or 29-bit) */
-    uint32_t rx_id;                  /*!< TWAI ID for receiving ISO-TP frames (11-bit or 29-bit) */
+    uint32_t tx_id;                  /*!< TWAI ID for transmitting ISO-TP frames (11-bit or 29-bit, auto-detected from value) */
+    uint32_t rx_id;                  /*!< TWAI ID for receiving ISO-TP frames (11-bit or 29-bit, auto-detected from value) */
     uint32_t tx_buffer_size;         /*!< Size of the transmit buffer (max message size to send) */
     uint32_t rx_buffer_size;         /*!< Size of the receive buffer (max message size to receive) */
-    bool use_extended_id;            /*!< true: use 29-bit extended ID, false: use 11-bit standard ID */
+    uint32_t tx_frame_pool_size;     /*!< Size of TX frame pool */
+
+    esp_isotp_rx_callback_t rx_callback; /*!< Receive completion callback (NULL for polling mode) */
+    esp_isotp_tx_callback_t tx_callback; /*!< Transmit completion callback (NULL to disable) */
+    void *callback_arg;               /*!< User argument passed to callbacks */
 } esp_isotp_config_t;
 
 /**
@@ -63,10 +101,15 @@ typedef struct {
 esp_err_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_config_t *config, esp_isotp_handle_t *out_handle);
 
 /**
- * @brief Send data over an ISO-TP link (non-blocking)
+ * @brief Send data over an ISO-TP link (non-blocking, ISR-safe)
  *
  * Immediately sends first/single frame and returns. For multi-frame messages,
  * remaining frames are sent during subsequent esp_isotp_poll() calls.
+ *
+ * @note This function is ISR-safe and can be called from interrupt context.
+ * @note TX completion callback timing:
+ *       - Single-frame: Called immediately from isotp_send()
+ *       - Multi-frame: Called from isotp_poll() when last frame is sent
  *
  * @param handle ISO-TP handle
  * @param data Data to send
@@ -83,11 +126,16 @@ esp_err_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_
 esp_err_t esp_isotp_send(esp_isotp_handle_t handle, const uint8_t *data, uint32_t size);
 
 /**
- * @brief Send data over an ISO-TP link with specified TWAI ID (non-blocking)
+ * @brief Send data over an ISO-TP link with specified TWAI ID (non-blocking, ISR-safe)
  *
  * Similar to esp_isotp_send(), but allows specifying a different TWAI ID for transmission.
  * This function is primarily used for functional addressing where multiple nodes
  * may respond to the same request.
+ *
+ * @note This function is ISR-safe and can be called from interrupt context.
+ * @note TX completion callback timing:
+ *       - Single-frame: Called immediately from isotp_send_with_id()
+ *       - Multi-frame: Called from isotp_poll() when last frame is sent
  *
  * @param handle ISO-TP handle
  * @param id TWAI identifier to use for transmission (overrides configured tx_id)
@@ -99,18 +147,23 @@ esp_err_t esp_isotp_send(esp_isotp_handle_t handle, const uint8_t *data, uint32_
  *     - ESP_ERR_NO_MEM: Data too large for buffer or no space available
  *     - ESP_ERR_INVALID_SIZE: Invalid data size
  *     - ESP_ERR_TIMEOUT: Send operation timed out
- *     - ESP_ERR_INVALID_ARG: Invalid parameters
+ *     - ESP_ERR_INVALID_ARG: Invalid parameters or ID exceeds maximum value
  *     - ESP_FAIL: Other send errors
  */
 esp_err_t esp_isotp_send_with_id(esp_isotp_handle_t handle, uint32_t id, const uint8_t *data, uint32_t size);
 
 /**
- * @brief Extract a complete received message (non-blocking)
+ * @brief Extract a complete received message (non-blocking, task context only)
  *
  * This function only extracts data that has already been assembled by esp_isotp_poll().
  * It does NOT process incoming TWAI frames - that happens in esp_isotp_poll().
  *
  * Process: TWAI frames → esp_isotp_poll() assembles → esp_isotp_receive() extracts
+ *
+ * @warning This function is NOT ISR-safe and must only be called from task context.
+ * @note When using callback mode (rx_callback != NULL), received messages are
+ *       automatically delivered via callback. This function should only be used
+ *       in polling mode (rx_callback == NULL).
  *
  * @param handle ISO-TP handle
  * @param data Buffer to store received data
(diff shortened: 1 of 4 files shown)


esp_isotp/isotp-c/isotp_user.h
```diff
@@ -31,7 +31,10 @@ int isotp_user_send_can(const uint32_t arbitration_id, const uint8_t *data, cons
                        );
 
 /**
- * @brief user implemented, gets the amount of time passed since the last call in microseconds
+ * @brief User-implemented: return current monotonic time in microseconds.
+ *
+ * Must return a 32-bit timestamp in microseconds that monotonically increases
+ * (wrapping is expected). The library uses IsoTpTimeAfter() to handle wrap.
```
suda-morris: this header file is going to be under a git submodule: https://github.com/espressif/idf-extra-components/pull/564

So we can't modify it directly. Keep the correct comment in esp_isotp.c is sufficient.

esp_isotp/inc/esp_isotp.h
```diff
@@ -57,63 +128,118 @@ typedef struct esp_isotp_link_t *esp_isotp_handle_t;
...
  *     - ESP_OK: Send initiated successfully
  *     - ESP_ERR_NOT_FINISHED: Previous send still in progress
- *     - ESP_ERR_NO_MEM: Data too large for buffer
+ *     - ESP_ERR_NO_MEM: Data too large for buffer or no space available
+ *     - ESP_ERR_INVALID_SIZE: Invalid data size
+ *     - ESP_ERR_TIMEOUT: Send operation timed out
  *     - ESP_ERR_INVALID_ARG: Invalid parameters
+ *     - ESP_FAIL: Other send errors
  */
 esp_err_t esp_isotp_send(esp_isotp_handle_t handle, const uint8_t *data, uint32_t size);
```
suda-morris, line 126: I guess there's no need to provide this function, as user can always call `esp_isotp_send_with_id`.

esp_isotp/inc/esp_isotp.h
```diff
@@ -30,14 +39,44 @@
...
+ * };
+ *
+ * esp_isotp_handle_t handle;
+ * esp_isotp_new_transport(twai_node, &config, &handle);
+ *
+ * while (1) {
+ *     esp_isotp_poll(handle);  // MUST still call regularly!
+ *     vTaskDelay(pdMS_TO_TICKS(5));
+ * }
+ * ```
```
suda-morris: I wonder if our doc build system will handle this part nicely. If not, I would suggest to put the example code in the programming guide.

esp_isotp/inc/esp_isotp.h
```diff
@@ -57,63 +128,118 @@ typedef struct esp_isotp_link_t *esp_isotp_handle_t;
...
  */
 typedef struct {
-    uint32_t tx_id;                  /*!< TWAI ID for transmitting ISO-TP frames */
-    uint32_t rx_id;                  /*!< TWAI ID for receiving ISO-TP frames */
+    uint32_t tx_id;                  /*!< TWAI ID for transmitting ISO-TP frames (11-bit or 29-bit) */
+    uint32_t rx_id;                  /*!< TWAI ID for receiving ISO-TP frames (11-bit or 29-bit) */
     uint32_t tx_buffer_size;         /*!< Size of the transmit buffer (max message size to send) */
     uint32_t rx_buffer_size;         /*!< Size of the receive buffer (max message size to receive) */
+    uint32_t tx_frame_pool_size;     /*!< Size of TX frame pool (0 for default: 8) */
+    bool use_extended_id;            /*!< true: use 29-bit extended ID, false: use 11-bit standard ID */
```
suda-morris: This configuration seems to be at the transaction-level. We should allow the user to use the same isotp handle to transmit both 11-bit and 29-bit IDs frames.

esp_isotp/inc/esp_isotp.h
```diff
@@ -48,6 +87,38 @@
...
+ *          which is typically ISR context. Keep callback execution time minimal
+ *          and avoid blocking operations.
+ * @note The data pointer is only valid during the callback execution.
+ *       Copy data immediately if needed beyond the callback scope.
+ *
+ * @param[in] data Pointer to received data (valid only during callback)
+ * @param[in] size Size of received data in bytes
+ * @param[in] user_arg User argument provided during configuration
+ */
+typedef void (*esp_isotp_rx_callback_t)(const uint8_t *data, uint32_t size, void *user_arg);
```
suda-morris: question, are these callbacks also be called in the `esp_isotp_poll`?

esp_isotp/inc/esp_isotp.h
eternal-echo: I was thinking it might still be useful to keep `esp_isotp_send()`. In many ISO-TP implementations, the usual practice is to configure CAN IDs statically and reuse them, so having a function without the ID parameter could be more convenient for common use cases.
>
> * **Linux ISO-TP**: you create a socket with `socket(PF_CAN, SOCK_DGRAM, CAN_ISOTP)`, then `bind()` it once with `tx_id` and `rx_id`. All subsequent `write()` / `read()` calls reuse these IDs, so there’s no need to pass the CAN ID each time ([docs.kernel.org](https://docs.kernel.org/networking/can.html)).
> * **AUTOSAR CanTp**: CAN IDs are defined statically in the configuration (mapping N-SDUs to specific CAN IDs). Application code just calls `CanTp_Transmit()`, without re-specifying the ID on every transmit ([medium article](https://abhishek-anand.medium.com/cantp-transport-protocol-for-can-communication-in-autosar-bsw-812786022240)).

So it seems that a function like `esp_isotp_send(handle, data, size)`—which reuses the ID bound to the handle—would align with established practice and be convenient for the common case. At the same time, `esp_isotp_send_with_id()` could stay as an advanced API for cases where dynamic CAN IDs are explicitly required.

esp_isotp/inc/esp_isotp.h
eternal-echo: No, this callback will be parsed and invoked immediately after the corresponding CAN frame is received. `poll` does not parse the received data — only `isotp_on_can_message` parses the received CAN frame and triggers the callback.

esp_isotp/src/esp_isotp.c
```diff
@@ -136,64 +367,213 @@ esp_err_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_
...
+    esp_err_t twai_ret = twai_node_disable(handle->twai_node);
+    if (twai_ret != ESP_OK) {
+        ESP_LOGW(TAG, "Failed to disable TWAI node: %s", esp_err_to_name(twai_ret));
+        if (ret == ESP_OK) {
+            ret = twai_ret;
+        }
+    }
+
+    // Unregister TWAI callbacks first to prevent use-after-free during disable
+    twai_event_callbacks_t cbs = { 0 };
```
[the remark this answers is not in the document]
eternal-echo: We should disable first.

esp_isotp/inc/esp_isotp.h
eternal-echo: Now is auto detect.

esp_isotp/src/esp_isotp.c
```diff
@@ -4,88 +4,280 @@
...
 /**
- * @brief ISO-TP link structure
+ * @brief Determine if the given ID requires extended (29-bit) format
+ *
+ * @param id TWAI identifier to check
+ * @return true if ID requires 29-bit extended format, false for 11-bit standard format
+ */
+static inline bool is_extended_id(uint32_t id)
+{
+    return (id > TWAI_STD_ID_MASK);
```
suda-morris, line 30: We can't stop users from using 29-digit IDs to express small numbers like 0x100, although 11 digits is enough

esp_isotp/inc/esp_isotp.h
suda-morris: It's hard for the user to know which isotp link invokes the callback, becuase there's no "ISOTP link handle" in the callback function parameter.

esp_isotp/src/esp_isotp.c
```diff
@@ -4,88 +4,280 @@
...
+ */
+int IRAM_ATTR isotp_user_send_can(const uint32_t arbitration_id, const uint8_t *data, const uint8_t size
 #ifdef ISO_TP_USER_SEND_CAN_ARG
-                        , void *user_data
+                                  , void *user_data
 #endif
-                       )
+                                 )
 {
 #ifdef ISO_TP_USER_SEND_CAN_ARG
```
suda-morris: it's always enabled in our port.

esp_isotp/src/esp_isotp.c
```diff
@@ -99,34 +291,87 @@ esp_err_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_
...
+        frame->frame.buffer_len = sizeof(frame->data_payload);
 
-    // Initialize ISO-TP link
-    isotp_init_link(&isotp->link, config->tx_id, isotp->tx_buffer,
-                    config->tx_buffer_size, isotp->rx_buffer, config->rx_buffer_size);
+        SLIST_INSERT_HEAD(&isotp->tx_frame_pool, frame, link);
+    }
+
+    // Initialize the isotp-c library link with our allocated buffers.
+    isotp_init_link(&isotp->link, config->tx_id, isotp->isotp_tx_buffer,
```
suda-morris, line 304: seems the rx_id is not needed to initialize the ISO TP link. Not sure if we really need it in the esp_isotp_config_t.

Can we handle different receive frames with different ID with the same ISOTP link?

esp_isotp/src/esp_isotp.c
```diff
@@ -136,64 +381,214 @@ esp_err_t esp_isotp_new_transport(twai_node_handle_t twai_node, const esp_isotp_
...
+ *  - ESP_ERR_NOT_FINISHED when the send is still in progress
+ *  - ESP_ERR_NO_MEM for buffer overflow conditions
+ *  - ESP_ERR_INVALID_SIZE for invalid sizes
+ *  - ESP_ERR_TIMEOUT on timeout
+ *  - ESP_FAIL for other errors
+ */
 esp_err_t esp_isotp_send(esp_isotp_handle_t handle, const uint8_t *data, uint32_t size)
 {
-    ESP_RETURN_ON_FALSE(handle && data && size, ESP_ERR_INVALID_ARG, TAG, "Invalid parameters");
+    ESP_RETURN_ON_FALSE_ISR(handle && data && size, ESP_ERR_INVALID_ARG, TAG, "Invalid parameters");
```
suda-morris: if this function can also be called in ISR, we can skip the error print but just return the error code.

esp_isotp/src/esp_isotp.c
eternal-echo: Great observation! Let me clarify the design:

## Why rx_id isn't in isotp_init_link?
The isotp-c library is a platform-agnostic state machine that only handles ISO-TP protocol logic, not underlying CAN hardware.

## Workflow:
1. isotp_init_link only needs tx_id for sending setup
2. ESP layer receives all CAN frames and filters by ID
3. Only frames matching rx_id are fed to isotp_on_can_message()

## Can one link handle multiple rx_ids?
No. An ISO-TP link represents a point-to-point connection with state machine tied to a specific peer. Multiple IDs would corrupt the state.

The correct approach is creating separate link instances for each communication channel.

## Conclusion:
1. rx_id in esp_isotp_config_t is essential for ESP-layer filtering
2. Multiple rx_id support requires multiple independent isotp_link instances

esp_isotp/inc/esp_isotp.h
eternal-echo: Thanks for the feedback!
The user_arg parameter is intended to address this — the user can pass the ISOTP link handle (or any custom context) when configuring the link, and the same pointer will be passed back to the callback. This way the callback can easily identify which link triggered the event.

I'll update the documentation to explicitly mention that user_arg can be used for this purpose.

suda-morris: the user_arg is used to pass other context, except the isotp handle because the isotp handle is must.

eternal-echo: Now, it's added

esp_isotp/src/esp_isotp.c
eternal-echo: Yes, maybe we can add a tx_addr like [isotp-zephyr](https://docs.zephyrproject.org/latest/doxygen/html/group__can__isotp.html#ga3723fae583a1d7deabc2deee475ba756).

esp_isotp/src/esp_isotp.c
```diff
@@ -4,76 +4,252 @@
...
+    // Remove frame from pool
+    SLIST_REMOVE_HEAD(&isotp_handle->tx_frame_pool, link);
+
+    // Initialize TWAI frame header and copy payload data into embedded buffer.
+    memset(&tx_frame->frame, 0, sizeof(twai_frame_t));
+    tx_frame->frame.header.id = arbitration_id;
+    tx_frame->frame.header.ide = is_extended_id(arbitration_id);  // Extended (29-bit) vs Standard (11-bit) ID
+
+    // Size validation - TWAI frames are max 8 bytes by protocol
+    ESP_RETURN_ON_FALSE_ISR(size <= 8, ISOTP_RET_ERROR, TAG, "Invalid TWAI frame size");
```
suda-morris, line 227: this check should be moved to the begging of the function. @eternal-echo

PR #579, opened 2025-09-25
feat(flash_dispatcher): Add component for flash dispatcher

mythbuster5:
# Checklist

- [ ] CI passing

# Change description


esp_flash_dispatcher/idf_component.yml
```diff
@@ -0,0 +1,7 @@
+version: "1.0.0"
+description: ESP flash driver dispatcher for working on spiram stack
+url: https://github.com/espressif/idf-extra-components/tree/master/esp_flash_dispatcher
+repository: https://github.com/espressif/idf-extra-components.git
+issues: https://github.com/espressif/idf-extra-components/issues
+dependencies:
+  idf: ">=6.0"
```
suda-morris: can't it work with 5.x release?

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+ESP_SYSTEM_INIT_FN(esp_flash_dispatcher_init, SECONDARY, BIT(0), 230)
+{
+    const esp_flash_dispatcher_config_t cfg = {
+        .task_stack_size = FLASH_DISPATCHER_TASK_STACK_SIZE,
+        .task_priority = FLASH_DISPATCHER_TASK_PRIORITY,
+        .task_core_id = tskNO_AFFINITY,
+        .queue_size = FLASH_DISPATCHER_TASK_QUEUE_SIZE,
+    };
+
+    s_flash_dispatcher_ctx.queue = xQueueCreate(cfg.queue_size, sizeof(flash_operation_request_t));
```
suda-morris: queue should also created with caps, otherwise it may be allocated from PSRAM

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+    ESP_RETURN_ON_FALSE(s_flash_dispatcher_ctx.queue, ESP_ERR_NO_MEM, TAG, "create flash operation queue failed");
+
+    uint8_t *stack_for_task = heap_caps_malloc(cfg.task_stack_size, MALLOC_CAP_INTERNAL);
+    ESP_RETURN_ON_FALSE(stack_for_task, ESP_ERR_NO_MEM, TAG, "no memory for flash stask");
+
+    static StaticTask_t task_buffer;
+    s_flash_dispatcher_ctx.task = xTaskCreateStaticPinnedToCore(
```
suda-morris: likewise, use "WithCaps" API

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+    flash_operation_request_t request;
+
+    while (true) {
+        if (xQueueReceive(s_flash_dispatcher_ctx.queue, &request, portMAX_DELAY) == pdTRUE) {
+            // Execute the actual flash operation based on the function pointer and arguments
+            // Note: This is a simplified example. A more robust solution would involve a switch-case
+            // or a more sophisticated argument passing mechanism to handle different esp_flash APIs.
+            if (request.operation_func) {
+                if (request.operation_func == (esp_err_t (*)(esp_flash_t *, ...))__real_esp_flash_read) {
+                    request.result = __real_esp_flash_read(request.chip, (void *)request.arg1, (uint32_t)(void *)request.arg2, (size_t)request.arg3);
```
igrr: Question 1, what's the point of passing the function pointer inside flash_operation_request_t, if we are not calling it? Would be be more readable to make operation_func an enum, then the code would look like
```c++
if (requres.operation_func == OP_READ) {
```
instead of the scary-looking type cast.

Question 2: why does the comment above say that a switch-case would be a more robust solution? What's more robust about a switch-case?

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+            }
+            // Notify the calling task about completion
+            if (request.calling_task_handle) {
+                xTaskNotify(request.calling_task_handle, request.result, eSetValueWithOverwrite);
+            }
+        }
+    }
+}
+
+ESP_SYSTEM_INIT_FN(esp_flash_dispatcher_init, SECONDARY, BIT(0), 998)
```
igrr: I would not recommend using ESP_SYSTEM_INIT_FN from components outside of IDF. It's a private API. Your code will break when IDF changes this API.

I think it's enough to make `esp_flash_dispatcher_init(const esp_flash_dispatcher_config_t* cfg)` public.

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+esp_err_t __wrap_esp_flash_read(esp_flash_t *chip, void *buffer, uint32_t address, uint32_t size)
+{
+    flash_operation_request_t request = {
+        .chip = chip,
+        .arg1 = (void *)buffer,
+        .arg2 = (void *)address,
+        .arg3 = (void *)size,
+        .operation_func = (esp_err_t (*)(esp_flash_t *, ...))__real_esp_flash_read,
+        .calling_task_handle = xTaskGetCurrentTaskHandle(),
+        .result = ESP_FAIL // Initialize with failure
```
igrr: Trivial comment not really helping?

esp_flash_dispatcher/esp_flash_dispatcher.c
mythbuster5: This queue is ok, It can be allocated from PSRAM. We want to prevent the cache might be disabled inside the flash operation. This queue is always used outside flash operations.

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+        ESP_EARLY_LOGE(TAG, "Failed to send flash read request to queue");
+        return ESP_ERR_TIMEOUT;
+    }
+
+    // Wait for the dispatcher task to complete the operation and notify us
+    uint32_t notification_value;
+    if (xTaskNotifyWait(0, 0xFFFFFFFF, &notification_value, portMAX_DELAY) == pdTRUE) {
+        return (esp_err_t)notification_value;
+    } else {
+        ESP_EARLY_LOGE(TAG, "Failed to receive notification for flash read operation");
```
igrr: In which case will we hit this branch?

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,248 @@
...
+        return ESP_ERR_TIMEOUT;
+    }
+
+    uint32_t notification_value;
+    if (xTaskNotifyWait(0, 0xFFFFFFFF, &notification_value, portMAX_DELAY) == pdTRUE) {
+        return (esp_err_t)notification_value;
+    } else {
+        ESP_EARLY_LOGE(TAG, "Failed to receive notification for flash write operation");
+        return ESP_ERR_INVALID_RESPONSE;
+    }
```
igrr, line 214: Can we factor out this repeated piece of code into a function: Send request and wait for result. Then the only thing that's different about the wrappers is the request structure.

mythbuster5:
hi

esp_flash_dispatcher/idf_component.yml
```diff
@@ -0,0 +1,7 @@
+version: "1.0.0"
+description: ESP flash driver dispatcher for working on spiram stack
```
igrr:
```suggestion
description: This component allows flash operations to be performed by tasks with stacks in PSRAM.
```

esp_flash_dispatcher/test_apps/flash_dispatcher/main/test_app_main.c
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "unity.h"
+#include "unity_test_utils.h"
+#include "esp_heap_caps.h"
+
+// Some resources are lazy allocated in esp_flash test, especially ccomp timer test, the threshold is left for that case
```
igrr: The test app doesn't seem to use ccomp timer.

esp_flash_dispatcher/test_apps/main/test_esp_flash_dispatcher.c
```diff
@@ -0,0 +1,97 @@
...
+#include "esp_flash.h"
+#include <esp_attr.h>
+#include "esp_log.h"
+#include "unity.h"
+#include "sdkconfig.h"
+
+// Buffer for PSRAM task stack
+static StackType_t *s_psram_task_stack = NULL;
+
+#define TEST_FLASH_ADDRESS 0x110000 // default address in flash for testing
```
igrr, line 25: It's probably safer to create a partition and get the address of that partition?

esp_flash_dispatcher/.build-test-rules.yml
```diff
@@ -0,0 +1,4 @@
+esp_flash_dispatcher/test_apps/flash_dispatcher:
+  disable:
+    - if: SOC_SPIRAM_SUPPORTED != 1
+      reason: "This feature only work on chips which have psram"
```
igrr:
```suggestion
      reason: "This feature is only necessary on chips which have PSRAM"
```

esp_flash_dispatcher/Kconfig
```diff
@@ -0,0 +1,17 @@
+config ESP_FLASH_DISPATCHER_TASK_STACK_SIZE
```
igrr: Shouldn't the options of a component be under a `menu`?

esp_flash_dispatcher/esp_flash_dispatcher.c
mythbuster5: I used create static memory for the stack, it's more controllable?

suda-morris: the WithCaps APIs are help functions, doing similar job

esp_flash_dispatcher/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+set(includes "include")
+
+idf_component_register(SRCS "esp_flash_dispatcher.c"
+                       INCLUDE_DIRS ${includes}
+                       PRIV_REQUIRES "spi_flash" "freertos" "esp_system"
+                       )
+
+set(WRAP_FUNCTIONS
```
suda-morris, line 8: question, what if the esp_flash.h introduced a new function which can also disable the cache, how can we remember to update this list?

mythbuster5: A series target test would be helpful. But I don't know whether we use extra component in idf test app is still a good idea nowadays.

esp_flash_dispatcher/README.md
```diff
@@ -0,0 +1,5 @@
+## Overview
+
+When a task stack resides in PSRAM, invoking `esp_flash_*` directly in that task can cause crashes during moments when the Flash driver temporarily disables the CPU cache (resulting in the PSRAM stack being inaccessible). Typical failures include Cache errors and Guru Meditation.
+
+The `esp_flash_dispatcher` component intercepts common Flash APIs and executes the real Flash operations in a dedicated background task whose stack lives in internal RAM. This guarantees that even while cache is disabled, Flash operations run on an always-accessible stack. Therefore, application tasks can keep their stacks in PSRAM without worrying about Flash operations breaking them.
```
igrr, line 5:
```suggestion
The `esp_flash_dispatcher` component intercepts common Flash APIs and executes the real Flash operations in a dedicated background task whose stack lives in internal RAM. This guarantees that even while cache is disabled, Flash operations run on an always-accessible stack. Therefore, application tasks can keep their stacks in PSRAM without worrying about Flash operations breaking them.

## Usage

1. Add esp_flash_dispatcher component to your project: `idf.py add-dependency espressif/esp_flash_dispatcher`
2. Initialize esp_flash_dispatcher in app_main:

    #include "esp_flash_dispatcher.h"
    
    ...
    
    const esp_flash_dispatcher_config_t cfg = {
        .task_stack_size = 2048,
        .task_priority = 10,
        .task_core_id = tskNO_AFFINITY,
        .queue_size = 5,
    };
    ESP_ERROR_CHECK(esp_flash_dispatcher_init(&cfg));
    
3. Now you can call any API which writes or reads SPI Flash from a task with the stack in PSRAM, no other changes are required.
```

(I couldn't add the C code block inside another code block, please add that manually)

esp_flash_dispatcher/include/esp_flash_dispatcher.h
```diff
@@ -0,0 +1,42 @@
...
+
+/* Configuration for the flash dispatcher task.
+ * The dispatcher serializes flash operations to a dedicated task.
+ */
+typedef struct {
+    uint32_t task_stack_size;   // Stack size for the dedicated flash task (in bytes)
+    uint32_t task_priority;     // Priority for the dedicated flash task
+    BaseType_t task_core_id;    // Core affinity (PRO_CPU_NUM, APP_CPU_NUM, or tskNO_AFFINITY)
+    uint32_t queue_size;        // Length of the request queue
+} esp_flash_dispatcher_config_t;
```
igrr, line 24: Perhaps adding ESP_FLASH_DISPATCHER_CONFIG_DEFAULT() macro would be reasonable.

igrr (approved):
LGTM aside from adding the usage instructions to the README.

esp_flash_dispatcher/CMakeLists.txt
igrr: I suppose the risk is not too high, I don't think we have added any new esp_flash APIs in a long while.

Perhaps next time you will be modifying spi_flash component, add a comment somewhere in esp_flash_api.c: `// When adding new APIs which disable the cache, don't forget to update esp_flash_dispatcher component.`

suda-morris: another way could be: define a macro e.g. `FLASH_FUNC_DIS_CACHE` beside the flash API declaration, and here we run a small cmake script to parse the header file and find out the APIs that need wrapper.

esp_flash_dispatcher/.build-test-rules.yml
```diff
@@ -0,0 +1,6 @@
+esp_flash_dispatcher/test_apps:
+  disable:
+    - if: SOC_SPIRAM_XIP_SUPPORTED != 1
+      reason: "This feature is only necessary on chips which have PSRAM, and can XIP from PSRAM"
```
suda-morris, line 4: Why does the chip have to support XIP PSRAM to use this component? Is ESP32 not possible? @mythbuster5

mythbuster5: This can be used on esp32, but the test only work with XIP.  But on the other word, ESP32 MSPI doesn't allow concurrent access of SPI0 and SPI1, so seems this component is not that important for esp32

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,208 @@
...
+            if (request.calling_task_handle) {
+                xTaskNotify(request.calling_task_handle, request.result, eSetValueWithOverwrite);
+            }
+        }
+    }
+}
+
+esp_err_t esp_flash_dispatcher_init(const esp_flash_dispatcher_config_t *cfg)
+{
+    // If already initialized, return success
```
suda-morris: I see we're return invalid state here, the comment is misleading

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,208 @@
...
+    if (s_flash_dispatcher_ctx.queue != NULL || s_flash_dispatcher_ctx.task != NULL) {
+        ESP_EARLY_LOGE(TAG, "flash dispatcher already initialized");
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    s_flash_dispatcher_ctx.queue = xQueueCreate(cfg->queue_size, sizeof(flash_operation_request_t));
```
suda-morris: The queue objects are still possible to be allocated from the PSRAM, what if the cache is disabled when the background task fetches an item that is located in the PSRAM?

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,207 @@
...
+                    cfg->task_stack_size,
+                    (void *)xTaskGetCurrentTaskHandle(),
+                    cfg->task_priority,
+                    &s_flash_dispatcher_ctx.task,
+                    cfg->task_core_id,
+                    MALLOC_CAP_INTERNAL);
+
+    if (rc != pdPASS) {
+        // Cleanup resources if task creation failed
+        vQueueDelete(s_flash_dispatcher_ctx.queue);
```
suda-morris:
```suggestion
        vQueueDeleteWithCaps(s_flash_dispatcher_ctx.queue);
```

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,208 @@
...
+        return ESP_ERR_INVALID_STATE;
+    }
+
+    s_flash_dispatcher_ctx.queue = xQueueCreate(cfg->queue_size, sizeof(flash_operation_request_t));
+    ESP_RETURN_ON_FALSE(s_flash_dispatcher_ctx.queue, ESP_ERR_NO_MEM, TAG, "create flash operation queue failed");
+
+    BaseType_t rc = xTaskCreatePinnedToCoreWithCaps(flash_dispatcher_task,
+                    "flash_dispatcher",
+                    cfg->task_stack_size,
+                    (void *)xTaskGetCurrentTaskHandle(),
```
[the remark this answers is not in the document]
suda-morris: this comment  is still valid @mythbuster5

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,207 @@
...
+        .result = ESP_FAIL
+    };
+
+    if (xQueueSend(s_flash_dispatcher_ctx.queue, &request, portMAX_DELAY) != pdTRUE) {
+        ESP_EARLY_LOGE(TAG, "Failed to send %s request to queue", op_name ? op_name : "flash");
+        return ESP_ERR_TIMEOUT;
+    }
+
+    uint32_t notification_value = ESP_FAIL;
+    xTaskNotifyWait(0, 0xFFFFFFFF, &notification_value, portMAX_DELAY);
```
suda-morris: I doubt if using the task notification is the best choice here. We can't control caller's task (e.g. task A). Task A may be also using the task notification for other purposes, but didn't realize that the Flash API is expecting a notification too. e.g. What if the caller's task is also expecting a notification from a gptimer ISR?

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,221 @@
...
+} flash_operation_t;
+
+// Structure to hold flash operation requests
+typedef struct {
+    esp_flash_t *chip;
+    void *arg1;
+    void *arg2;
+    void *arg3;
+    flash_operation_t op;            // Operation type to execute
+    esp_err_t *result_out;            // Pointer where the result will be written
```
suda-morris: why do we want to save the address of the result, instead of save the **value** itself? sizeof(esp_err_t) typically is 4, which is same as a pointer.

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,221 @@
...
+            default:
+                ESP_EARLY_LOGE(TAG, "Unsupported flash operation type: %d", (int)request.op);
+                result = ESP_FAIL;
+                break;
+            }
+            // Publish result and signal completion to the waiting caller
+            if (request.result_out) {
+                *request.result_out = result;
+            }
+            if (s_flash_dispatcher_ctx.done_sem) {
```
suda-morris: this check is redundant.

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -0,0 +1,221 @@
...
+            case FLASH_OP_ERASE_CHIP:
+                result = __real_esp_flash_erase_chip(request.chip);
+                break;
+            default:
+                ESP_EARLY_LOGE(TAG, "Unsupported flash operation type: %d", (int)request.op);
+                result = ESP_FAIL;
+                break;
+            }
+            // Publish result and signal completion to the waiting caller
+            if (request.result_out) {
```
suda-morris: also this check is redundant

esp_flash_dispatcher/esp_flash_dispatcher.c
mythbuster5: This should be pointer, because get the result and return the result are in different tasks. But I changed this to a result queue.

PR #585, opened 2025-10-06
feat: add cJSON component

mahavirj:
# Checklist

- [ ] CI passing

# Change description


cjson/sbom_cJSON.yml
```diff
@@ -0,0 +1,8 @@
+name: cJSON
+version: "1.7.19"
+cpe: cpe:2.3:a:cjson_project:cjson:{}:*:*:*:*:*:*:*
+cpe: cpe:2.3:a:davegamble:cjson:{}:*:*:*:*:*:*:*
```
mahavirj: @fhrbata Please check if we can specify multiple cpe records here. At-least in the scan, I see only latest cpe record after this change:

```
  "records": [
    {
      "vulnerable": "NO",
      "pkg_name": "cJSON",
      "pkg_version": "1.7.19",
      "cve_id": "",
      "cvss_base_score": "",
      "cvss_base_severity": "",
      "cvss_version": "",
      "cvss_vector_string": "",
      "cpe": "cpe:2.3:a:davegamble:cjson:1.7.19:*:*:*:*:*:*:*",
      "keyword": "",
      "cve_link": "",
      "cve_desc": "",
      "exclude_reason": "",
      "status": ""
    },
```

cjson/sbom_cJSON.yml
```diff
@@ -0,0 +1,8 @@
...
+supplier: 'Person: Dave Gamble'
+description: "cJSON: Ultralightweight JSON parser in ANSI C"
+url: https://github.com/DaveGamble/cJSON
+hash: c859b25da02955fef659d658b8f324b5cde87be3
```
Ashish285: Maybe we also need to add the excluded CVE list as this will now perform name based search as well

```
cve-exclude-list:
  - cve: CVE-2024-31755
    reason: Resolved in v1.7.18
  - cve: CVE-2023-26819
    reason: Resolved in commit a328d65ad490b64da8c87523cbbfe16050ba5bf6
  - cve: CVE-2023-53154
    reason: Resolved in v1.7.18
  - cve: CVE-2025-57052
    reason: Resolved in v1.7.19
```

mahavirj: I checked using `esp-idf-sbom manifest check --name --local-db  --format=json   cjson/` and the scan looks clean.

cjson/sbom_cJSON.yml
mahavirj: Syntax issue - now fixed

cjson/sbom_cJSON.yml
Ashish285: Ok 👍

cjson/sbom_cJSON.yml
fhrbata: Yeah, IIRC the default yaml parser in python uses the last key seen. I think we would need to adjust the parser to detect this. So yes, the `cpe` for multiple cpes should be a list.

cjson/sbom_cJSON.yml
fhrbata: I believe these were added over the time, because the CVEs were not yet processed in NVD. I also checked and it seems these excludes are no longer needed.

fhrbata (approved):
LGTM, thank you!

PR #593, opened 2025-10-14
fix: Update CI workflow to fix Linux host tests (IEC-399)

adokitkat:
It seems like Linux host tests were broken. This PR fixes the issue and also creates a separate job for running Linux host tests so they don't need to wait for all (even non-Linux apps) to be built. Changed Linux runners from self-hosted ones to Github runners.

Uploads .elf binaries to artifacts, gives the execution permission, selects correct runners, etc.

Here is a passing pipeline with these changes applied (look at "Build Apps for Linux target" and "Run apps for Linux target"): https://github.com/espressif/idf-extra-components/actions/runs/18497009630 (CI related commits will be removed from the other PR when this is merged)

# Checklist

- [x] CI passing

### DIFF
--- a/.github/workflows/build_and_run_apps.yml
+++ b/.github/workflows/build_and_run_apps.yml
@@ -131,7 +131,8 @@ jobs:
           - "release-v5.5"
           - "latest"
     runs-on: ubuntu-22.04
-    container: espressif/idf:${{ matrix.idf_ver }}
+    container:
+      image: espressif/idf:${{ matrix.idf_ver }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -156,22 +157,16 @@ jobs:
         with:
           name: app_binaries_${{ matrix.idf_ver }}_linux
           path: |
-            */examples/*/build_linux*/bootloader/bootloader.bin
-            */examples/*/build_linux*/partition_table/partition-table.bin
-            */examples/*/build_linux*/*.bin
-            */examples/*/build_linux*/flasher_args.json
+            */examples/*/build_linux*/*.elf
             */examples/*/build_linux*/config/sdkconfig.json
-            */test_app*/**/build_linux*/bootloader/bootloader.bin
-            */test_app*/**/build_linux*/partition_table/partition-table.bin
-            */test_app*/**/build_linux*/*.bin
-            */test_app*/**/build_linux*/flasher_args.json
+            */test_app*/**/build_linux*/*.elf
             */test_app*/**/build_linux*/config/sdkconfig.json
             build_info*.json
 
   run-target:
     name: Run apps on target
     if: github.repository_owner == 'espressif' && needs.prepare.outputs.build_only != '1'
-    needs: [build, build-linux]
+    needs: [build]
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_run_apps.yml
```diff
@@ -253,10 +243,72 @@ jobs:
...
 
+  run-target-linux:
+    name: Run apps on Linux target
+    if: github.repository_owner == 'espressif' && needs.prepare.outputs.build_only != '1'
+    needs: [build-linux]
+    strategy:
+      fail-fast: false
+      matrix:
+        idf_ver:
+          # - "release-v5.1" # Not testing for 5.1 with linux target due to limited support
```
pacucha42, line 254: Just wondering: is there a chance to use something like "idf_ver: > release-v5.1"? First: better readability Second: no need to add new IDF versions one by one

Not a blocker for this PR

.github/workflows/build_and_run_apps.yml
```diff
@@ -253,10 +243,72 @@ jobs:
...
+            target: "linux"
+            pytest_args: "--embedded-services idf"
+        exclude:
+          - idf_ver: "release-v5.2" # Bug with Unity IDF test apps
+            runner:
+              target: "linux"
+    env:
+      TEST_RESULT_NAME: test_results_${{ matrix.runner.target }}_${{ matrix.runner.marker }}_${{ matrix.idf_ver }}
+      TEST_RESULT_FILE: test_results_${{ matrix.runner.target }}_${{ matrix.runner.marker }}_${{ matrix.idf_ver }}.xml
+    runs-on: ubuntu-24.04
```
pacucha42: Just wondering 2: using specific system version sounds too specific... what happens once the host system gets upgraded? I see this setting comes from the previous code but it's still a valid question

adokitkat: When new LTS images come out, the old ones get cycled out in some time. They have a N - 1 support policy for their public runners. So when Ubuntu 26.04 comes out, 22.04 stops working after some time I guess and users will have to upgrade their scripts. I think I can change this to "ubuntu-latest" in the Linux case since it is just a runner OS running Docker on it. I don't know if the other job is pinned to Ubuntu 22.04 for a reason or if it's ok to update it.

.github/workflows/build_and_run_apps.yml
adokitkat: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations#using-contexts-to-create-matrices I found something like this possible but it still isn't e.g. range based. It may be worth it so these strings aren't repeated but they still need to be written manually somewhere.

pacucha42: Hmm, fancy - that said, I think it's not worthy of any effort. Thanks for the explanation

.github/workflows/build_and_run_apps.yml
pacucha42: "OS-latest" sounds like the safest option, unless we are bound to OS-specific version

adokitkat: updated

PR #594, opened 2025-10-16
feat: add esp_schedule v1.3.2 from esp-rainmaker (IEC-402)

Junius00:
# Checklist

- [?] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8) (link is broken)
- [ ] CI passing

# Change description
Migration of esp_schedule to here to have a common version usable by both ESP RainMaker and ESP RMNG.

### DIFF
--- a/esp_schedule/CMakeLists.txt
+++ b/esp_schedule/CMakeLists.txt
@@ -1,0 +1,7 @@
+set(component_srcs "src/esp_schedule.c"
+                   "src/esp_schedule_nvs.c")
+
+idf_component_register(SRCS "${component_srcs}"
+                       INCLUDE_DIRS "include"
+                       PRIV_INCLUDE_DIRS "src"
+                       PRIV_REQUIRES nvs_flash esp_netif)
--- a/esp_schedule/README.md
+++ b/esp_schedule/README.md
@@ -1,0 +1,28 @@
+# ESP Scheduling
+
+[![Component Registry](https://components.espressif.com/components/espressif/esp_schedule/badge.svg)](https://components.espressif.com/components/espressif/esp_schedule)
+
+This component is used to implement scheduling for:
+
+- **One-shot events** with a relative time difference (e.g., 30 seconds into the future)
+- **Periodic events** based on a certain time[^1] on days of the week (e.g., every Monday or Wednesday)
+- **Periodic/one-shot events** on a certain time[^1] based on the date:
+  - e.g., *(periodic)* every 23rd of January to April
+  - e.g., *(one-shot)* 9th of August, 2026
+- **Periodic events** at an offset from sunrise/sunset
+
+[^1]: By default, the time is w.r.t. UTC. If the timezone has been set, then the time is w.r.t. the specified timezone.
+
+## Example Usage
+
+See the comprehensive example in [`examples/get_started/`](examples/get_started/) for a complete demonstration of all ESP Schedule features, including:
+
+- **Days of Week Scheduling** - Recurring events on specific weekdays
+- **Date-based Scheduling** - Monthly and yearly recurring events
+- **Relative Scheduling** - One-time delayed events
+- **Solar Scheduling** - Sunrise/sunset based events with location coordinates and day-of-week filtering
+- **Schedule Persistence** - NVS storage and recovery
+- **Callback Handling** - Trigger and timestamp callbacks
+- **Schedule Management** - Create, edit, enable, and disable schedules
+
+The example includes detailed documentation, build instructions, and demonstrates all schedule types with practical use cases.
--- a/esp_schedule/examples/get_started/main/network/app_network.h
+++ b/esp_schedule/examples/get_started/main/network/app_network.h
@@ -1,0 +1,72 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/event_groups.h"
+#include "network_provisioning/manager.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Event group bits for network provisioning states */
+#define NETWORK_CONNECTED_BIT      BIT0
+#define NETWORK_DISCONNECTED_BIT   BIT1
+#define PROVISIONING_SUCCESS_BIT   BIT2
+#define PROVISIONING_FAILED_BIT    BIT3
+#define TIME_SYNC_SUCCESS_BIT      BIT4
+#define TIME_SYNC_FAILED_BIT       BIT5
+
+/**
+ * @brief Initialize network provisioning with event group handling
+ *
+ * This function initializes the network provisioning system and sets up event handlers
+ * for network provisioning and IP events. It uses either BLE or SoftAP provisioning
+ * based on the Kconfig selection.
+ *
+ * @param event_group Pointer to the event group that will be used for synchronization
+ * @return esp_err_t ESP_OK on success, or error code on failure
+ */
+esp_err_t app_network_init(EventGroupHandle_t event_group);
+
+/**
+ * @brief Start network provisioning and wait for connection
+ *
+ * This function checks if the device is already provisioned. If provisioned,
+ * it starts WiFi station and waits for network connection. If not provisioned,
+ * it starts the provisioning sequence and waits for network connection.
+ *
+ * @param event_group The event group to wait on
+ * @param timeout_ms Timeout in milliseconds
+ * @return esp_err_t ESP_OK if network connected, ESP_FAIL if failed or timed out
+ */
+esp_err_t app_network_start(EventGroupHandle_t event_group, uint32_t timeout_ms);
+
+/**
+ * @brief Start time synchronization
+ *
+ * This function starts SNTP time synchronization and waits for it to complete.
+ *
+ * @param event_group The event group to signal completion on
+ */
+void app_network_start_time_sync(EventGroupHandle_t event_group);
+
+/**
+ * @brief Wait for time synchronization to complete
+ *
+ * This function waits for time synchronization to complete successfully.
+ *
+ * @param event_group The event group to wait on
+ * @param timeout_ms Timeout in milliseconds
+ * @return esp_err_t ESP_OK if time sync succeeded, ESP_FAIL if failed or timed out
+ */
+esp_err_t app_network_wait_for_time_sync(EventGroupHandle_t event_group, uint32_t timeout_ms);
+
+#ifdef __cplusplus
+}
+#endif
--- a/esp_schedule/idf_component.yml
+++ b/esp_schedule/idf_component.yml
@@ -1,0 +1,11 @@
+## IDF Component Manager Manifest File
+version: "1.3.2"
+description: Task scheduling based on periodic and one-time events
+url: https://github.com/espressif/idf-extra-components/tree/master/components/esp_schedule
+repository: https://github.com/espressif/idf-extra-components.git
+issues: https://github.com/espressif/esp-rainmaker/issues
+dependencies:
+  idf: ">=5.1"
+  espressif/esp_daylight:
+    version: "~1.0.1"
+    override_path: "../esp_daylight"
--- a/esp_schedule/include/esp_schedule.h
+++ b/esp_schedule/include/esp_schedule.h
@@ -1,0 +1,257 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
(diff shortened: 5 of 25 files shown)


esp_schedule/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(component_srcs "src/esp_schedule.c"
+                   "src/esp_schedule_nvs.c")
+
+idf_component_register(SRCS "${component_srcs}"
+                       INCLUDE_DIRS "include"
+                       PRIV_INCLUDE_DIRS "src"
+                       PRIV_REQUIRES "rmaker_common"
+                       REQUIRES "nvs_flash" "esp_daylight")
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-function")
```
igrr: Recommend using 

```suggestion
target_compile_options(${COMPONENT_LIB} PRIVATE -Wno-unused-function)
```
instead of manipulating CMake variables

esp_schedule/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(component_srcs "src/esp_schedule.c"
+                   "src/esp_schedule_nvs.c")
+
+idf_component_register(SRCS "${component_srcs}"
+                       INCLUDE_DIRS "include"
+                       PRIV_INCLUDE_DIRS "src"
+                       PRIV_REQUIRES "rmaker_common"
+                       REQUIRES "nvs_flash" "esp_daylight")
```
igrr:
```suggestion
                       PRIV_REQUIRES nvs_flash)
```
1. esp_daylight is already in idf_component.yml, no need to add it here
2. header files from nvs_flash aren't used in public header files of esp_schedule, they are only used in source files. Hence it's sufficient to use PRIV_REQUIRES.

esp_schedule/README.md
```diff
@@ -0,0 +1,50 @@
+# ESP Scheduling
+
+[![Component Registry](https://components.espressif.com/components/espressif/esp_schedule/badge.svg)](https://components.espressif.com/components/espressif/esp_schedule)
+
+This component is used internally by ESP RainMaker to implement schedules.
```
igrr: If this is now a general-purpose component, this line could be rewritten.

esp_schedule/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+> Note: By default, the time is w.r.t. UTC. If the timezone has been set, then the time is w.r.t. the specified timezone.
```
igrr: For notes and other admonitions, please use GFM syntax — it will render nicely both on Github and in the component registry.

esp_schedule/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Test code:
+
+```
```
igrr:
```suggestion
```cpp
```

esp_schedule/README.md
```diff
@@ -0,0 +1,50 @@
...
+#include <esp_schedule.h>
```
igrr:
```suggestion
#include "esp_schedule.h"
```

By convention, angle brackets are used for system and C or C++ standard library header files.

esp_schedule/idf_component.yml
```diff
@@ -0,0 +1,12 @@
...
+description: ESP Schedules, used in RainMaker
+url: https://github.com/espressif/idf-extra-components/tree/master/components/esp_schedule
+repository: https://github.com/espressif/idf-extra-components.git
+issues: https://github.com/espressif/esp-rainmaker/issues
+dependencies:
+  idf: ">=5.1"
+  espressif/rmaker_common:
+    version: "~1.5.0"
+  espressif/esp_daylight:
+    version: "~1.0.1"
```
igrr, line 10:
```suggestion
    version: "~1.0.1"
    override_path: "../esp_daylight"
```

This way, CI will use the actual code from esp_daylight directory of this repository, and won't fetch esp_daylight from the registry.

esp_schedule/idf_component.yml
```diff
@@ -0,0 +1,12 @@
+## IDF Component Manager Manifest File
+version: "1.3.1"
+description: ESP Schedules, used in RainMaker
+url: https://github.com/espressif/idf-extra-components/tree/master/components/esp_schedule
+repository: https://github.com/espressif/idf-extra-components.git
+issues: https://github.com/espressif/esp-rainmaker/issues
+dependencies:
+  idf: ">=5.1"
+  espressif/rmaker_common:
```
igrr: I'd prefer to not make idf-extra-components CI depend on the Rainmaker repository. What specifically does this component use from rmaker_common? Can we avoid that anyhow?

esp_schedule/include/esp_schedule.h
```diff
@@ -0,0 +1,254 @@
...
+ * esp_schedule_get() followed by esp_schedule_edit() with the correct callbacks.
+ *
+ * @param[in] enable_nvs If NVS is to be enabled or not.
+ * @param[in] nvs_partition (Optional) The NVS partition to be used. If NULL is passed, the default partition is used.
+ * @param[out] schedule_count Number of active schedules found in NVS.
+ *
+ * @return Array of schedule handles if any schedules have been found.
+ * @return NULL if no schedule is found in NVS (or if NVS is not enabled).
+ */
+esp_schedule_handle_t *esp_schedule_init(bool enable_nvs, char *nvs_partition, uint8_t *schedule_count);
```
igrr, line 174: Not something that should be fixed in this PR, but a refactoring suggestion for the next version:

In newer components written for IDF, we usually use different factory functions for different implementations. For example, for a non-NVS enabled scheduler, one would call `esp_schedule_create_default`. For an NVS-enabled scheduler, one would call `esp_schedule_create_nvs`, or similar.

The main advantage of this approach is that the optional dependency (NVS in this case) can be removed by the linker from the final binary, if the application only calls esp_schedule_create_default and never calls esp_schedule_create_nvs.

Plus, a nitpick: I'd recommend returning an esp_err_t and then passing esp_schedule_handle as an output pointer parameter, as that allows you to communicate the errors more precisely to the caller.

esp_schedule/src/esp_schedule.c
```diff
@@ -0,0 +1,852 @@
+/*
+ * SPDX-FileCopyrightText: 2023 Espressif Systems (Shanghai) CO LTD
```
igrr: Please update the years.

esp_schedule/src/esp_schedule.c
```diff
@@ -0,0 +1,852 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#include <string.h>
+#include <inttypes.h>
+#include <esp_log.h>
+#include <esp_sntp.h>
+#include <esp_rmaker_utils.h>
+#include <esp_daylight.h>
```
igrr:
```suggestion
#include "esp_daylight.h"
```
and likewise for esp_log, esp_sntp, etc. Only string.h and inttypes.h are standard headers.

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -0,0 +1,278 @@
...
+        *schedule_count = 0;
+        return NULL;
+    }
+    int handle_count = 0;
+
+    nvs_entry_info_t nvs_entry;
+    nvs_iterator_t nvs_iterator = NULL;
+    esp_err_t err = nvs_entry_find(esp_schedule_nvs_partition, ESP_SCHEDULE_NVS_NAMESPACE, NVS_TYPE_BLOB, &nvs_iterator);
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "No entry found in NVS");
```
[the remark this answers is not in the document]
Junius00: Ownership of handle_list is transferred to the caller

esp_schedule/idf_component.yml
Junius00, line 1: @igrr when this component is used directly from the ESP component registry (as part of managed_components), is this override path still valid? or will it cause errors?

igrr: It won't cause issues, the registry will remove the overrides when the component is uploaded.

Junius00:
> Since the component doesn't contain an example, please add at least a minimal unit test app. Otherwise, CI won't compile any code from this component at all.

I've added an example under examples/get_started

esp_schedule/src/esp_schedule.c
```diff
@@ -0,0 +1,852 @@
...
+{
+    if (schedule_config == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    if (handle == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    esp_schedule_t *schedule = (esp_schedule_t *)handle;
+
+    strcpy(schedule_config->name, schedule->name);
```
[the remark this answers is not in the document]
igrr: I think this is reasonable, can we use strlcpy instead?

esp_schedule/src/esp_schedule_nvs.c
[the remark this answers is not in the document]
igrr: Also in case of the error?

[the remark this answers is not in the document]
Junius00: Added free on failure

esp_schedule/src/esp_schedule.c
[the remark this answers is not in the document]
Junius00: Changed to strlcpy

Junius00:
fixed all remaining issues and squashed to a single commit

esp_schedule/examples/get_started/main/network/app_network.h
```diff
@@ -0,0 +1,72 @@
...
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/event_groups.h"
+#include "network_provisioning/manager.h"
```
suda-morris:
```suggestion
#include "freertos/FreeRTOS.h"
#include "freertos/event_groups.h"
#include "network_provisioning/manager.h"

#ifdef __cplusplus
extern "C" {
#endif
```
C linkage is ensured by the included header files internally

esp_schedule/include/esp_schedule.h
```diff
@@ -0,0 +1,256 @@
...
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <time.h>
+#include "esp_err.h"
```
suda-morris:
```suggestion
#include <stdint.h>
#include <stdbool.h>
#include <time.h>
#include "esp_err.h"

#ifdef __cplusplus
extern "C" {
#endif
```

esp_schedule/include/esp_schedule.h
```diff
@@ -0,0 +1,256 @@
...
+ */
+typedef void (*esp_schedule_timestamp_cb_t)(esp_schedule_handle_t handle, uint32_t next_timestamp, void *priv_data);
+
+/** Schedule type */
+typedef enum esp_schedule_type {
+    ESP_SCHEDULE_TYPE_INVALID = 0,
+    ESP_SCHEDULE_TYPE_DAYS_OF_WEEK,
+    ESP_SCHEDULE_TYPE_DATE,
+    ESP_SCHEDULE_TYPE_RELATIVE,
+#if CONFIG_ESP_SCHEDULE_ENABLE_DAYLIGHT
```
suda-morris, line 51: please include "sdkconfig.h" explicitly

esp_schedule/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+set(component_srcs "src/esp_schedule.c"
+                   "src/esp_schedule_nvs.c")
+
+idf_component_register(SRCS "${component_srcs}"
+                       INCLUDE_DIRS "include"
+                       PRIV_INCLUDE_DIRS "src"
+                       PRIV_REQUIRES nvs_flash esp_netif)
+
+target_compile_options(${COMPONENT_LIB} PRIVATE -Wno-unused-function)
```
suda-morris: why not fix the "unused-function" warning since the source code are all in our control?

Junius00: probably a leftover from another template; didn't manage to find any unused functions when removing this suppression. Will remove.

suda-morris (approved):
Sorry, there should also be a [.build-test-rules.yml](https://github.com/espressif/idf-extra-components/blob/master/esp_daylight/.build-test-rules.yml) file for the esp_schedule component, to control the build and test of the example.

Junius00:
> Sorry, there should also be a [.build-test-rules.yml](https://github.com/espressif/idf-extra-components/blob/master/esp_daylight/.build-test-rules.yml) file for the esp_schedule component, to control the build and test of the example.

@suda-morris how is this done? do I add a blank .build-test-rules.yml file in the esp_schedule directory and edit the top-level .build-test-rules.yml like the rest of the components?

Junius00:
@suda-morris added to top-level .build-test-rules.yml, please check if the format is correct

PR #597, opened 2025-10-28
[esp_linenoise] Fix small bugs and findings

SoucheSouche:
# Checklist
- [x] CI passing

# Change description
- Added getters for out_fd and in_fd config structure fields
- Removed self field in esp_linenoise_instance_t and related sanity check performed on it
- Added NULL pointer checks in public API
- Fixed typo in macro

PR #598, opened 2025-10-29
Fix prompt length issue (IEC-405)

sidwarkd:
# Checklist

- [ ] CI passing

# Change description
When writing the prompt, the length was incorrectly always the size of a char* pointer. The change uses strlen to properly write the correct size of the provided prompt string.


sidwarkd:
@SoucheSouche Here is an issue I found with the new component. Otherwise it seems to be working very well.

SoucheSouche:
> @SoucheSouche Here is an issue I found with the new component. Otherwise it seems to be working very well.

@sidwarkd, nice catch! thanks for taking the time to check it out :)
I opened a PR myself with a few fixes here and there and new getter functions (see #597)

PR #599, opened 2025-10-30
feat(json_generator): add int64 support (IEC-406)

Junius00:
# Checklist

- [ ] CI passing

# Change description
Add int64_t support for JSON generation to support larger values (e.g., millisecond timestamps)

### DIFF
--- a/json_generator/idf_component.yml
+++ b/json_generator/idf_component.yml
@@ -1,3 +1,3 @@
-version: "1.1.3"
+version: "1.2.0"
 description: A simple JSON (JavasScript Object Notation) generator with flushing capability
 url: https://github.com/espressif/json_generator


json_generator/src/json_generator.c
```diff
@@ -18,10 +18,12 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <string.h>
+#include<inttypes.h>
```
suda-morris:
```suggestion
#include <inttypes.h>
```

suda-morris (approved):

json_generator/idf_component.yml
```diff
@@ -1,3 +1,3 @@
-version: "1.1.2"
+version: "1.1.3"
```
mahavirj: Suggest changing this to `1.2.0` instead - minor version update to highlight new feature addition

mahavirj:
@shahpiyushv PTAL

json_generator/idf_component.yml
Junius00: updated to 1.2.0

shahpiyushv (approved):
Changes look fine to me

PR #603, opened 2025-11-03
feat(esp_schedule): v2.0.0 (IEC-408)

Junius00:
# Checklist

- [ ] CI passing

# Change description

Added in v2.0.0:
- Make NVS usage optional via Kconfig
- Refactor trigger logic:
    - Allow more than one trigger per schedule
    - More flexible date constraints
- Add glue layer for NVS, timer, time, memory allocation, logging implementations used by component:
    - This is necessary for RMNG, which has to compile under normal CMake conditions and cannot rely on ESP-IDF
    - Component still behaves like a normal ESP-IDF component with default implementations
- Add Unity test app


esp_schedule/src/esp_schedule.c
```diff
@@ -4,848 +4,743 @@
...
-    schedule_config->trigger.minutes = schedule->trigger.minutes;
-    if (schedule->trigger.type == ESP_SCHEDULE_TYPE_DAYS_OF_WEEK) {
-        schedule_config->trigger.day.repeat_days = schedule->trigger.day.repeat_days;
-    } else if (schedule->trigger.type == ESP_SCHEDULE_TYPE_DATE) {
-        schedule_config->trigger.date.day = schedule->trigger.date.day;
-        schedule_config->trigger.date.repeat_months = schedule->trigger.date.repeat_months;
-        schedule_config->trigger.date.year = schedule->trigger.date.year;
-        schedule_config->trigger.date.repeat_every_year = schedule->trigger.date.repeat_every_year;
-    }
+    strcpy(schedule_config->name, schedule->name);
```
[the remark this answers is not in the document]
Junius00: replaced with strlcpy

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -25,232 +26,372 @@
...
-    nvs_entry_info_t nvs_entry;
-    nvs_iterator_t nvs_iterator = NULL;
-    esp_err_t err = nvs_entry_find(esp_schedule_nvs_partition, ESP_SCHEDULE_NVS_NAMESPACE, NVS_TYPE_BLOB, &nvs_iterator);
-    if (err != ESP_OK) {
-        ESP_LOGE(TAG, "No entry found in NVS");
-        free(handle_list);
+    esp_schedule_nvs_iterator_t nvs_iterator = NULL;
+    esp_schedule_nvs_error_t err = esp_schedule_nvs_entry_find_blobs(esp_schedule_nvs_partition, ESP_SCHEDULE_NVS_NAMESPACE, &nvs_iterator);
+    if (err != ESP_SCHEDULE_NVS_OK) {
+        ESP_SCHEDULE_LOGE(TAG, "No entry found in NVS");
```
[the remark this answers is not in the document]
Junius00: added free on error return

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
+{
+#if !ESP_SCHEDULE_NVS_ENABLED
+    // force disable NVS if it is not enabled
+    enable_nvs = false;
+#endif
+
+    /* Handle default initialization */
+    if (!enable_nvs) {
+        esp_schedule_init_default();
+        *schedule_count = 0;
```
shahpiyushv, line 754: v1 never touched `schedule_count` when `enable_nvs` was false, so existing callers legally pass NULL here (esp-rainmaker does exactly this). Since this function exists for backward compatibility, it needs a NULL check or it crashes them on boot.

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
+                    }
+
+                    int dst_adjust = 0;
+                    if (!current_tm.tm_isdst && candidate_tm.tm_isdst) {
+                        dst_adjust = -3600;
+                    } else if (current_tm.tm_isdst && !candidate_tm.tm_isdst) {
+                        dst_adjust = 3600;
+                    }
+                    ESP_SCHEDULE_LOGD(TAG, "DST adjust seconds: %d", dst_adjust);
+                    result_time += dst_adjust;
```
shahpiyushv, line 147: This manual DST compensation makes schedules fire an hour off across DST transitions — verified on host: a daily 12:00 schedule in America/New_York fires at 11:00 after spring-forward and 13:00 after fall-back. `mktime()` already resolves the epoch using the candidate's `tm_isdst`, so the ±3600 double-corrects. v1 needed this hack because it fed `mktime()` the *current* day's isdst; now that `candidate_tm` comes from `localtime()` of the target date, this block should just be dropped (and ideally set `tm_isdst = -1` before the final `mktime()`).

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
+                    }
+                    if (!validity || validity->start_time == 0 || result_time >= validity->start_time) {
+                        *next_time = result_time;
+                        return true;
+                    }
+                    /* else fall through to search next valid day */
+                }
+            }
+
+            candidate_time += SECONDS_IN_DAY;
```
shahpiyushv, line 161: Advancing by wall-clock seconds skips a calendar day when `now` is within an hour of midnight and the walk crosses spring-forward — e.g. querying "next Sunday 23:45" on Sat Mar 8 23:30 (America/New_York) returns Mar 16 instead of Mar 9. Safer to increment `tm_mday` and re-`mktime()` (careful: the month-rollover check below relies on `candidate_tm` being un-normalized).

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
         schedule->trigger_cb((esp_schedule_handle_t)schedule, schedule->priv_data);
     }
-
-restart_schedule:
-
-    if (esp_schedule_is_expired(&schedule->trigger)) {
-        /* Not deleting the schedule here. Just not starting it again. */
-        return;
-    }
     esp_schedule_start_timer(schedule);
```
shahpiyushv, line 428: No expiry check before re-arming, and in the new engine a mask of 0 means "any day" — so `ESP_SCHEDULE_DAY_ONCE` schedules now fire every day, and one-time DATE schedules every month. v1 called `is_expired()` here and stopped. Related: `date.repeat_every_year` is no longer read anywhere.

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
+            schedule->trigger_cb = NULL;
+            schedule->timestamp_cb = NULL;
+            schedule->timer = NULL;
+            /* Check for ONCE and expired schedules and delete them. */
+            if (esp_schedule_is_expired(schedule)) {
+                /* This schedule has already expired. */
+                ESP_SCHEDULE_LOGI(TAG, "Schedule %s does not repeat and has already expired. Deleting it.", schedule->name);
+                esp_schedule_delete((esp_schedule_handle_t)schedule);
+                /* Removing the schedule from the list */
+                for (int i = handle_count; i < *schedule_count - 2; i++) {
```
shahpiyushv, line 729: Off-by-one: bound should be `*schedule_count - 1`. As is, deleting an expired entry from `[A,B,C]` leaves `[B,B]` — one handle duplicated (double-free risk) and one schedule leaked with its timer running. v1's swap-with-last here was simpler and correct.

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
 
-    /* Check if schedule is not this year itself, it is in future. */
-    if (trigger->date.year > (current_time->tm_year + 1900)) {
-        /* Find first schedule month of next year */
-        next_schedule_month = ffs(repeat_months);
-        /* Year will be handled by the caller. So no need to add any additional months */
-        return next_schedule_month;
-    }
+    time_t candidate_time = mktime(&candidate_tm);
+    candidate_tm = *localtime(&candidate_time);
```
shahpiyushv, line 74: Bare `localtime()` (also at 100, 133, 162, 193) mixed with `localtime_r()` elsewhere — the static buffer isn't safe with both the timer task and app tasks in this code. v1 used `localtime_r()` throughout.

esp_schedule/src/esp_schedule.c
```diff
@@ -4,849 +4,784 @@
...
+            tr->next_scheduled_time_utc = 0;
         }
     }
+    return ESP_SCHEDULE_RET_OK;
+}
+
+static ESP_SCHEDULE_RETURN_TYPE esp_schedule_set(esp_schedule_t *schedule, esp_schedule_config_t *schedule_config)
+{
+    /* Deep copy trigger list from config, freeing previous if any. */
+    if (schedule->triggers.list) {
```
shahpiyushv, line 519: The old trigger list is freed before the new allocation is attempted, so an OOM during edit leaves the schedule with no triggers. Allocate-then-swap would keep it intact. (The second free check at 526 is dead code from the same block.)

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -25,232 +26,374 @@ static const char *TAG = "esp_schedule_nvs";
...
+        ESP_SCHEDULE_LOGE(TAG, "Could not allocate schedule");
+        esp_schedule_nvs_close(nvs_handle);
+        ESP_SCHEDULE_FREE(blob_buffer);
         return NULL;
     }
-    nvs_close(nvs_handle);
-    ESP_LOGI(TAG, "Schedule %s found in NVS", schedule->name);
+
+    /* Copy persistent schedule structure from beginning of blob */
+    esp_schedule_persistent_t *persistent = (esp_schedule_persistent_t *) (blob_buffer);
```
shahpiyushv, line 295: The blob is parsed with no validation — no version/magic, no check that `buf_size` covers the persistent struct plus `count * sizeof(trigger)`, and `data_len` below underflows on a short blob. A v1-format blob after OTA, a corrupt entry, or even toggling `ESP_SCHEDULE_ENABLE_DAYLIGHT` (which changes `sizeof(esp_schedule_trigger_t)`) all lead to overreads. A format-version byte plus size checks would make bad blobs safely rejectable.

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -25,232 +26,374 @@ static const char *TAG = "esp_schedule_nvs";
...
+        return ESP_SCHEDULE_RET_NO_MEM;
+    }
+
+    /* Create persistent schedule structure and copy to beginning of buffer */
+    esp_schedule_persistent_t *persistent = (esp_schedule_persistent_t *) (blob_buffer);
+    persistent->next_scheduled_time_diff = schedule->next_scheduled_time_diff;
+    persistent->next_scheduled_time_utc = schedule->next_scheduled_time_utc;
+    persistent->validity = schedule->validity;
+    strlcpy(persistent->name, schedule->name, sizeof(persistent->name));
+    persistent->triggers = schedule->triggers; // Copy trigger list metadata
```
shahpiyushv, line 106: This writes the live heap pointer into flash. Only `count` is worth persisting — storing the pointer also makes the format depend on pointer width and padding.

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -25,232 +26,374 @@ static const char *TAG = "esp_schedule_nvs";
...
-        ESP_LOGI(TAG, "Found schedule in NVS with key: %s", nvs_entry.key);
-        handle_list[handle_count] = esp_schedule_nvs_get(nvs_entry.key);
+    while (err == ESP_SCHEDULE_NVS_OK) {
+        char *next_key = NULL;
+        esp_schedule_nvs_entry_get_key(nvs_iterator, &next_key);
+        if (next_key == NULL) {
+            break;
+        }
+        ESP_SCHEDULE_LOGI(TAG, "Found schedule in NVS with key: %s", next_key);
+        handle_list[handle_count] = esp_schedule_nvs_get(next_key);
```
shahpiyushv, line 386: `handle_list` is sized from the count key but filled from the blob iterator with no bounds check — if they ever disagree (e.g. power loss between `set_blob` and the count update in `nvs_add`), this overflows the array. Worth clamping to `handle_count < *schedule_count`.

esp_schedule/glue/esp/timer.c
```diff
@@ -0,0 +1,132 @@
...
+    TimerHandle_t timer_handle = (TimerHandle_t) * p_timer_handle;
+    if (timer_handle != NULL) {
+        esp_schedule_timer_cancel(p_timer_handle);
+    }
+
+    __timer_priv_data_t *timer_priv_data = __timer_priv_data_create(cb, priv_data);
+    if (timer_priv_data == NULL) {
+        return;
+    }
+    timer_handle = xTimerCreate("schedule", ( ((uint64_t) delay_seconds) * 1000 ) / portTICK_PERIOD_MS, pdFALSE, timer_priv_data, __timer_common_cb);
```
shahpiyushv, line 89: The multiply is 64-bit now (nice), but `xTimerCreate`/`xTimerChangePeriod` take a 32-bit `TickType_t`, so anything beyond ~49.7 days at `CONFIG_FREERTOS_HZ=1000` still truncates and fires early. The engine happily returns dates years out — maybe cap the period and re-arm on expiry?

esp_schedule/Kconfig
```diff
@@ -1,5 +1,21 @@
...
+        bool "Enable NVS Persistence"
+        default y
+        help
+            Enable NVS (Non-Volatile Storage) persistence for schedules.
+
+            This allows schedules to be saved to flash memory and automatically
+            restored after device reboot or power cycle.
+
+            Disabling this option will:
+            - Remove NVS-related API functions (esp_schedule_create_nvs, etc.)
```
shahpiyushv, line 13: Help text mentions `esp_schedule_create_nvs()` / `esp_schedule_create_default()` — actual APIs are `esp_schedule_init_nvs()` / `esp_schedule_init_default()`.

esp_schedule/include/common/esp_schedule_untyped.h
```diff
@@ -0,0 +1,444 @@
...
+    struct {
+        /** Latitude in decimal degrees (-90 to +90, positive North) */
+        double latitude;
+        /** Longitude in decimal degrees (-180 to +180, positive East) */
+        double longitude;
+        /** Offset in minutes from sunrise/sunset (positive = after, negative = before) */
+        int offset_minutes;
+    } solar;
+#endif
+    /** For type ESP_SCHEDULE_TYPE_SECONDS */
```
shahpiyushv, line 173: Nit: `ESP_SCHEDULE_TYPE_SECONDS` doesn't exist (should be `ESP_SCHEDULE_TYPE_RELATIVE`?). Also the note at the top requires `ESP_SCHEDULE_RET_NVS_NOT_FOUND` to be defined, but `include/esp/esp_schedule.h` doesn't define it and nothing uses it.

shahpiyushv (commented):
Thanks — the glue layer and multi-trigger refactor are a clear step forward, and the new date engine fixes real v1 bugs (day-31 in short months, Feb 29 in non-leap years). I pulled the engine into a host test harness to verify behavior though, and a few things regressed vs 1.x — most visibly DST handling (schedules fire ±1h around transitions) and one-time schedules repeating forever. Details inline.

One suggestion on sequencing: since there's no backport flow in this repo, could this be split into two PRs? First a 1.3.4 with just the bug fixes that don't depend on the new architecture — day-31 schedules firing on Mar 3 after February, Feb 29 firing on Mar 1 in non-leap years, and the 32-bit timer period overflow — and then the 2.0.0 breaking changes on top of that. That way existing 1.x consumers (esp-rainmaker pins `~1.3.2`) get the fixes with a plain version bump instead of having to take the breaking migration first.

Junius00:
Will be broken into smaller PRs instead. See #783

PR #606, opened 2025-11-10
feat(spi_nand_flash): Add block device support

RathiSonika:
# Change description

* Add block device support (nand_flash_bdl and nand_wl_bdl)
* Please refer spi_nand_flash/layered_architecture.md for the details and migration guide
* Refactor the component for improved structure, maintainability and readability
            - Added page-based API terminology (`read_page`, `write_page`, `get_page_count`, `get_page_size`) with backward-compatible sector aliases.

* Removed the fatfs/vfs component dependency; it will be relocated (along with examples) to another component in a different PR. 
* Add Kconfig option NAND_FLASH_ENABLE_BDL to enable/disable BDL support
* Updated test_apps and host_tests to verify the same

* OTP and OOB area handling are still missing. I’ll address them in a separate PR, as the scope of this one is already growing.


spi_nand_flash/include/esp_nand_blockdev.h
```diff
@@ -0,0 +1,268 @@
...
+ *
+ * Used with:
+ * - ESP_BLOCKDEV_CMD_IS_BAD_BLOCK
+ * - ESP_BLOCKDEV_CMD_MARK_BAD_BLOCK
+ * - ESP_BLOCKDEV_CMD_IS_FREE_PAGE
+ */
+typedef struct {
+    uint32_t num;               /*!< IN:  Block number or page number */
+    bool status;                /*!< OUT: Bad block status or page free status (true/false) */
+} esp_blockdev_cmd_arg_status_t;
```
omnitex, line 153: 1. `ESP_BLOCKDEV_CMD_MARK_BAD_BLOCK` shouldn't be listed here as "used with", correct? In other places I see that his command has only a single `uint32_t` as an argument.

2. As multiple (at least `XX_IS_YY`) commands re-use the same structure, wouldn't it be better to `typedef` it for each command separately? -- so each command uses it's own uniquely named struct (even though they might be the same behind `typedef`).
- From ease-of-use standpoint, not having to keep track of which command structure is being reused by what command, make each command have 1+1 (define + structure).
- So e.g. add `typedef esp_blockdev_cmd_arg_status_t esp_blockdev_cmd_arg_is_free_page_t` et.c (or also with `status` in there as well, something that would make sense, but the name would grow long quite quickly...)
- The part after `esp_blockdev_cmd_arg_...` could fully match (right now some command defines have additional `GET`) the command define name after `ESP_BLOCKDEV_CMD_...` so the pairing is very clear.

RathiSonika: 1. Yes, I already noticed that and fixed it. I’ll push the changes soon.
2. Good suggestion, using a typedef is a better option. I’ll update it accordingly. 
Thank you!

spi_nand_flash/src/dhara_glue.c
```diff
@@ -18,21 +18,25 @@
...
-static esp_err_t dhara_init(spi_nand_flash_device_t *handle)
+static esp_err_t dhara_init(spi_nand_flash_device_t *handle, esp_blockdev_handle_t flash_handle)
 {
     // create a holder structure for dhara context
     spi_nand_flash_dhara_priv_data_t *dhara_priv_data = calloc(1, sizeof(spi_nand_flash_dhara_priv_data_t));
     // save the holder inside the device structure
     handle->ops_priv_data = dhara_priv_data;
     // store the pointer back to device structure in the holder structure
     dhara_priv_data->parent_handle = handle;
+    dhara_priv_data->flash_handle = flash_handle;
```
omnitex: For the newly added `flash_handle`: wouldn't something like `bdl_flash_handle` or just `bdl_handle` be a better name?

- `spi_nand_flash_device_t` *has* "flash" in the name, in `dhara_priv_data` it's `parent_handle` -- fair enough (yet smt like `parent_spi_flash_handle`?)
- `esp_blockdev_handle_t` *doesn't have* "flash" in the name, in `dhara_priv_data` it's `flash_handle` -- any reason why we wouldn't want naming it explicitly `bdl...`?

spi_nand_flash/src/dhara_glue.c
```diff
@@ -18,21 +18,25 @@
...
     struct dhara_map dhara_map;
+    esp_blockdev_handle_t flash_handle;
     spi_nand_flash_device_t *parent_handle;
 } spi_nand_flash_dhara_priv_data_t;
 
-static esp_err_t dhara_init(spi_nand_flash_device_t *handle)
+static esp_err_t dhara_init(spi_nand_flash_device_t *handle, esp_blockdev_handle_t flash_handle)
 {
     // create a holder structure for dhara context
     spi_nand_flash_dhara_priv_data_t *dhara_priv_data = calloc(1, sizeof(spi_nand_flash_dhara_priv_data_t));
```
omnitex: I see that for basically all other malloc memory allocations, `heap_caps_malloc()` is used (so I assume it's the preferred way, even when `MALLOC_CAP_DEFAULT` is used, in `test_app/main/test_spi_nand_flash.c`).

Would it be beneficial for clarity to use `heap_caps_calloc()` here (and other calloc instances) as well? The docs specify [In IDF, calloc(p) is equivalent to heap_caps_calloc(p, MALLOC_CAP_8BIT)](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/mem_alloc.html#_CPPv416heap_caps_calloc6size_t6size_t8uint32_t), so the 8bit access requirement is "hidden" in the implementation details of IDF calloc?

I just realized changes like this are not really in the goal and scope of this PR so please ignore if it's not relevant.

spi_nand_flash/src/dhara_glue.c
RathiSonika: Well, when I was writing the code I had a different idea in mind. But yes, using bdl_handle makes more sense from a code-readability perspective. Fixed

spi_nand_flash/src/dhara_glue.c
RathiSonika: That’s okay — your input is always welcome. I’ve addressed this comment. Thank you.

spi_nand_flash/diskio/diskio_nand_blockdev.c
```diff
@@ -0,0 +1,177 @@
...
+        ret = bdl->ops->sync(bdl);
+        if (ret != ESP_OK) {
+            ESP_LOGE(TAG, "sync failed with error 0x%X", ret);
+            return RES_ERROR;
+        }
+        break;
+
+    case GET_SECTOR_COUNT:
+        uint32_t sector_count;
+        ret = bdl->ops->ioctl(bdl, ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS, &sector_count);
```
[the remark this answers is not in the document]
pacucha42: Resolved. I can't close the thread, though (lack of permissions?). @RathiSonika please have a look

spi_nand_flash/test_app/main/test_spi_nand_flash_bdl.c
```diff
@@ -0,0 +1,324 @@
...
+    esp_blockdev_handle_t bdl_handle;
+    setup_nand_flash(&spi, SPI_NAND_IO_MODE_SIO, SPI_DEVICE_HALFDUPLEX, &bdl_handle);
+
+    TEST_ESP_OK(bdl_handle->ops->erase(bdl_handle, 0, bdl_handle->geometry.disk_size));
+
+    do_single_write_test(bdl_handle, 1, 1);
+    deinit_nand_flash(spi, bdl_handle);
+}
+
+static void check_buffer(uint32_t seed, const uint8_t *src, size_t count)
```
pacucha42: Please:
* update both `check_buffer()` and `fill_buffer()` to be more effective
* if possible, put the code into a shared library (file) to avoid duplicates in both components

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
```diff
@@ -0,0 +1,106 @@
...
+    nand_file_mmap_emul_config_t conf = {"", 50 * 1024 * 1024, true};
+    spi_nand_flash_config_t nand_flash_config = {&conf, 0, SPI_NAND_IO_MODE_SIO, 0};
+    esp_blockdev_handle_t nand_bdl;
+    REQUIRE(nand_flash_get_blockdev(&nand_flash_config, &nand_bdl) == 0);
+
+    uint32_t block_size = nand_bdl->geometry.erase_size;
+    uint32_t block_num = nand_bdl->geometry.disk_size / block_size;
+
+    uint32_t test_block = 15;
+    if (test_block < block_num) {
```
pacucha42: What happens when there is less than 15 blocks? based on the code, the test provides no results but we won't see any warning/failure note

spi_nand_flash/include/esp_nand_blockdev.h
```diff
@@ -0,0 +1,295 @@
...
+
+/** @brief Get number of available logical sectors (WL layer only)
+ *
+ * Usage:
+ * @code
+ * uint32_t available_sectors;
+ * esp_err_t ret = wl_bdl->ops->ioctl(wl_bdl, ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS, &available_sectors);
+ * @endcode
+ */
+#define ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS    (ESP_BLOCKDEV_CMD_USER_BASE + 1)
```
pacucha42: Why do you exhaust user command space here, please? SPI NAND BDL should be considered IDF system internals

spi_nand_flash/include/esp_nand_blockdev.h
```diff
@@ -0,0 +1,295 @@
...
+ *         - ESP_FAIL: Other initialization failure
+ *
+ * @note The returned block device handle must be released with bdl->ops->release(bdl)
+ *       when no longer needed.
+ *
+ * @note This creates the FLASH layer. For filesystem use, you typically want
+ *       the WEAR-LEVELING layer instead (see spi_nand_flash_wl_get_blockdev).
+ *
+ */
+esp_err_t nand_flash_get_blockdev(spi_nand_flash_config_t *config,
```
pacucha42, line 235: Can we create new BDL instance for existing NAND flash device driver?

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
pacucha42: Consider deploying `assert` instead of the condition check. Thank you

spi_nand_flash/include/esp_nand_blockdev.h
```diff
@@ -0,0 +1,295 @@
...
+/**
+ * @brief Argument structure for block/page status commands
+ *
+ * Used with:
+ * - ESP_BLOCKDEV_CMD_IS_BAD_BLOCK
+ * - ESP_BLOCKDEV_CMD_IS_FREE_PAGE
+ */
+typedef struct {
+    uint32_t num;               /*!< IN:  Block number or page number */
+    bool status;                /*!< OUT: Bad block status or page free status (true/false) */
```
pacucha42: Cosmetics: please, emphasize the difference between the two statuses in the comment: bad-block status [good/bad] and page status [free/occupied]. Thanks

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
RathiSonika: Updated

spi_nand_flash/test_app/main/test_spi_nand_flash_bdl.c
RathiSonika: Updated. Created separate spi_nand_flash_test_helper.h to use these functions

spi_nand_flash/include/esp_nand_blockdev.h
RathiSonika: I used ESP_BLOCKDEV_CMD_USER_BASE because these ioctl commands are implemented in the WL layer,  which acts as the user/client of the spi_nand_flash driver (BDL). Based on that, I chose ESP_BLOCKDEV_CMD_USER_BASE.

However, since these commands are actually provided by the driver component, I think you may be right that ESP_BLOCKDEV_CMD_SYSTEM_BASE would be more appropriate.

I’m not entirely clear on when ESP_BLOCKDEV_CMD_USER_BASE should be used.

spi_nand_flash/include/esp_nand_blockdev.h
RathiSonika: As discussed offline, we have a single BDL flash instance mapped to the WL layer. A single WL instance manages the entire flash.

Therefore, there is no need to create multiple BDL instances for the NAND flash driver. The flash BDL and WL BDL will be created once during flash initialization.

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,326 @@
...
+static const esp_blockdev_ops_t nand_flash_blockdev_ops = {
+    .read = nand_flash_blockdev_read,
+    .write = nand_flash_blockdev_write,
+    .erase = nand_flash_blockdev_erase,
+    .ioctl = nand_flash_blockdev_ioctl,
+    .sync = nand_flash_blockdev_sync_no_op,
+    .release = nand_flash_blockdev_release,
+};
+
+esp_err_t nand_flash_get_blockdev(spi_nand_flash_config_t *config, esp_blockdev_handle_t *out_bdl_handle_ptr)
```
pacucha42, line 312: There is no `device_flags` initialization in this BDL getter. Please, use `ESP_BLOCKDEV_FLAGS_INST_CONFIG_DEFAULT()` at least

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,326 @@
...
+        ESP_LOGE(TAG, "Write address 0x%llx not aligned to page size %u", dst_addr, handle->geometry.write_size);
+        return ESP_ERR_INVALID_ARG;
+    }
+    spi_nand_flash_device_t *dev_handle = (spi_nand_flash_device_t *)handle->ctx;
+    uint32_t page = dst_addr / handle->geometry.write_size;
+    esp_err_t res = nand_prog(dev_handle, page, src_buf);
+    return res;
+}
+
+static esp_err_t nand_flash_blockdev_erase(esp_blockdev_handle_t handle, uint64_t start_addr, size_t erase_len)
```
pacucha42, line 124: It seems you are deleting only one block here (!!!). `erase_len` is never used in this context.

Am I missing anything?

spi_nand_flash/src/nand_wl_blockdev.c
```diff
@@ -0,0 +1,234 @@
...
+        esp_blockdev_cmd_arg_erase_t *trim_arg = (esp_blockdev_cmd_arg_erase_t *)args;
+        uint32_t page_size = handle->geometry.write_size;
+        if ((trim_arg->start_addr % page_size) != 0 ||
+                (trim_arg->erase_len % page_size != 0)) {
+            ESP_LOGE(TAG, "Start address/length 0x%llx/0x%llx not aligned to page size %u", trim_arg->start_addr, trim_arg->erase_len, page_size);
+            return ESP_ERR_INVALID_ARG;
+        }
+        uint32_t start_page = trim_arg->start_addr / page_size;
+        uint32_t page_count = trim_arg->erase_len / page_size;
+        for (uint32_t page_id = start_page; page_id <= (start_page + page_count); page_id++) {
```
pacucha42: This is wrong - off by one, you go out of range on the last page! Proper condition check is `page_id < (start_page + page_count)`.

Such points should be easily detected by properly designed tests.

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,326 @@
...
+            return ESP_ERR_INVALID_SIZE;
+        }
+
+        uint32_t num_pages = handle->geometry.disk_size / handle->geometry.write_size;
+        bool is_free = true;
+        uint32_t ecc_err_total_count = 0;
+        uint32_t ecc_err_exceeding_threshold_count = 0;
+        uint32_t ecc_err_not_corrected_count = 0;
+        esp_err_t ret = ESP_OK;
+        for (uint32_t page = 0; page < num_pages; page++) {
```
pacucha42, line 254: Note: this can be very slow on large devices... any idea for performance improvement? Like continuous stats caching?

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,326 @@
...
+    return res;
+}
+
+static esp_err_t nand_flash_blockdev_sync_no_op(esp_blockdev_handle_t handle)
+{
+    if (handle == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+    esp_err_t res = ESP_OK;
+    return res;
```
pacucha42:
```suggestion
    return ESP_OK;
```

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,326 @@
...
+
+    case ESP_BLOCKDEV_CMD_GET_NAND_FLASH_INFO: {
+        esp_blockdev_cmd_arg_nand_flash_info_t *flash_info = (esp_blockdev_cmd_arg_nand_flash_info_t *)args;
+        flash_info->device_info.manufacturer_id = dev->device_info.manufacturer_id;
+        flash_info->device_info.device_id = dev->device_info.device_id;
+        memcpy(flash_info->device_info.chip_name, dev->device_info.chip_name, sizeof(dev->device_info.chip_name));
+        memcpy(&flash_info->geometry, &dev->chip, sizeof(nand_flash_geometry_t));
+        return ESP_OK;
+    }
+    break;
```
pacucha42: Unreachable `break`

spi_nand_flash/src/nand_wl_blockdev.c
```diff
@@ -0,0 +1,234 @@
...
+    esp_err_t ret = ESP_OK;
+    if (handle == NULL) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    esp_blockdev_handle_t nand_handle = (esp_blockdev_handle_t)handle->ctx;
+    spi_nand_flash_device_t *dev_handle = (spi_nand_flash_device_t *)nand_handle->ctx;
+
+#ifdef CONFIG_IDF_TARGET_LINUX
+    ret = nand_emul_deinit(dev_handle);
```
pacucha42: Just wondering: is this call consitent with `nand_handle->ops->release(nand_handle)` ? Why is this call needed? (haven't checked the details, it just looks suspicious)

spi_nand_flash/test_app/main/test_spi_nand_flash_bdl.c
```diff
@@ -0,0 +1,304 @@
...
+static void do_single_write_test(esp_blockdev_handle_t bdl, uint32_t start_page, uint16_t page_count)
+{
+    uint8_t *temp_buf = NULL;
+    uint8_t *pattern_buf = NULL;
+    uint32_t page_size = bdl->geometry.write_size;
+
+    uint32_t num_pages;
+    bdl->ops->ioctl(bdl, ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS, &num_pages);
+
+    TEST_ESP_OK((start_page + page_count) > num_pages);
```
pacucha42: This is wrong check - `TEST_ESP_OK` verifies a value of `esp_err_t`, not a boolean value (which happens as 0/1 integer in your case)

spi_nand_flash/src/nand_flash_blockdev.c
RathiSonika: You’re right. For context: Dhara’s NAND interface only issues single-block erase (see dhara_nand_erase() in dhara_glue.c: it takes one block index b and calls BDL erase with erase_len = erase_size). So in our stack, all Dhara-driven erase was effectively one block, and the application normally uses trim (MARK_DELETED) on the WL BDL rather than calling Flash BDL erase directly. So the bug didn’t show up in normal use, but the Flash BDL still has to honour erase_len for API consistency and for direct use. I’ve updated the Flash BDL erase to loop over the requested range and erase each block. Thank you.

spi_nand_flash/src/nand_flash_blockdev.c
RathiSonika: Agreed it can be slow on large devices. But this is intended for debug/diagnostics (e.g. detecting flash degradation, development and field diagnostics), not for use in the normal I/O path. So I kept the implementation simple: a full scan to give accurate, on-demand stats without maintaining extra state. 

If we ever need this in a latency-sensitive or production path, we could add optional caching (e.g. background stats or sampling) in a follow-up. I can add a short comment in the code and/or in the API doc that it may be slow on large devices and is meant for diagnostics.
What do you think?

spi_nand_flash/src/dhara_glue.c
```diff
@@ -158,33 +180,46 @@ esp_err_t nand_unregister_dev(spi_nand_flash_device_t *handle)
...
+                    uint8_t *data, dhara_error_t *err)
 {
     spi_nand_flash_dhara_priv_data_t *dhara_priv_data = __containerof(n, spi_nand_flash_dhara_priv_data_t, dhara_nand);
-    spi_nand_flash_device_t *dev_handle = dhara_priv_data->parent_handle;
-    bool is_bad_status = false;
-    if (nand_is_bad(dev_handle, b, &is_bad_status)) {
-        return 1;
+    spi_nand_flash_device_t *dev_handle = NULL;
+    esp_err_t ret = ESP_OK;
+#ifdef CONFIG_NAND_FLASH_ENABLE_BDL
```
omnitex, line 191:
```c
#ifdef CONFIG_NAND_FLASH_ENABLE_BDL
    if (dhara_priv_data->bdl_handle != NULL) {
        esp_blockdev_handle_t bdl_handle = dhara_priv_data->bdl_handle;
        dev_handle = (spi_nand_flash_device_t *)bdl_handle->ctx;
        ret = bdl_handle->ops->read(bdl_handle, data, bdl_handle->geometry.read_size,
                                    (p * bdl_handle->geometry.read_size) + offset, length);
    } else
#endif
    {
        dev_handle = dhara_priv_data->parent_handle;
        ret = nand_read(dev_handle, p, offset, length, data);
    }
    if (ret != ESP_OK) {
        if (dev_handle->chip.ecc_data.ecc_corrected_bits_status == NAND_ECC_NOT_CORRECTED) {
            dhara_set_error(err, DHARA_E_ECC);
        }
        return -1;
    }
```
Bug in `dhara_glue.c`?

When `CONFIG_NAND_FLASH_ENABLE_BDL` was set, each Dhara NAND callback had:
- An `#ifdef CONFIG_NAND_FLASH_ENABLE_BDL` block that only ran when `bdl_handle != NULL`.
- An `#else` block that used `parent_handle` and the legacy `nand_*` APIs, but that block was compiled only when BDL was disabled.

So with BDL enabled and the legacy API (e.g. `spi_nand_flash_init_device` passing `NULL` as `bdl_handle`), the BDL branch was skipped and there was no fallback: **callbacks did nothing and still returned success.**

Changes
- The `#else` block was replaced with an `else `that runs when BDL is enabled but `bdl_handle == NULL`.
- That `else` uses `parent_handle` and the same legacy `nand_*` calls as before.

Resulting behavior:
- BDL enabled, bdl_handle != NULL → BDL path (blockdev ops).
- BDL enabled, bdl_handle == NULL or BDL disabled → legacy path (parent_handle + nand_read / nand_prog / etc.).

RathiSonika: The current `#ifdef CONFIG_NAND_FLASH_ENABLE_BDL / #else` structure is intentional. When BDL is enabled, all I/O must go through the BDL stack -- falling back to the legacy nand_* APIs would defeat the purpose of the BDL architecture.

The bdl_handle is already validated as non-NULL in spi_nand_flash_wl_get_blockdev() before dhara_init is called (line 177 in nand_wl_blockdev.c). So when BDL is enabled, bdl_handle will always be non-NULL in these callbacks.

The legacy spi_nand_flash_init_device() path (which passes NULL) is not intended to be used when BDL is enabled -- they are mutually exclusive usage models.

Maybe I'll replace the if (bdl_handle != NULL) runtime check with an assert(dhara_priv_data->bdl_handle != NULL) so that any incorrect usage fails loudly during development rather than silently doing nothing.

omnitex: > The legacy spi_nand_flash_init_device() path (which passes NULL) is not intended to be used when BDL is enabled -- they are mutually exclusive usage models.

but the mutual exclusivity is not currently strongly enforced, as if I either by mistake or intentionally enable BDL but use legacy path, it would lead to "silently doing nothing", right?

I understand the intended usage, my AI agent found this and considered it a bug. And as you say, it might be better to treat "any incorrect usage" loudly

.build-test-rules.yml
```diff
@@ -79,22 +79,26 @@ esp_lcd_qemu_rgb/examples/lcd_qemu_rgb_panel:
...
       reason: The spi_nand_flash component is compatible with IDF version v5.0 and above, due to a change in the f_mkfs API in versions above v5.0, which is not supported in older IDF versions.
+    - if: IDF_VERSION_MAJOR < 6 and CONFIG_NAME == "bdl"
+      reason: BDL support requires the esp_blockdev interface, available starting from IDF version v6.0.
 
 spi_nand_flash/host_test:
   enable:
     - if: IDF_TARGET == "linux"
+  disable:
+    - if: IDF_VERSION_MAJOR <= 5 and IDF_VERSION_MINOR < 3
+      reason: The spi_nand_flash component is compatible with IDF version v5.0 and above, due to a change in the f_mkfs API in versions above v5.0, which is not supported in older IDF versions.
```
omnitex: > The spi_nand_flash component is compatible with IDF version v5.0 and above

why does the condition specify `IDF_VERSION_MAJOR <= 5 and IDF_VERSION_MINOR < 3`? 5.0 vs 5.3?

also 

The single condition `IDF_VERSION_MAJOR <= 5` and `IDF_VERSION_MINOR < 3` only disables when `MINOR < 3`. 

So:
- IDF 4.4 (MAJOR=4, MINOR=4): 4 <= 5 ✓, 4 < 3 ✗ → not disabled (wrong).
- IDF 4.3 (MAJOR=4, MINOR=3): same → not disabled (wrong).
- ...?

if the check is supposed to be on 5.3

```yml
 - if: IDF_VERSION_MAJOR < 5
      reason: The spi_nand_flash component is compatible with IDF version v5.0 and above, due to a change in the f_mkfs API in versions above v5.0, which is not supported in older IDF versions.
- if: IDF_VERSION_MAJOR == 5 and IDF_VERSION_MINOR < 3
      reason: The spi_nand_flash component is compatible with IDF version v5.0 and above, due to a change in the f_mkfs API in versions above v5.0, which is not supported in older IDF versions.
```

something like this?

spi_nand_flash/src/nand_wl_blockdev.c
```diff
@@ -0,0 +1,231 @@
...
+
+static const char *TAG = "nand_wl_blockdev";
+
+/**************************************************************************************
+ **************************************************************************************
+ * Block Device Layer interface implementation
+ **************************************************************************************
+ */
+
+static esp_err_t spi_nand_flash_wl_blockdev_read(esp_blockdev_handle_t handle, uint8_t *dst_buf, size_t dst_buf_size, uint64_t src_addr, size_t data_read_len)
```
omnitex, line 23: wouldn't it be better for `dst_buf_size` and `data_read_len` which are unused to either

1) cast them as `(void)` plus comment about unused & why? I think that's the best practice right?
2) use compiler `__attribute__((unused))`
3) add some checks/asserts for the values even though they are not used?

omnitex: this applies to other unused arguments in e.g. write functions as well

spi_nand_flash/src/nand_wl_blockdev.c
```diff
@@ -0,0 +1,231 @@
...
+        uint32_t page_size = handle->geometry.write_size;
+        if ((trim_arg->start_addr % page_size) != 0 ||
+                (trim_arg->erase_len % page_size != 0)) {
+            ESP_LOGE(TAG, "Start address/length 0x%llx/0x%llx not aligned to page size %u", trim_arg->start_addr, trim_arg->erase_len, page_size);
+            return ESP_ERR_INVALID_ARG;
+        }
+        uint32_t start_page = trim_arg->start_addr / page_size;
+        uint32_t page_count = trim_arg->erase_len / page_size;
+        for (uint32_t page_id = start_page; page_id < (start_page + page_count); page_id++) {
+            ret = spi_nand_flash_trim(dev_handle, page_id);
```
omnitex, line 176: useless saving of return value, when it's not checked each iteration?

any intermediate fail of the trim operation does not terminate the loop and only the last return value is saved and ultimately returned

should we do the same checking for the return value as is the case in `spi_nand_flash_wl_blockdev_erase()`?

omnitex: also, is there some "sanitization" of the `trim_arg` values? could we go beyond the address space in the loop, beyond the number of pages in the device?

or is the ioctl for trim only possible to be called from Dhara so we trust the values?

spi_nand_flash/src/dhara_glue.c
```diff
@@ -18,21 +18,32 @@
...
+#endif
     spi_nand_flash_device_t *parent_handle;
 } spi_nand_flash_dhara_priv_data_t;
 
-static esp_err_t dhara_init(spi_nand_flash_device_t *handle)
+static esp_err_t dhara_init(spi_nand_flash_device_t *handle, void *bdl_handle)
 {
     // create a holder structure for dhara context
-    spi_nand_flash_dhara_priv_data_t *dhara_priv_data = calloc(1, sizeof(spi_nand_flash_dhara_priv_data_t));
+    spi_nand_flash_dhara_priv_data_t *dhara_priv_data = heap_caps_calloc(1, sizeof(spi_nand_flash_dhara_priv_data_t), MALLOC_CAP_DEFAULT);
```
omnitex, line 39: missing check if `heap_caps_calloc()` returns `NULL`?

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,340 @@
...
+
+/**************************************************************************************
+ **************************************************************************************
+ * Block Device Layer interface implementation
+ **************************************************************************************
+ */
+
+static esp_err_t nand_flash_blockdev_read(esp_blockdev_handle_t handle, uint8_t *dst_buf, size_t dst_buf_size, uint64_t src_addr, size_t data_read_len)
+{
+    if (handle == NULL) {
```
omnitex: I found this NULL check a bit confusing at first glance, because `handle` is not a pointer, as the pointer is hidden behind a typedef

do you think it would be better to check here against `ESP_BLOCKDEV_HANDLE_INVALID` (typedefed NULL) so it's cleaner imo?

would apply to all the `handle == NULL` checks))

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,340 @@
...
+
+    case ESP_BLOCKDEV_CMD_IS_FREE_PAGE: {
+        esp_blockdev_cmd_arg_is_free_page_t *page_free_status = (esp_blockdev_cmd_arg_is_free_page_t *) args;
+        esp_err_t ret = nand_is_free(dev, page_free_status->num, &page_free_status->status);
+        return ret;
+    }
+
+    case ESP_BLOCKDEV_CMD_GET_PAGE_ECC_STATUS: {
+        esp_blockdev_cmd_arg_ecc_status_t *page_ecc_status = (esp_blockdev_cmd_arg_ecc_status_t *) args;
+        esp_err_t ret = nand_get_ecc_status(dev, page_ecc_status->page_num);
```
omnitex, line 196: the return value of `nand_get_ecc_status()` is ignored here?

compared with call of the same method in`case ESP_BLOCKDEV_CMD_GET_ECC_STATS`

spi_nand_flash/src/nand_impl_linux.c
```diff
@@ -13,6 +13,98 @@
...
+    (*handle)->chip.ecc_data.ecc_data_refresh_threshold = 4;
+    (*handle)->chip.log2_ppb = 6;         // 64 pages per block is standard
+    (*handle)->chip.log2_page_size = 11;  // 2048 bytes per page is fairly standard
+    (*handle)->chip.num_planes = 1;
+    (*handle)->chip.flags = 0;
+
+    ESP_GOTO_ON_ERROR(detect_chip(*handle), fail, TAG, "Failed to detect nand chip");
+
+    (*handle)->chip.page_size = 1 << (*handle)->chip.log2_page_size;
+    (*handle)->chip.block_size = (1 << (*handle)->chip.log2_ppb) * (*handle)->chip.page_size;
```
omnitex: `detect_chip()` calculates and sets the `block_size` using `emulated_page_size`, but then that gets overwritten here, calculated by the non-emulated (without oob) page size?

spi_nand_flash/src/nand_impl_linux.c
```diff
@@ -13,6 +13,98 @@
...
+    return ret;
+
+fail:
+    free((*handle)->work_buffer);
+    free((*handle)->read_buffer);
+    free((*handle)->temp_buffer);
+    if ((*handle)->mutex) {
+        vSemaphoreDelete((*handle)->mutex);
+    }
+    free(*handle);
```
omnitex, line 100: should set handle to NULL after freeing, so it's not a dangling pointer, on failure path after successful allocation?

spi_nand_flash/src/dhara_glue.c
omnitex: ^^ like I have code that uses lagacy API, but by mistake I turn on BDL in menuconfig. Then it would due to the `#ifdef` compile only the BDL part of the code, but because I am using legacy, it passes NULL, `#else` is not-compiled in, so it's skipped, `ret` is still left as `ESP_OK`, so the whole function returns 0, without actually doing anything, right?

spi_nand_flash/src/nand_flash_blockdev.c
RathiSonika: Good point. But these op functions are only ever called via handle->ops->read(handle, ...) (and similarly for write/erase/etc.), so the caller has already dereferenced handle to get the function pointer. If handle were NULL, the fault would happen at the call site, not inside the op. So a check for handle == NULL or ESP_BLOCKDEV_HANDLE_INVALID inside the op is redundant. So removed these redundant checks.

spi_nand_flash/src/dhara_glue.c
RathiSonika: Added `assert(dhara_priv_data->bdl_handle != NULL)` so that any incorrect usage fails loudly rather than silently doing nothing. Thank you for pointing this out.

spi_nand_flash/src/nand_wl_blockdev.c
RathiSonika: Fixed

spi_nand_flash/src/nand_wl_blockdev.c
RathiSonika: Added multipage read/write support. This was missing in both WL and flash BDL. Thank you!

spi_nand_flash/src/nand_flash_blockdev.c
```diff
@@ -0,0 +1,356 @@
...
+        }
+        free(handle);
+        return ESP_ERR_NO_MEM;
+    }
+    blockdev->ctx = (void *)handle;
+    *out_bdl_handle_ptr = blockdev;
+
+    blockdev->ops = &nand_flash_blockdev_ops;
+
+    ESP_BLOCKDEV_FLAGS_INST_CONFIG_DEFAULT(blockdev->device_flags);
```
haberturdeur: This is incorrect, as this sets erase_before_write to true, which should not be the case.

haberturdeur: Oh, sorry, this is the raw non-dhara version...

In that case it makes sense...

But I would still explicitly write out the flags, because it makes it easier to see the full context in one place. It also makes it safer against a change in the macro (which might happen, because there is a "bug" (difference between the intended behaviour: "NOR-flash-like" and the actual flags.

Feel free to close this...

pacucha42: Yes, unfortunately, `ESP_BLOCKDEV_FLAGS_INST_CONFIG_DEFAULT` doesn't provide a good default (`erase_before_write` can't be combined with and `and_type_write`). This is a relict from early stage of BDL development... let me see how to fix.

For the meantime, please use manual flags setup. Thanks

RathiSonika: Yes, I agree—I’ll make the update. Thanks for the review!

RathiSonika: Updated!

spi_nand_flash/include/esp_nand_blockdev.h
```diff
@@ -0,0 +1,295 @@
...
+/** @brief Check if a block is marked as bad
+ *
+ * Usage:
+ * @code
+ * esp_blockdev_cmd_arg_status_t cmd = { .num = block_num };
+ * esp_err_t ret = bdl->ops->ioctl(bdl, ESP_BLOCKDEV_CMD_IS_BAD_BLOCK, &cmd);
+ * bool is_bad = cmd.status;
+ * @endcode
+ */
+#define ESP_BLOCKDEV_CMD_IS_BAD_BLOCK             (ESP_BLOCKDEV_CMD_SYSTEM_BASE + 10)
```
haberturdeur: @pacucha42 we need to reserve these in the main esp-idf

pacucha42: Not sure - is there any harm, if we keep it there? ESP_BLOCKDEV_CMD_IS_BAD_BLOCK sounds general enough.

Anyway, @RathiSonika could you please provide more of the background why we need to provide such info through I/O control? This sounds more like a status check or possibly write attempt error.

pacucha42: Wait, I missed it's NOT `esp_blockdev.h `!!!! You are right, these definitions should be in there.

I'll make the IDF update and link it here. @RathiSonika please provide the backgrounds anyway, it is helpful for our users to see why we decided to go this way. Thank you

RathiSonika: Yes, as per the offline discussion, I will remove these definations from the header and these commands will bee added in esp_blockdev.h

To give an idea about why these ioctl commands are implemented: For NAND flash, bad block handling must be supported explicitly. Before erase or program operations, software needs to check whether a block is bad, and it must also be able to mark a block as bad when needed.
To keep the BDL interface minimal, we expose NAND-specific operations through ioctl. Therefore, we added the following commands required by WL BDL: ESP_BLOCKDEV_CMD_MARK_BAD_BLOCK, ESP_BLOCKDEV_CMD_IS_BAD_BLOCK, ESP_BLOCKDEV_CMD_IS_FREE_PAGE etc. and there are some diagnostics related commands as well to get ECC stats.
These ioctls allow WL BDL to handle NAND bad block management without extending the generic block device interface.

spi_nand_flash/include/esp_nand_blockdev.h
```diff
@@ -0,0 +1,295 @@
...
+
+/** @brief Get number of available logical sectors (WL layer only)
+ *
+ * Usage:
+ * @code
+ * uint32_t available_sectors;
+ * esp_err_t ret = wl_bdl->ops->ioctl(wl_bdl, ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS, &available_sectors);
+ * @endcode
+ */
+#define ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS    (ESP_BLOCKDEV_CMD_SYSTEM_BASE + 14)
```
pacucha42: Btw, I'm going to remove this command - available "sectors" (ie pages) can be easily calculated from device geometry like `available_sectors ≡ geometry.disk_size / geometry.read_size `

"Sector" is a poor-defined term and should be avoided as far as possible

pacucha42: (Hopefully) final setup:
- the NAND specific commands stay in the `spi_nand_flash` component
- NAND reserved range and all command IDs used are to be documented in `IDF/esp_blockdev.h`
- the above-mentioned documentation will be provided in  a propriatery code-comment format to allow pre-commit checkers for `IDF` x `spi_nand_flash` consistency verification
- `ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS` will be removed completely as the command is actually not needed => which implies the remaining command IDs are regenerated and need to be re-verified
- possible future WL implementations are expected to be `IDF external components` (the same arrangement we have for Dhara) - that guarantees we don't need NAND IOCTL definitions in the main IDF repo

@RathiSonika please, implement related changes, thanks

omnitex: ~~Wasn't `ESP_BLOCKDEV_CMD_GET_AVAILABLE_SECTORS` the command to expose the "logical number of sectors" -- the number after various reserved portions for e.g. Dhara's garbage collection etc.? So it's not replacable by the simple physical geometry calculation?~~

the number of available pages gets correctly propagated via the geometry, no need for an additional ioctl, resolved

RathiSonika: Updated

PR #607, opened 2025-11-11
Add support of user_data in read_cb (Fixes #604) (IEC-413)

### LINKED ISSUE #604: Feature Request: Add `void *user_data` Context to esp_delta_ota Callbacks (IEC-410)
## Description
In the upstream `detools` library, there is a `void*` context pointer that allows injecting user-defined data into the `read_cb`  function like `write_cb`.
However, this functionality appears to have been removed during the abstraction layer implementation in the ESP Delta OTA component:
[https://github.com/espressif/idf-extra-components/blob/86bd7d98a0f6678e51ed21bf7211be1d5f61691a/esp_delta_ota/include/esp_delta_ota.h#L24C1-L25C81](https://github.com/espressif/idf-extra-components/blob/86bd7d98a0f6678e51ed21bf7211be1d5f61691a/esp_delta_ota/include/esp_delta_ota.h#L24C1-L25C81)

## Use Case
I’m using an ESP device to update firmware on external MCUs.
To manage multiple update targets and custom file I/O operations, it’s necessary to pass context-specific data (e.g., file handles, device interfaces, metadata) into the read callbacks.
Without a `user_data` pointer, this becomes cumbersome and requires using global or static variables, which is not ideal in concurrent or multi-instance setups.

## Suggested Improvement
Add a `void *user_data` field to `esp_delta_ota_cfg_t` and pass it to both `read_cb` as of `write_cb` callbacks—similar to how detools handles it.
This change would preserve backward compatibility while greatly improving flexibility and modularity.

erwinschrodinger1:
# Checklist
- [x] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CmakeLists.txt#L8)
- [x] CI passing

# Change description
### Add support of user_data in read_cb
Added in reference to write callback.
Is backward compatible.
Tested with file read using FileIO in esp32s3.


esp_delta_ota/src/esp_delta_ota.c
```diff
@@ -57,11 +60,20 @@ static int esp_delta_ota_read_cb(void *arg_p, uint8_t *buf_p, size_t size)
...
-        return ESP_FAIL;
+    esp_err_t err = ESP_OK;
+    if (!handle->user_data) {
+        err = handle->read_cb(buf_p, size, handle->src_offset);
+        if (err != ESP_OK) {
+            ESP_LOGE(TAG, "Error in read_cb(): %s", esp_err_to_name(err));
+            return ESP_FAIL;
+        }
+    } else {
+        err = handle->read_cb_with_user_data(buf_p, size, handle->src_offset, handle->user_data);        if (err != ESP_OK) {
```
hrushikesh430:
```suggestion
        err = handle->read_cb_with_user_data(buf_p, size, handle->src_offset, handle->user_data);       
        if (err != ESP_OK) {
```

hrushikesh430:
@erwinschrodinger1 Could you please Bump the the component version and update the Changelog.md.

You can take reference from this [PR](https://github.com/espressif/idf-extra-components/commit/e26de5141a66755957e9afa3933b4f2e2ee0f395)

PR #623, opened 2025-11-24
feat(sh2lib): distinct recv and send funcs (IEC-429)

bryghtlabs-richard:
# Checklist

- [ ] CI passing

# Change description

Enable processing received HTTP client traffic before sending HTTP client traffic. This can reduce latency in a few cases:

* When HTTP response triggers queueing a new HTTP request
* When HTTP client wants to reset stream without resetting connection
*  When partial HTTP response is received, and HTTP client needs to send WINDOW_UPDATE

### DIFF
--- a/sh2lib/sh2lib.c
+++ b/sh2lib/sh2lib.c
@@ -343,6 +343,11 @@ int sh2lib_execute(struct sh2lib_handle *hd)
 
 int sh2lib_execute_recv(struct sh2lib_handle *hd)
 {
+    if (hd == NULL) {
+        ESP_LOGE(TAG, "[sh2-execute-recv] pointer to sh2lib handle cannot be NULL");
+        return -1;
+    }
+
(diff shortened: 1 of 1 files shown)


sh2lib/sh2lib.c
```diff
@@ -328,21 +328,39 @@ void sh2lib_free(struct sh2lib_handle *hd)
...
-        ESP_LOGE(TAG, "[sh2-execute] HTTP2 session recv failed %d", ret);
         return ret;
     }
 
     return 0;
 }
 
+int sh2lib_execute_recv(struct sh2lib_handle *hd)
+{
+    int ret = nghttp2_session_recv(hd->http2_sess);
```
mahavirj, line 351: Suggest adding input param validation check - some of the other public APIs are also missing out on those. Please see if you could update in same PR.

PR #626, opened 2025-11-25
fix(libpng): Update to version 1.6.51

tore-espressif:
This version includes important security fixes:
- CVE-2025-64720
- CVE-2025-65018
- CVE-2025-64505
- CVE-2025-64506


libpng/sbom_libpng.yml
```diff
@@ -1,7 +1,16 @@
 name: libpng
-version: 1.6.39
-cpe: cpe:2.3:a:libpng:libpng:{}:*:*:*:*:*:*:*
-supplier: 'Organization: libpng'
+version: 1.6.51
+cpe: cpe:2.3:a:pnggroup:libpng:{}:*:*:*:*:*:*:*
```
fhrbata: Hello,

I'm trying to determine the origin of the `cpe:2.3:a:pnggroup:libpng` CPE. I couldn't locate it on the NVD CPE search page at https://nvd.nist.gov/products/cpe/search. I believe we should retain `cpe:2.3:a:libpng:libpng`, as failing to do so might result in missing new CVE reports unless they are specifically filed under `cpe:2.3:a:pnggroup:libpng`. Please note that the `cve` field can accommodate multiple CPE entries, so we can include both CPEs if necessary.

Thank you

tore-espressif: @fhrbata Thank you for having a look. I based the change on this https://cveawg.mitre.org/api/cve/CVE-2025-65018

image

Could you please doubldcheck and confirm that we should rever to 'vendor = libpng'?

fhrbata: Yes, it's possible that the new CVEs will have the `cpe:2.3:a:pnggroup:libpng` assigned by NVD instead of `cpe:2.3:a:libpng:libpng`, but at this point I don't see it. We will know for sure once the CVEs are analyzed by NVD and the CPE is actually assigned. For example this mbed_tls CVE is already analyzed https://services.nvd.nist.gov/rest/json/cves/2.0?cveId=CVE-2025-59438 and you can see

```json
        "configurations": [
          {
            "nodes": [
              {
                "operator": "OR",
                "negate": false,
                "cpeMatch": [
                  {
                    "vulnerable": true,
                    "criteria": "cpe:2.3:a:arm:mbed_tls:*:*:*:*:*:*:*:*",
                    "versionEndExcluding": "3.6.5",
                    "matchCriteriaId": "D9DA7861-39B5-40B3-BF4F-6A79ED188844"
                  }
                ]
              }
            ]
          }
        ],
```

The `configurations` entry is what esp-idf-sbom uses when scanning for vulnerabilities. Note that MITRE maintains only the CVE identifier, while the NVD [enriches](https://nvd.nist.gov/general/FAQ-Sections/General-FAQs?utm_source=chatgpt.com#faqLink0) it with additional information, such as which versions of a product are affected.

Since we don't know the CPE, I think it would be best to keep both of them in the sbom manifest for now.
```yaml
cpe:
    - cpe:2.3:a:libpng:libpng:{}:*:*:*:*:*:*:*
    - cpe:2.3:a:pnggroup:libpng:{}:*:*:*:*:*:*:*
```

Same as we have e.g. for [cJSON](https://github.com/espressif/idf-extra-components/blob/master/cjson/sbom_cJSON.yml)

tore-espressif: keeping both sounds the best to me! fixed, thanks!

PR #633, opened 2025-11-28
feat(qrcode): Add user_data for qrcode (IEC-446)

espressif2022:
# Checklist

- [ ] CI passing

# Change description
Add user_data for qrcode.

### DIFF
--- a/qrcode/include/qrcode.h
+++ b/qrcode/include/qrcode.h
@@ -21,9 +21,13 @@ typedef const uint8_t *esp_qrcode_handle_t;
   * @brief  QR Code configuration options
   */
 typedef struct {
-    void (*display_func)(esp_qrcode_handle_t qrcode);   /**< Function called for displaying the QR Code after encoding is complete */
+    union {
+        void (*display_func)(esp_qrcode_handle_t qrcode);                    /**< Function called for displaying the QR Code (without user_data, for backward compatibility) */
+        void (*display_func_with_cb)(esp_qrcode_handle_t qrcode, void *user_data);   /**< Function called for displaying the QR Code with user_data callback */
+    };                                                                        /**< Display function union for QR Code display */
     int max_qrcode_version;                             /**< Max QR Code Version to be used. Range: 2 - 40 */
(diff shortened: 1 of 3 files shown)


espressif2022:
@vikramdattu 

I'd like to add a `user_data` field to the qrcode repository.

This library is used a lot by rainmaker. Could you please check if this will have any impact?

qrcode/include/qrcode.h
```diff
@@ -21,9 +21,10 @@ typedef const uint8_t *esp_qrcode_handle_t;
   * @brief  QR Code configuration options
   */
 typedef struct {
-    void (*display_func)(esp_qrcode_handle_t qrcode);   /**< Function called for displaying the QR Code after encoding is complete */
+    void (*display_func)(esp_qrcode_handle_t qrcode, void *user_data);   /**< Function called for displaying the QR Code after encoding is complete */
```
mahavirj: This is a breaking change. Maybe we could use unnamed union to keep backward compatibility:

```
union {
        display_func()
        display_func_with_cb()
}
```

igrr: It's consistent with the version bump from "0.1.0~2" to "0.2.0", though?

espressif2022: So in this case, should I maintain backward compatibility instead of directly bumping the version from 0.1.0 to 0.2.0? I’ll update it accordingly.

mahavirj: > It's consistent with the version bump from "0.1.0~2" to "0.2.0", though?

That will be a minor version update. If we would like to break the interface then we should bump up the version to 1.0.0?

espressif2022: > > It's consistent with the version bump from "0.1.0~2" to "0.2.0", though?
> 
> That will be a minor version update. If we would like to break the interface then we should bump up the version to 1.0.0?

If I use a union to ensure backward compatibility, is it still appropriate to keep using version 0.2.0?
I just tested it, and the existing RainMaker can still remain compatible even if it pulls this version.

mahavirj: Yes. Current changes LGTM!

PR #635, opened 2025-12-01
thorvg: Fix compile issue and bump version to v0.15.16

suda-morris:
## Description

Fixes ThorVG component build failure when using ESP-IDF v6 due to incompatibility between ESP-IDF's response file syntax and Meson's handling of compiler flags.

## Problem

Starting with ESP-IDF v6 (commit [0c1d917](https://github.com/espressif/esp-idf/commit/0c1d917f782b03030ada09c8ad9b618acc7a329b)), the build system uses `@"file"` syntax (with quotes) for response files in `CMAKE_*_FLAGS` variables. This causes the ThorVG example build to fail with:

```
ERROR: Compiler cannot compile programs.
xtensa-esp-elf-g++: error: @"/path/to/cxxflags": linker input file not found
```

### Root Cause

The issue stems from different handling of response file syntax across build systems:

| Tool | Behavior with `@"file"` (with quotes) | Behavior with `@file` (without quotes) |
|------|---------------------------------------|----------------------------------------|
| **CMake** | ✅ Expands internally before invoking compiler | ✅ Expands internally |
| **Meson** | ❌ Passes literally to compiler (not recognized) | ✅ Passes to compiler (GCC handles) |
| **GCC** | ❌ Treats as linker input file (error) | ✅ Reads as response file |

When our CMakeLists.txt generates a Meson cross file with `@"file"` syntax, Meson passes it literally to GCC, which fails because GCC only understands `@file` without quotes.

## Solution

Detect when `CMAKE_CXX_FLAGS` and `CMAKE_EXE_LINKER_FLAGS` contain `@"file"` syntax, read the actual flags from those response files, and expand them as individual list items in the Meson cross file.

**Why not just remove the quotes?**
- ESP-IDF supports Windows where project paths may contain spaces
- GCC's `@file` syntax doesn't support quoting the file path itself
- Expanding the contents ensures compatibility across all platforms

## Changes

- Modified `CMakeLists.txt` to detect and expand ESP-IDF response files before passing flags to Meson
- Added detailed comments explaining the response file handling differences

## Testing

- ✅ Builds successfully with ESP-IDF v6.1+ on `esp32s3` target
- ✅ ThorVG library compiles without errors
- ✅ Example application links and generates binary successfully

### Build Output
```
-- CMAKE_CXX_FLAGS: -mlongcalls;-fno-builtin-memcpy;-fno-builtin-memset;-fno-builtin-bzero;-mdisable-hardware-atomics
...
[50/50] Linking static target src/libthorvg.a
Project build complete.
```

## Related

- ESP-IDF issue: Response file syntax introduced in commit 0c1d917f782b03030ada09c8ad9b618acc7a329b
- Affects: ESP-IDF v6.0 and later

### DIFF
--- a/thorvg/idf_component.yml
+++ b/thorvg/idf_component.yml
@@ -1,9 +1,9 @@
-version: "0.15.15"
+version: "0.15.16"
 description: "ThorVG is an open-source graphics library designed for creating vector-based scenes and animations"
 url: https://github.com/espressif/idf-extra-components/tree/master/thorvg
 repository: "https://github.com/espressif/idf-extra-components.git"
 documentation: "https://github.com/thorvg/thorvg/blob/main/README.md"
-issues: "https://github.com/espressif/idf-extra-components/issues" # URL of the issue tracker
+issues: "https://github.com/espressif/idf-extra-components/issues"
 dependencies:
   idf: ">=5.1"
 sbom:
--- a/thorvg/sbom_thorvg.yml
+++ b/thorvg/sbom_thorvg.yml
@@ -1,7 +1,7 @@
 name: thorvg
-version: 0.15.15
+version: 0.15.16
 cpe: cpe:2.3:a:thorvg:thorvg:{}:*:*:*:*:*:*:*
 supplier: 'Organization: thorvg <https://github.com/thorvg/thorvg>'
 description: ThorVG is an open-source graphics library designed for creating vector-based scenes and animations.
 url: https://github.com/thorvg/thorvg
-hash: 651c71537e74766e3ef123ecc64f620a903ec8d4
+hash: e15069de7afcc5e853edf1561e69d9b8383e2c6c


thorvg/CMakeLists.txt
```diff
@@ -52,13 +52,37 @@ idf_component_register(
 
 
 # Convert argument strings to lists, for easier processing
-set(linker_args ${CMAKE_EXE_LINKER_FLAGS})
-separate_arguments(linker_args)
-list(FILTER linker_args EXCLUDE REGEX "^$")
+# ESP-IDF v6 uses @"file" syntax for response files, but Meson doesn't support this.
```
mahavirj, line 55: Is this specific to CMake version?

suda-morris: not specific to cmake version.

igrr: I think response files are almost always by CMake used on Windows (even in old IDF versions) due to the limit on the command line length.

PR #646, opened 2025-12-17
feat(argtable3): allow usage alongside with IDF's copy of argtable3

### LINKED ISSUE #539: Can argtable3 work with anything less than idf 6? I (along with many people) will want to be using it with 5.5 since that's the stable release. (IEC-350)
### Answers checklist.

- [x] I have read the documentation of the component in question and the issue is not addressed there.

### General issue report

There is no documentation for the component.

igrr:
argtable3 component now checks if IDF contains its own copy of argtable3, and excludes that one from the build if so.

A test app is added to verify this.

Known issues:
1. The generated SBOM will still acknowledge the existence of IDF's copy of argtable3, even though it's not used.
2. IDF's console component still provides argtable3 headers, so this solution only works until argtable3 API changes.

Closes #539

### DIFF
--- a/.idf_build_apps.toml
+++ b/.idf_build_apps.toml
@@ -42,5 +42,5 @@ check_warnings = true
 
 # build related options
 build_dir = "build_@t_@w"
-config = "sdkconfig.ci"
+config = ["sdkconfig.ci","sdkconfig.ci.*="]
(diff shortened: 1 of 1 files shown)


SoucheSouche:
Thanks for taking care of it @igrr, LGTM

PR #660, opened 2026-01-12
fix(spi_nand_flash): fix compilation error with idf v6.0 and above

RathiSonika:
# Change description
* Fixed a compilation error caused by the missing freertos/FreeRTOS.h header when building with ESP-IDF v6.0 and later.

* Starting with ESP-IDF v6.0, esp_driver_spi requires data length alignment in addition to buffer alignment for both RX and TX operations. Updated spi_nand_oper.c to validate length alignment at the lower-level functions and adjust the length as needed.

* In the spi_nand_flash driver, marking the bad_block and used_page status previously involved writing 2 bytes, which was not DMA-aligned (DMA requires 4-byte alignment). The code has been updated to read and write these OOB markers together instead of handling them individually, avoiding length alignment issues in the SPI driver.

* In the case of TX operations, the SPI driver validates both buffer alignment and TX length alignment. When writing markers (4 bytes), if the alignment length is 64 bytes, this could result in writing 64 bytes instead, which is not recommended. To avoid this behavior, the NAND flash driver allows the SPI driver to handle alignment/buffer allocation when the length is not aligned.

Accordingly, in spi_nand_oper.c, within the spi_nand_program_load function, SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL flag is enabled only when the data length is already aligned to the DMA/L1 cache requirements.

### DIFF
--- a/esp_cli_commands/.build-test-rules.yml
+++ b/esp_cli_commands/.build-test-rules.yml
@@ -1,35 +1,35 @@
 esp_cli_commands/test_apps:
   enable:
-    - if: (IDF_TARGET in ["esp32", "linux"])
+    - if: IDF_TARGET in ["esp32", "linux"]
   disable:
     - if: IDF_VERSION_MAJOR <= 5 and IDF_VERSION_MINOR < 3
 
 esp_cli_commands/examples/command_set:
   enable:
-    - if: (IDF_TARGET == "linux")
+    - if: IDF_TARGET == "linux"
   disable:
     - if: IDF_VERSION_MAJOR <= 5 and IDF_VERSION_MINOR < 3
 
 esp_cli_commands/examples/command_with_arg:
   enable:
-    - if: (IDF_TARGET == "linux")
+    - if: IDF_TARGET == "linux"
   disable:
     - if: IDF_VERSION_MAJOR <= 5 and IDF_VERSION_MINOR < 3
 
 esp_cli_commands/examples/dynamic_registration:
   enable:
-    - if: (IDF_TARGET == "linux")
+    - if: IDF_TARGET == "linux"
   disable:
     - if: IDF_VERSION_MAJOR <= 5 and IDF_VERSION_MINOR < 3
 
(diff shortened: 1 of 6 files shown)


spi_nand_flash/src/nand_impl.c
```diff
@@ -95,18 +95,19 @@ static uint16_t get_column_address(spi_nand_flash_device_t *handle, uint32_t blo
 esp_err_t nand_is_bad(spi_nand_flash_device_t *handle, uint32_t block, bool *is_bad_status)
 {
     uint32_t first_block_page = block * (1 << handle->chip.log2_ppb);
-    uint16_t bad_block_indicator;
+    uint32_t markers;  // Combined: [bad_block_marker:16][page_used_marker:16]
```
omnitex: Any reason why here and in `nand_is_free()` the `markers` are kept as `uint32_t` instead of `uint8_t[]` as well?

RathiSonika: Sorry, my bad. It’s fixed now. Updated markers to uint8_t[].

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -237,17 +249,13 @@ static esp_err_t spi_nand_fast_read(spi_nand_flash_device_t *handle, uint8_t *da
...
-    }
-#else
-    if (!half_duplex) {
-        memcpy(data, data_read + 1, length);
+    if (aligned_len != length) {
+        if (!half_duplex) {
+            memcpy(data, data_read + 1, length);
+        } else {
+            memcpy(data, data_read, length);
+        }
```
pacucha42, line 257: Nitpicking: this fix looks better than the previous code but it could be "phrased" even nicer way. Eg:

```suggestion
    if (aligned_len != length) {
       uint32_t src_off = half_duplex ? data_read : (data_read+1);
       memcpy(data, src_off, length);
    }
```
Pure cosmetics, feel free to ignore 🙂

PR #662, opened 2026-01-13
feat(esp_cli_commands): Add examples in the component

SoucheSouche:
# Checklist

- [x] CI passing

# Change description
Added example directory in esp_cli_commands to provide how-to support for users.


esp_cli_commands/examples/static_registration/main/static_registration_main.c
```diff
@@ -0,0 +1,84 @@
...
+
+// Glossary callback signature: const char *(*)(void *)
+static const char *hello_cmd_glossary_cb(void *ctx)
+{
+    (void)ctx;
+    return "This command prints a hello message for demonstration purposes.";
+}
+
+// Static registration of the hello command with all fields
+ESP_CLI_COMMAND_REGISTER(
```
ESP-Marius, line 35: Since this is our simple example we could consider showing the simplest `ESP_CLI_COMMAND_REGISTER` and remove the unneeded config params.

esp_cli_commands/examples/static_registration/main/static_registration_main.c
```diff
@@ -0,0 +1,84 @@
...
+    hello,                  // Command name
+    example,                // Command group
+    "Prints a hello message", // Help string
+    hello_cmd_handler,      // Handler function
+    &hello_cmd_ctx,         // Context pointer (must be address of static object)
+    hello_cmd_hint_cb,      // Hint callback
+    hello_cmd_glossary_cb   // Glossary callback
+);
+
+void app_main(void)
```
ESP-Marius, line 45: Do you think it is worth adding a simple pytest test file for all of these examples and run them on linux?

esp_cli_commands/examples/math_op_static/main/math_op_static.c
```diff
@@ -0,0 +1,118 @@
+#include <stdio.h>
+#include <string.h>
+#include "esp_cli_commands.h"
+
+// Context for the command (must be constant for static registration)
+static int math_op_cmd_ctx = 0;
+
+// Handler function signature must match: int (*)(void *, esp_cli_commands_exec_arg_t *, int, char **)
+static int math_op_cmd_handler(void *ctx, esp_cli_commands_exec_arg_t *exec_arg, int argc, char **argv)
```
ESP-Marius: Is having an example which integrates with argtable for arguments also something that would be nice?

esp_cli_commands/examples/debug_and_unregister_dynamic/README.md
```diff
@@ -0,0 +1,3 @@
+# debug_and_unregister_dynamic
```
ESP-Marius: This example is mostly meant to demostrate dynamic registration right? Consider if we should just call it "Dynamic registration" instead so it is easier for users to find the functionality they are looking to learn.

esp_cli_commands/examples/math_op_static/README.md
```diff
@@ -0,0 +1,3 @@
+# math_op_static
```
ESP-Marius: Same here as for `debug_and_unreg..`. I think users will be looking for examples showing a certain feature of esp-cli-cmds that they want to learn about, so `math_op_static` might not be very explainatory.

esp_cli_commands/examples/command_with_arg/main/command_with_arg_main.c
```diff
@@ -0,0 +1,118 @@
...
+
+// Handler function signature must match: int (*)(void *, esp_cli_commands_exec_arg_t *, int, char **)
+static int math_op_cmd_handler(void *ctx, esp_cli_commands_exec_arg_t *exec_arg, int argc, char **argv)
+{
+    (void)ctx;
+    (void)exec_arg;
+    if (argc != 4) {
+        int color = 0;
+        bool bold = false;
+        const char *hint = esp_cli_commands_get_hint(NULL, "math_op", &color, &bold);
```
ESP-Marius, line 13: Should all examples be using the `write_func` to encourage developers to not put printfs into their cmds?

esp_cli_commands/examples/dynamic_registration/main/dynamic_registration_main.c
```diff
@@ -0,0 +1,139 @@
...
+
+void app_main(void)
+{
+    printf("esp_cli_commands debug_and_unregister_dynamic example started.\n");
+
+    int ret = -1;
+    esp_err_t err = ESP_FAIL;
+
+    // Dynamically register debug command
+    esp_cli_command_t debug_cmd;
```
ESP-Marius: Initializing like this will break if we add a new member (as it will have a random value from the stack unless you expliciltly init it below)

esp_cli_commands/examples/dynamic_registration/main/dynamic_registration_main.c
```diff
@@ -0,0 +1,139 @@
...
+    debug_cmd.group = "example";
+    debug_cmd.help = "Prints debug information";
+    debug_cmd.func = debug_cmd_handler;
+    debug_cmd.func_ctx = &debug_cmd_ctx;
+    debug_cmd.hint_cb = debug_cmd_hint_cb;
+    debug_cmd.glossary_cb = debug_cmd_glossary_cb;
+    ESP_ERROR_CHECK(esp_cli_commands_register_cmd(&debug_cmd));
+
+    // Dynamically register unregister command
+    esp_cli_command_t unregister_cmd;
```
ESP-Marius: Same as above

esp_cli_commands/examples/static_registration/main/static_registration_main.c
SoucheSouche: i can add a comment next to the optional ones, but I would rather keep it all so at least user have the complete picture.

esp_cli_commands/examples/debug_and_unregister_dynamic/README.md
SoucheSouche: updated

esp_cli_commands/examples/math_op_static/main/math_op_static.c
SoucheSouche: added an example for argtable3 usage

esp_cli_commands/examples/command_with_arg/main/command_with_arg_main.c
SoucheSouche: nice catch, thanks for noticing :)

esp_cli_commands/examples/static_registration/main/static_registration_main.c
SoucheSouche: pytest file added

esp_cli_commands/examples/command_set/main/command_set_main.c
```diff
@@ -0,0 +1,109 @@
...
+#include <string.h>
+#include "esp_cli_commands.h"
+#include "command_utils.h"
+
+static int cmd_a_handler(void *context, esp_cli_commands_exec_arg_t *cmd_arg, int argc, char **argv)
+{
+    (void)context; (void)argc; (void)argv;
+    const char *msg = "cmd_a executed\n";
+    const size_t msg_len = strlen(msg);
+    const int nwrite = cmd_arg->write_func(cmd_arg->out_fd, msg, msg_len);
```
o-marshmallow, line 24: Nitpick, you can use an array instead, so that you don't need to calculate the length of the string:
```
const char msg[] = "cmd_a executed\n";
const int nwrite = cmd_arg->write_func(cmd_arg->out_fd, msg, sizeof(msg) - 1);
```

esp_cli_commands/examples/command_with_arg/main/command_with_arg_main.c
```diff
@@ -0,0 +1,126 @@
...
+    esp_cli_command_t *cmd = esp_cli_commands_find_command(NULL, "math_op");
+    if (cmd) {
+        printf("Found command: %s\n", cmd->name);
+    } else {
+        printf("Command 'math_op' not found!\n");
+    }
+
+    // Execute the 'math_op' command programmatically (example: add 3 5)
+    ret = -1;
+    err = esp_cli_commands_execute("math_op add 3 5", &ret, NULL, &cmd_args);
```
o-marshmallow, line 96: Since the commands receive the parameters as `argc` and `argv`, is there a way, like `exec` function on Linux, to pass the array of parameters directly?
So something like:
```
    err = esp_cli_commands_execute_args("math_op",
     { "math_op", "add", "3", "5" }
     , &ret, NULL, &cmd_args);
```
In this example it doesn't really make sense, but if you have strings that are not constants, that would allow you to exec a command without the need to format it first with `sprints`, `strcat`, etc...

esp_cli_commands/examples/dynamic_registration/main/dynamic_registration_main.c
```diff
@@ -0,0 +1,146 @@
...
+        .write_func = write,
+        .dynamic_ctx = NULL
+    };
+
+    int ret = -1;
+    esp_err_t err = ESP_FAIL;
+
+    // Dynamically register debug command
+    esp_cli_command_t debug_cmd = {
+        debug_cmd.name = "debug",
```
o-marshmallow: If  `debug_cmd` is not a field of the `esp_cli_command_t` structure, you can remove it from this declaration:
```
    esp_cli_command_t debug_cmd = {
        .name = "debug",
        .group = "example", 
        ...
    }
```

esp_cli_commands/examples/static_registration/main/static_registration_main.c
```diff
@@ -0,0 +1,92 @@
+
+#include <stdio.h>
+#include "esp_cli_commands.h"
+#include "command_utils.h"
+
+
+// Context for the command (must be constant for static registration)
```
o-marshmallow: I don't think you meant `constant`, you mean that its address must be resolvable at compile/link time?

esp_cli_commands/examples/command_set/main/command_set_main.c
```diff
@@ -0,0 +1,109 @@
...
+    // create command sets. One with command name, another with group name
+    const char *cmd_set_a[] = { "cmd_a" };
+    const char *cmd_set_b[] = { "group_b" };
+    esp_cli_command_set_handle_t set_a = ESP_CLI_COMMANDS_CREATE_CMD_SET(cmd_set_a, ESP_CLI_COMMAND_FIELD_ACCESSOR(name));
+    esp_cli_command_set_handle_t set_b = ESP_CLI_COMMANDS_CREATE_CMD_SET(cmd_set_b, ESP_CLI_COMMAND_FIELD_ACCESSOR(group));
+
+    // Test all combinations
+    int ret = -1;
+    printf("-- Executing cmd_a with set_a (should succeed) --\n");
+    ESP_ERROR_CHECK(esp_cli_commands_execute("cmd_a", &ret, set_a, &cmd_args));
```
o-marshmallow, line 79: Since in most cases we will perform filtering (sets) on the group, would it make sense to have an `execute` function that also takes a group?
```
esp_cli_commands_execute_group("cmd_a", &ret, "group_b", &cmd_args)
```

o-marshmallow (approved):
Changes LGTM! I left a few questions and nitckpicks, feel free to ignore them

esp_cli_commands/examples/command_set/main/command_set_main.c
SoucheSouche: The command_set can be made of groups or command names or both. The API provides a generic way to address filtering that covers (hopefully) all use cases.

I won't add another execute function that handles only 1 use case based on assumptions on how users will interact with the API. I would rather wait and see.

esp_cli_commands/examples/dynamic_registration/main/dynamic_registration_main.c
SoucheSouche: thanks for noticing, it was left over from a previous version of the example.

esp_cli_commands/examples/command_with_arg/main/command_with_arg_main.c
SoucheSouche: that's a good point. I will not handle this in the PR but I will create a ticket to follow up on this.

esp_cli_commands/examples/command_set/main/command_set_main.c
SoucheSouche: updated.

PR #668, opened 2026-01-21
feat(argtable3): dump version after changes in component CMakeLists.txt

SoucheSouche:
# Checklist

- [ ] CI passing

# Change description

### DIFF
--- a/argtable3/idf_component.yml
+++ b/argtable3/idf_component.yml
@@ -1,4 +1,4 @@
-version: "3.3.1"
+version: "3.3.1~1"
 description: "Argtable3 - GNU-style command-line option parsing C library"
 url: https://github.com/espressif/idf-extra-components/tree/master/argtable3
 documentation: https://www.argtable.org/docs/


argtable3/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.3.1"
+version: "3.3.2"
```
igrr: Argtable's upstream version is still 3.3.1, right?
https://github.com/espressif/idf-extra-components/blob/29b1acdbcef121a2b807f5c0e52b44210d43838a/argtable3/sbom_argtable3.yml#L2
In that case, it should be

```suggestion
version: "3.3.1~1"
```

argtable3/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.3.1"
+version: "3.3.1-1"
```
igrr:
```suggestion
version: "3.3.1~1"
```

(the intention is to set the `revision` field, not `pre-release` one, referring to the [docs](https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/versioning.html#versioning-scheme))

SoucheSouche: that makes more sense, thanks for clarifying.

igrr (approved):

PR #675, opened 2026-01-29
fix: added conftest file to apply ignore (IEC-482)

horw:
Somehow ignore doesn't work; this conftest is supposed to handle ignoring.


SoucheSouche:
@horw 

It seems that in some cases (e.g, https://github.com/espressif/idf-extra-components/actions/runs/19633490516/job/56218832804) the tests marked as skipped are ignored properly.

The changes seem to have fixed the issue though.

PR #687, opened 2026-02-19
fix(esp_cli): update the call to on_exit in esp_cli()

SoucheSouche:
# Checklist

- [x] CI passing

# Change description
Make sure the on_exit function is called before the semaphore is released to make sure any operation done in the on_exit callback will be done before the esp_cli_stop returns.


ESP-Marius (approved):
LGTM

PR #690, opened 2026-02-23
fix(libsodium): improve esp port layer

Ashish285:
# Change description
This PR fixes a build issue reported in https://github.com/espressif/esp-homekit-sdk/issues/150.

* `libsodium` marks arguments as [nonnull](https://github.com/jedisct1/libsodium/blob/84e25e8f813323cc414b47b9cc96ff5ed498b0f3/src/libsodium/include/sodium/crypto_hash_sha256.h#L38). This creates build warning where `non null` argument is checked against `NULL`
* Adds upon the port headers to support mbedTLS/PSA hash contexts. This allows libsodium to use hardware implementation for the SHA operations.
* Removes unconditional inclusion of internal hash implementation. This always used the internal implementation even though the `LIBSODIUM_USE_MBEDTLS_SHA` was set.

### DIFF
--- a/.github/workflows/deploy_gh_pages.yml
+++ b/.github/workflows/deploy_gh_pages.yml
@@ -52,7 +52,9 @@ jobs:
           args: >-
             --no-progress
             --include-fragments
-            './docs_build_output/**/*.html'
+            --root-dir docs_build_output
+            --exclude-path '.*/404\.html$'
+            docs_build_output/**/*.html
           # Fail the action if broken links are found
           fail: true
 


mahavirj:
@Ashish285 Please bump the component version for this fix

Ashish285:
> @Ashish285 Please bump the component version for this fix

Sure, updated the component version.

libsodium/CMakeLists.txt
```diff
@@ -120,14 +118,16 @@ if(CONFIG_LIBSODIUM_USE_MBEDTLS_SHA)
     list(APPEND srcs
         "port/crypto_hash_mbedtls/crypto_hash_sha256_mbedtls.c"
         "port/crypto_hash_mbedtls/crypto_hash_sha512_mbedtls.c")
+    set(include_dirs port_include ${SRC}/include)
```
mahavirj, line 121: Include order should be optional if you are already injecting in the includes using `SHELL-include` construct?

libsodium/port_include/sodium.h
```diff
@@ -0,0 +1,22 @@
...
+* Angle brackets search -I paths directly, so port_include/sodium/
+* is found before the original. Once included, their guards are set,
+* so when the original sodium.h tries to include them via relative
+* double-quote paths, they are silently skipped. */
+#include <sodium/crypto_hash_sha256.h>
+#include <sodium/crypto_hash_sha512.h>
+
+/* Delegate to the original sodium.h. #include_next skips port_include/
+* and finds the next sodium.h in the search path. */
+#include_next "sodium.h"
```
mahavirj, line 20: Probably `#include <sodium.h>` should also work - it will try to find include from next directory that comes in the sequence

libsodium/port_include/sodium/crypto_hash_sha512.h
```diff
@@ -0,0 +1,66 @@
...
+extern "C" {
+#endif
+
+typedef struct crypto_hash_sha512_state {
+#ifdef MBEDTLS_PSA_CRYPTO
+    psa_hash_operation_t _psa_op;
+#else
+    mbedtls_sha512_context ctx;
+#endif /* MBEDTLS_PSA_CRYPTO */
+} crypto_hash_sha512_state;
```
mahavirj, line 34: Needs some static assert to confirm that indeed this struct is being used - maybe in test app?

libsodium/Kconfig
mahavirj: This can be removed now?

libsodium/port_include/sodium.h
Ashish285: Yes, but it probably seems clearer in intention using `#include_next` as it explicitly tells that we are including next header. Maybe it will be slightly harder to follow if `sodium.h` included `sodium.h`.

libsodium/CMakeLists.txt
Ashish285: I think they are still required as the `SHELL:-include` is private for the libsodium. Any external component using the libsodium headers will still need this include order to include in the port header files instead of the internal headers.

libsodium/Kconfig
Ashish285: This still is an option for the user to use mbedTLS implementation or not. If there is no mbedTLS use case in the user binary, this will be more efficient. What do you think?

libsodium/port_include/sodium/crypto_hash_sha512.h
Ashish285: We do have some tests in the [libsodium test suite](https://github.com/jedisct1/libsodium/blob/fb4533b0a941b3a5b1db5687d1b008a5853d1f29/test/default/hash.c#L43) itself that test the size of the structs.

I have added some static asserts check in the `test_sodium.c` as well.

libsodium/Kconfig
mahavirj: I meant the text that was referring to hardware SHA limitation from this config

Ashish285: Ah yes, that's removed now. Thanks!

PR #696, opened 2026-03-03
ci(github): Fix missing wheel issue at dl.espressif.com (IEC-495)

dobairoland:
- We recently cleaned dl.espressif.com from unnecessary wheels. This caused https://github.com/espressif/idf-extra-components/actions/runs/22621525426/job/65549889176?pr=695
- The submitted fix should fix the issue and explain the pitfall in comments.


dobairoland:
@hfudev @jakub-kocka PTAL

PR #698, opened 2026-03-05
change(freetype): exclude cve-2026-23865

Ashish285:
# Change description
`freetype` was recently update to v2.14.2 with this [PR](https://github.com/espressif/idf-extra-components/pull/694), however even after the update, `esp-idf-sbom` still mentions [CVE-2026-23865](https://nvd.nist.gov/vuln/detail/CVE-2026-23865) (The CVE description mentions `Freetype`).

This PR adds `CVE-2026-23865` to the exclude list as it is fixed in the current version.

### DIFF
--- a/freetype/sbom_freetype.yml
+++ b/freetype/sbom_freetype.yml
@@ -5,3 +5,6 @@ supplier: 'Organization: freetype <https://www.freetype.org>'
 description: FreeType is a software font engine
 url: https://github.com/freetype/freetype
 hash: f4205da14867c5387cd6a329b90ee10a6df6eeff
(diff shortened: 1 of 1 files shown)


Ashish285:
> The reason it's still reported is that the CVE has not yet been processed in NVD and it does not have a CPE assigned. The extended scan looks for keywords, like FreeType, in the description, and that the reason it's still reported. Adding the CVE into the exclude list is the proper way how to handle this situation, until CPE is assigned to CVE by NVD. Thank you

Ok, thanks for letting me know. Updated the description. Thanks!

Issue #684, opened 2026-02-12
SPI NAND program load may fail with non-DMA-capable buffer (IEC-488)


EeeeBin:
### Answers checklist.

- [x] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [x] I am using target and esp-idf version as defined in component's idf_component.yml
- [x] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

spi_nand_flash

### ESP-IDF version.

v5.5.1

### Development Kit.

ESP32P4

### Used Component version.

v0.17.0

### More Information.

Run the nand_flash + FatFs example. spi_nand_program_load() may fail because the provided data buffer can reside in non-DMA-capable memory (and/or be not 64-byte aligned). The current implementation conditionally sets SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL based only on length alignment, without validating whether the buffer address and memory capabilities satisfy SPI DMA requirements.

When the data buffer is not DMA-capable or properly aligned, the SPI driver rejects the transaction, causing the NAND program operation to fail.
```
esp_err_t spi_nand_program_load(spi_nand_flash_device_t *handle, const uint8_t *data, uint16_t column, uint16_t length)
{
    ESP_LOGI("nand", "spi_nand_program_load: column=%"PRIu16", length=%"PRIu16, column, length);
    ESP_LOGI("nand", "data: %p, align 64 ? %d, dma capable ? %d", data, (uintptr_t)data % 64 == 0, esp_ptr_dma_capable(data));
    uint8_t cmd = CMD_PROGRAM_LOAD;
    uint32_t spi_flags = 0;
    if (handle->config.io_mode == SPI_NAND_IO_MODE_QOUT || handle->config.io_mode == SPI_NAND_IO_MODE_QIO) {
        cmd = CMD_PROGRAM_LOAD_X4;
        spi_flags = SPI_TRANS_MODE_QIO;
    }

    const uint8_t *data_write = data;
    uint16_t data_write_len = length;
    // memcpy(handle->temp_buffer, data_write, length);
    spi_nand_transaction_t t = {
        .command = cmd,
        .address_bytes = 2,
        .address = column,
        .mosi_len = data_write_len,
        .mosi_data = data_write,
        .flags = spi_flags,
    };

[...]

PR #700, opened 2026-03-09
fix(spi_nand_flash): spi_nand_program_load fails if buffer is not DMA aligned

RathiSonika:
# Change description

Fix failure in spi_nand_program_load when the input buffer is not DMA-aligned.

Mainly it handles-
- Page-sized writes : always length-aligned on all chips → uses temp_buffer bounce only when buffer isn't DMA-ok, avoids SPI driver malloc
- Small marker writes (infrequent, 4 bytes): unaligned on P4 → delegated to SPI driver, which is safe for such tiny allocations. On non-P4 chips, 4 bytes is aligned to 4, so even these use the direct/temp_buffer path

Closes https://github.com/espressif/idf-extra-components/issues/684


spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -102,17 +103,22 @@ esp_err_t spi_nand_read_page(spi_nand_flash_device_t *handle, uint32_t page)
...
 #else
     // For non-L1CACHE targets, use DMA alignment of 4 bytes
     alignment = 4;
 #endif
+    return alignment;
+}
+
+static uint16_t check_length_alignment(spi_nand_flash_device_t *handle, uint16_t length)
+{
+    uint16_t data_len = length;
```
pacucha42, line 120: Unused variable, please, remove

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -294,6 +300,28 @@ esp_err_t spi_nand_program_load(spi_nand_flash_device_t *handle, const uint8_t *
...
+
+    if (aligned_len == length) {
+        // Length is already DMA-aligned. If the buffer address is not DMA-capable
+        // or not aligned, bounce through temp_buffer which is pre-allocated as
+        // DMA-capable. We can safely set SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL since
+        // both buffer and length now satisfy DMA requirements.
+        if (!buf_dma_ok) {
+            memcpy(handle->temp_buffer, data, length);
+            data_write = handle->temp_buffer;
+        }
```
pacucha42:
```suggestion
    if (aligned_len == length) {
        // Length is already DMA-aligned. If the buffer address is not DMA-capable
        // or not aligned, bounce through temp_buffer which is pre-allocated as
        // DMA-capable. We can safely set SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL since
        // both buffer and length now satisfy DMA requirements.
        bool buf_dma_ok = esp_ptr_dma_capable(data) && (((uintptr_t)data % alignment) == 0);
        if (!buf_dma_ok) {
            memcpy(handle->temp_buffer, data, length);
            data_write = handle->temp_buffer;
        }
```

spi_nand_flash/src/spi_nand_oper.c
RathiSonika: This variable is used in the check_length_alignment() function.

spi_nand_flash/src/spi_nand_oper.c
```diff
@@ -294,6 +300,28 @@ esp_err_t spi_nand_program_load(spi_nand_flash_device_t *handle, const uint8_t *
...
+    size_t alignment = get_alignment();
+
+    if (aligned_len == length) {
+        // Length is already DMA-aligned. If the buffer address is not DMA-capable
+        // or not aligned, bounce through temp_buffer which is pre-allocated as
+        // DMA-capable. We can safely set SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL since
+        // both buffer and length now satisfy DMA requirements.
+        bool buf_dma_ok = esp_ptr_dma_capable(data) && (((uintptr_t)data % alignment) == 0);
+        if (!buf_dma_ok) {
+            memcpy(handle->temp_buffer, data, length);
```
omnitex, line 314: wouldn't there be an issue with the `temp_buffer` which is allocated in `nand.c` as

`(*handle)->temp_buffer = heap_caps_malloc((*handle)->chip.page_size + 4, MALLOC_CAP_DMA | MALLOC_CAP_8BIT);`

with the length of `page_size + 4` ?

here we check the length of the to-be-programmed data is DMA aligned, and when the `data` buffer is not DMA capable, we use the `temp_buffer`, which we have guaranteed is DMA capable (thanks to `MALLOC_CAP_DMA`) but we don't guarantee the length to be aligned according to what `get_alignment()` returns?

does it matter in any way that the `temp_buffer` has fixed length like this?

I need to look into DMA further to understand the context more in depth

omnitex: possible issue description
```c
// 1. ALLOCATION (nand.c:179)
page_size = 2048
temp_buffer = heap_caps_malloc(2048 + 4, MALLOC_CAP_DMA | MALLOC_CAP_8BIT)
           = 0x3FC81008  // Only guaranteed 4-byte aligned
// 2. USER CALLS WRITE (from FatFS example)
uint8_t user_data[2048] = {...};  // Stack allocated at 0x480c1494
spi_nand_program_load(handle, user_data, 0, 2048);
// 3. ALIGNMENT CHECK (spi_nand_oper.c:305-312)
alignment = get_alignment() = 64
aligned_len = check_length_alignment(handle, 2048)
           = 2048  // Already aligned to 64 bytes ✓
// Check if user buffer is OK
buf_dma_ok = esp_ptr_dma_capable(0x480c1494) && (0x480c1494 % 64 == 0)
          = false && false
          = false  // User buffer fails both checks
// 4. BOUNCE THROUGH temp_buffer (line 314-315)
memcpy(temp_buffer, user_data, 2048);
data_write = temp_buffer = 0x3FC81008
// 5. SET THE FLAG (line 317)
spi_flags |= SPI_TRANS_DMA_BUFFER_ALIGN_MANUAL;
// "I promise the buffer at 0x3FC81008 is 64-byte aligned"
// 6. SPI DRIVER VALIDATION FAILS
if (0x3FC81008 % 64 != 0) {  // 0x08 % 64 = 8 ≠ 0
    ESP_LOGE("spi_master", "...but TX buffer addr&len not align to 64 byte...");
    return ESP_ERR_INVALID_ARG;
}
```

RathiSonika: Thank you for raising this point—you’re absolutely right. I’ve fixed the issue. PTAL.

spi_nand_flash/src/nand.c
```diff
@@ -176,7 +176,12 @@ esp_err_t spi_nand_flash_init_device(spi_nand_flash_config_t *config, spi_nand_f
     (*handle)->read_buffer = heap_caps_malloc((*handle)->chip.page_size, MALLOC_CAP_DMA | MALLOC_CAP_8BIT);
     ESP_GOTO_ON_FALSE((*handle)->read_buffer != NULL, ESP_ERR_NO_MEM, fail, TAG, "nomem");
 
+#ifndef CONFIG_IDF_TARGET_LINUX
+    size_t dma_alignment = spi_nand_get_dma_alignment();
+    (*handle)->temp_buffer = heap_caps_aligned_alloc(dma_alignment, (*handle)->chip.page_size + dma_alignment, MALLOC_CAP_DMA | MALLOC_CAP_8BIT);
+#else
     (*handle)->temp_buffer = heap_caps_malloc((*handle)->chip.page_size + 4, MALLOC_CAP_DMA | MALLOC_CAP_8BIT);
```
pacucha42: Please, remove `MALLOC_CAP_DMA` flag, it makes no sense on Linux target

pacucha42: You can use `MALLOC_CAP_DEFAULT`, as the cleanest solution

RathiSonika: There is no need for temp_buffer for the Linux target. It is mainly used by spi_nand_oper.c, which is compiled only for non-Linux targets. Therefore, this allocation has been removed for the Linux target.

spi_nand_flash/src/nand.c
pacucha42, line 173: It seems we are using MALLOC_CAP_DMA for Linux target anyway - but it's alright, this doesn't cause any harm and removing is not worthy of worse readability. Just a side note

PR #702, opened 2026-03-12
onewire_bus: using fixed mem_block_symbols for RMT RX (IEC-498)

lukecyca:
# Checklist

- [ ] CI passing

# Change description

Currently the `onewire_bus` component uses the RMT peripheral, and configures `mem_block_symbols` to 80 in some cases. This works for a simple example, but it prevents us from making good use of the other RMT channels, because it uses more memory than the nominal per-channel amount (64 on ESP32 and ESP32S2, 48 on others).

With this change, we allocate every RMT channel using the nominal number of symbols for that system (64 or 48). If the application requests reading more bytes than that, we read them in chunks.

I have tested this with ds18b20, in both single configurations (one sensor on the bus, with no address specified) and in dual configuration (two sensors on the bus, enumerated).

### DIFF
--- a/onewire_bus/CHANGELOG.md
+++ b/onewire_bus/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.1.1
+
+- When reading more bytes than fit in one RMT memory block on ESP32 and ESP32-S2, data will be read in chunks.
+
 ## 1.1.0
 
 - Add UART backend support for 1-Wire bus (`onewire_new_bus_uart`) alongside the existing RMT backend.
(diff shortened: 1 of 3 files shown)


lukecyca:
@Kainarx I have made the changes as you suggest. Thanks!

lukecyca:
This has been open for a couple months and as far as I know I have addressed all of the review comments.

If this patch is not of interest, we can close this and I will simply vendor the patch directly into my firmware.

suda-morris (approved):

PR #715, opened 2026-03-30
fix(libsodium): disable GCC static analyzer for libsodium

### LINKED ISSUE #18373: idf.py build hangs on Completed 'bootloader' step
### Answers checklist.

- [x] I have read the documentation [ESP-IDF Programming Guide](https://docs.espressif.com/projects/esp-idf/en/latest/) and the issue is not addressed there.
- [x] I have updated my IDF branch (master or release) to the latest version and checked that the issue is present there.

### IDF version.

5.5.3

### Operating System used.

Windows

### How did you build your project?

Command line with idf.py

### If you are using Windows, please specify command line type.

CMD

### What is the expected behavior?

The expected behavior is for it to move past this step fairly quickly / nearly instantly as using -v shows that it is only making a few directories and touching files. It is expected that the build should be able to finish.

### What is the actual behavior?

The idf.py build command hangs on the Completed 'bootloader' step and never proceeds past it. Checking task manager there is a cc1 process running on one thread with max cpu usage, but it never exits or seems to do anything.

### Steps to reproduce.

1. Ryzen 9 9955HX, 32gb ram, fairly certain it's not hanging because my computer is slow but maybe because of some multithreading lockup?
2. Run fullclean
3. Run build
sdkconfig: https://pastebin.com/ByMRthKp

### Build or installation Logs.

```plain
buildlog: https://pastebin.com/SFQTZ5Su
```

### Diagnostic report archive.

[idf-diag-74a94598-ced1-4ea8-8dcc-053f2db5f4e5.zip](https://github.com/user-attachments/files/26161452/idf-diag-74a94598-ced1-4ea8-8dcc-053f2db5f4e5.zip)

### More Information.

Haven't encountered this issue before, might be project specific? Maybe a bad sdkconfig setting?

fhrbata:
## Summary

- Adds `-fno-analyzer` compile option to the libsodium component when `CONFIG_COMPILER_STATIC_ANALYZER` is enabled with GCC
- GCC's `-fanalyzer` causes the compiler to hang indefinitely on libsodium's elliptic curve code (`ed25519_ref10.c`) due to combinatorial state explosion in the interprocedural analysis
- This is the same approach already used by ESP-IDF for mbedtls, wpa_supplicant, freertos, and other components

Closes https://github.com/espressif/esp-idf/issues/18373

## Test plan

- [x] Verified that without the fix, building a minimal project with `CONFIG_COMPILER_STATIC_ANALYZER=y` and `espressif/libsodium` dependency hangs indefinitely (GCC's `cc1` stuck at 100% CPU on `ed25519_ref10.c`)
- [x] Verified that with the fix, the same project compiles `ed25519_ref10.c` successfully and the build completes

### DIFF
--- a/libsodium/CMakeLists.txt
+++ b/libsodium/CMakeLists.txt
@@ -187,3 +187,9 @@ if(CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE)
     # some libsodium variables are only used for asserts
     target_compile_options(${COMPONENT_LIB} PRIVATE -Wno-unused-but-set-variable)
 endif()
+
+if(CONFIG_COMPILER_STATIC_ANALYZER AND CMAKE_C_COMPILER_ID STREQUAL "GNU")
+    # GCC's -fanalyzer causes the compiler to hang on libsodium's elliptic curve code
+    # due to combinatorial state explosion in the interprocedural analysis
+    target_compile_options(${COMPONENT_LIB} PRIVATE "-fno-analyzer")
+endif()
--- a/libsodium/idf_component.yml
+++ b/libsodium/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.0.20~4"
+version: "1.0.20~5"
 description: libsodium port to ESP
 url: https://github.com/espressif/idf-extra-components/tree/master/libsodium
 dependencies:


libsodium/CMakeLists.txt
```diff
@@ -187,3 +187,7 @@ if(CONFIG_COMPILER_OPTIMIZATION_ASSERTIONS_DISABLE)
     # some libsodium variables are only used for asserts
     target_compile_options(${COMPONENT_LIB} PRIVATE -Wno-unused-but-set-variable)
 endif()
+
```
mahavirj, line 190: Please drop in a comment for this change

mahavirj (approved):

PR #718, opened 2026-04-02
fix(cbor): Fix build with picolibc (IEC-512)

abobija:
# Checklist

- [x] CI passing

# Change description
Build of project that uses `espressif/cbor^0.6.1~4` component and picolibc (with idf v6+) fails since picolibc does not have implementation of `fopencookie` function and `cookie_io_functions_t` typedef.

By defining `__APPLE__` we are switching to picolibc supported `funopen` alternative that makes project successfully compiled on idf v6+ using picolibc. Note that `CONFIG_LIBC_PICOLIBC_NEWLIB_COMPATIBILITY` still needs to be enabled to compile with picolibc, even if we switch to `__APPLE__`.

It would be nice if cbor component could work without defining `CONFIG_LIBC_PICOLIBC_NEWLIB_COMPATIBILITY` and rely completely on picolibc without any hacks.

I've opened a PR in origin tinycbor's repo as well to support picolibc, so maybe we will accomplish something:
https://github.com/intel/tinycbor/pull/319

Maybe to wait if intel's tinycbor is going to support picolibc...

### DIFF
--- a/cbor/CMakeLists.txt
+++ b/cbor/CMakeLists.txt
@@ -13,8 +13,13 @@ idf_component_register(SRCS "tinycbor/src/cborencoder_close_container_checked.c"
                     INCLUDE_DIRS "tinycbor/src")
 
 # for open_memstream.c
-set_source_files_properties(tinycbor/src/open_memstream.c PROPERTIES
-                            COMPILE_DEFINITIONS "__linux__")
+if(CONFIG_LIBC_PICOLIBC)
+        set_source_files_properties(tinycbor/src/open_memstream.c PROPERTIES
+                                COMPILE_DEFINITIONS "__APPLE__")
+else()
+        set_source_files_properties(tinycbor/src/open_memstream.c PROPERTIES
+                                COMPILE_DEFINITIONS "__linux__")
+endif()
 
 # Fix unreachable macro redefinition issue between ESP-IDF toolchain and tinycbor
 # by force-including a header that resolves the conflict


mahavirj:
@Lapshin PTAL

Lapshin:
@abobija , thanks for the PR.

> Note that CONFIG_LIBC_PICOLIBC_NEWLIB_COMPATIBILITY still needs to be enabled to compile with picolibc, even if we switch to __APPLE__.

Could you please elaborate on why it still needs to be enabled?
Regardless of whether this is required, I don't consider this change valuable. Rather than defining `__APPLE__`, `__linux__`, `__windows__`, and other unrelated macros (ESP32 does not fall under any of these), the original https://github.com/intel/tinycbor sources should detect the availability of libc functions in their configure script (or whatever mechanism they use for this purpose).

Lapshin:
@abobija , I thought that commit https://github.com/espressif/esp-idf/commit/51f18614484d63e25f4b6f4dfd618cdfe046e205 exists for v6.0, but for some reason is present only in master branch. Will make a backport

Lapshin:
https://github.com/espressif/esp-idf/commit/ae8d48470bd02f4fed4e6c56f01801926b041bd1 in release/v6.0 branch. It just did not add in time for v6.0 release. Should be released with bugfix version v6.0.1

Lapshin:
@abobija , FYI, v6.0.1 [is released](https://github.com/espressif/esp-idf/releases/tag/v6.0.1) and has [a workaround](https://github.com/espressif/esp-idf/commit/ae8d48470bd02f4fed4e6c56f01801926b041bd1)

PR #722, opened 2026-04-07
test(spi_nand_flash): add FTL-level host tests via logical-sector and BDL APIs (IEC-515)

omnitex:
## spi_nand_flash: Linux mmap fix, FTL host tests, and BDL host tests

This PR contains three commits (oldest first):

1. **`test(spi_nand_flash): add FTL-level host tests via logical-sector API`** — adds `host_test/main/test_nand_flash_ftl.cpp` (~785 lines, 28 Catch2 cases) and wires it in `host_test/main/CMakeLists.txt`. Exercises the public logical-sector API on the Linux mmap emulator without going through raw NAND page paths (those stay in `test_nand_flash.cpp`).

2. **`test(spi_nand_flash): add comprehensive BDL tests for write/read operations`** — extends `host_test/main/test_nand_flash_bdl.cpp` with a large set of Block Device Layer cases (~778 lines total in tree, ~537 lines added in that commit), covering Flash BDL and WL BDL geometry, IOCTLs, error paths, sequential/stress workloads, and raw-page behaviour behind BDL.

3. **`fix(spi_nand_flash): correct Linux emulator block-size and mmap offset calculation`** — fixes `nand_impl_linux.c` and `nand_linux_mmap_emul.c`: `chip.block_size` must be user-visible data only (`ppb * page_size`), not data+OOB. Adds `linux_mmap_block_file_offset()` so file layout (page slot = `page_size` + OOB) matches what `nand_is_bad` / `nand_mark_bad` / `nand_erase_block` and `nand_emul_erase_block` expect, so bad-block handling and erase clear the full emulated block including OOB.

**Why the tests:** FTL-level coverage gives a baseline for Dhara/FTL/WL experiments without hardware. BDL coverage validates the block-device path (including wear-levelling integration) on the same Linux emulator. The mmap fix aligns the emulator with the rest of the stack so those tests exercise correct geometry and file offsets.

### DIFF
--- a/spi_nand_flash/host_test/main/CMakeLists.txt
+++ b/spi_nand_flash/host_test/main/CMakeLists.txt
@@ -4,6 +4,7 @@ if(CONFIG_NAND_FLASH_ENABLE_BDL)
     list(APPEND src "test_nand_flash_bdl.cpp")
 else()
     list(APPEND src "test_nand_flash.cpp")
+    list(APPEND src "test_nand_flash_ftl.cpp")
 endif()
 
 idf_component_register(SRCS ${src}
--- a/spi_nand_flash/src/nand_impl_linux.c
+++ b/spi_nand_flash/src/nand_impl_linux.c
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <inttypes.h>
+#include <stdint.h>
 #include <string.h>
 #include "esp_check.h"
 #include "esp_err.h"
@@ -12,6 +14,22 @@
 #include "nand_linux_mmap_emul.h"
 
 static const char *TAG = "nand_linux";
+
+/* Start of erase block `block` in the mmap file: ppb slots of (data + OOB) per page. */
+static esp_err_t linux_mmap_block_file_offset(const spi_nand_flash_device_t *h, uint32_t block, size_t *out_offset)
+{
+    ESP_RETURN_ON_FALSE(out_offset != NULL, ESP_ERR_INVALID_ARG, TAG, "out_offset is NULL");
+    ESP_RETURN_ON_FALSE(h->chip.num_blocks > 0, ESP_ERR_INVALID_STATE, TAG, "num_blocks is 0");
+    ESP_RETURN_ON_FALSE(block < h->chip.num_blocks, ESP_ERR_INVALID_ARG, TAG, "block index out of range");
+
+    const uint64_t ppb = 1ull << h->chip.log2_ppb;
+    const uint64_t bytes_per_block = ppb * (uint64_t)h->chip.emulated_page_size;
+    const uint64_t off = (uint64_t)block * bytes_per_block;
+
+    ESP_RETURN_ON_FALSE(off <= (uint64_t)SIZE_MAX, ESP_ERR_INVALID_SIZE, TAG, "mmap block offset overflow");
+    *out_offset = (size_t)off;
+    return ESP_OK;
+}
 
 static esp_err_t detect_chip(spi_nand_flash_device_t *dev)
 {
@@ -31,15 +49,19 @@ static esp_err_t detect_chip(spi_nand_flash_device_t *dev)
         dev->chip.emulated_page_oob = 128;
     }
     dev->chip.emulated_page_size = dev->chip.page_size + dev->chip.emulated_page_oob;
-    dev->chip.block_size = (1 << dev->chip.log2_ppb) * dev->chip.emulated_page_size;
+    /* chip.block_size: user-visible data per erase block (matches nand_impl.c / BDL).
+     * File layout interleaves OOB after each page; use file_bytes_per_block for num_blocks
+     * and linux_mmap_block_file_offset() for mmap byte addresses. */
+    dev->chip.block_size = (1u << dev->chip.log2_ppb) * dev->chip.page_size;
+    const uint32_t file_bytes_per_block = (1u << dev->chip.log2_ppb) * dev->chip.emulated_page_size;
 
-    if (dev->chip.block_size == 0) {
+    if (dev->chip.block_size == 0 || file_bytes_per_block == 0) {
         ESP_LOGE(TAG, "Invalid block size (0)");
         ret = ESP_ERR_INVALID_SIZE;
         goto fail;
     }
 
-    dev->chip.num_blocks = config->emul_conf->flash_file_size / dev->chip.block_size;
+    dev->chip.num_blocks = config->emul_conf->flash_file_size / file_bytes_per_block;
     dev->chip.erase_block_delay_us = 3000;
     dev->chip.program_page_delay_us = 630;
     dev->chip.read_page_delay_us = 60;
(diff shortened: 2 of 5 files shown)


omnitex:
# `chip.block_size` vs Linux mmap page stride

Trace of the discrepancy between `dev->chip.block_size` and the Linux emulated page size, and how FTL/BDL host tests exposed it.

## What was wrong

In `detect_chip()` **before**, `dev->chip.block_size` was set to:

`ppb × emulated_page_size` (= data + OOB per page × pages per block)

so it really described **how many bytes one erase block occupies in the mmap file**, not “user-visible bytes per erase block.”

Everywhere else in this component, **`chip.block_size` is the NAND data erase unit**: `ppb × page_size` (see `nand_impl.c` and the blockdev setup).

```108:108:spi_nand_flash/src/nand_impl.c
    (*handle)->chip.block_size = (1 << (*handle)->chip.log2_ppb) * (*handle)->chip.page_size;
```

```347:358:spi_nand_flash/src/nand_flash_blockdev.c
    uint32_t page_size = handle->chip.page_size;
    uint32_t block_size = handle->chip.block_size;
    uint32_t num_blocks = handle->chip.num_blocks;

    blockdev->geometry.disk_size = num_blocks * block_size;
    blockdev->geometry.write_size = page_size;
    blockdev->geometry.read_size = page_size;
    blockdev->geometry.erase_size = block_size;
```

So the discrepancy was not “erase vs read” in isolation; it was **`chip.block_size` overloading two different strides**:

| Meaning | Stride |
|--------|--------|
| Logical / BDL / FTL “block” (data only) | `ppb × page_size` |
| Mmap file layout (per page slot) | `emulated_page_size` = `page_size + OOB` |

Read/program already used **`page × emulated_page_size`** for file offsets (`nand_prog`, `nand_read`, …). Bad-block and erase paths used **`block × chip.block_size`** and **`nand_emul_erase_block(..., chip.block_size)`**. With `block_size == ppb × emulated_page_size`, those mmap offsets and erase length **matched the file**, but **`chip` lied to anything that treats `block_size` as data geometry**.

## Where erase tied in

`nand_flash_blockdev_erase` does **not** use `geometry.erase_size` as a bit-shift; it uses **`log2_page_size + log2_ppb`**, i.e. the **data** erase block in a power-of-two address space:

```146:152:spi_nand_flash/src/nand_flash_blockdev.c
    spi_nand_flash_device_t *dev_handle = (spi_nand_flash_device_t *)handle->ctx;
    uint8_t log2_block_size = dev_handle->chip.log2_page_size + dev_handle->chip.log2_ppb;
    uint32_t start_block = (uint32_t)(start_addr >> log2_block_size);
    uint32_t num_blocks = (uint32_t)(erase_len >> log2_block_size);
    esp_err_t ret = ESP_OK;
    for (uint32_t block = start_block; block < (start_block + num_blocks); block++) {
        ret = nand_erase_block(dev_handle, block);
```

So:

- **Alignment / length** checks use `geometry.erase_size` (from `chip.block_size`).
- **Which physical block index** is erased uses **`1 << (log2_page + log2_ppb)` = `ppb × page_size`**.

If `erase_size` includes OOB, it is **not** the same as `1 << log2_block_size` whenever `OOB > 0`. You get inconsistent “block device” math: same `start_addr` can be “aligned” to one notion of block size while the **block index** is computed with another. That is exactly the kind of bug BDL tests stress (erase boundaries, geometry vs `GET_NAND_FLASH_INFO`, multi-block erase).

Separately, once **`chip.block_size` is corrected** to `ppb × page_size` for the rest of the stack, **`nand_emul_erase_block` must not** keep using `chip.block_size` for `memset` length: it would only wipe the **data** span and leave **OOB** stale. The commit fixes that by using an explicit `nbytes = ppb × emulated_page_size` in `nand_linux_mmap_emul.c`, while **`linux_mmap_block_file_offset()`** supplies `block × ppb × emulated_page_size` for the mmap base address.

## How the FTL/BDL host tests exposed it

The new/extended host path goes through **real BDL geometry** (`disk_size`, `erase_size`) and **ioctl geometry** (`GET_NAND_FLASH_INFO` copies `dev->chip`). Tests assert things like **`erase_size == flash_info.geometry.block_size`** and drive **byte-granular erase/write** using `bdl->geometry.erase_size` and `page_size`. With the old `chip.block_size`, reported erase block size and capacity did not match the **page-index / log2 erase-block model** the blockdev uses internally, so you could get wrong block indices, inconsistent totals, or bad-block handling that disagreed with what the FTL assumed—failures that simple raw NAND tests might never hit because they do not reconcile **BDL linear addressing** with **mmap stride + OOB**.

## Bottom line

The bug was **`chip.block_size` meaning “file bytes per block” instead of “data bytes per erase block.”** Erase was the sharp edge because **blockdev erase maps byte addresses → block index via data-only `log2` math** while also trusting **`chip.block_size`** for exposed geometry and alignment; the mmap emulator then had to split **`file_bytes_per_block`** vs **`block_size`** and erase **`nbytes`** in the file, which is what `481ea564` implements.

spi_nand_flash/host_test/main/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ if(CONFIG_NAND_FLASH_ENABLE_BDL)
     list(APPEND src "test_nand_flash_bdl.cpp")
 else()
     list(APPEND src "test_nand_flash.cpp")
+    list(APPEND src "test_nand_flash_ftl.cpp")
```
RathiSonika, line 7:
```suggestion
    list(APPEND src  "test_nand_flash.cpp" "test_nand_flash_ftl.cpp")
```
nitpick

spi_nand_flash/src/nand_impl_linux.c
RathiSonika, line 2:
```suggestion
 * SPDX-FileCopyrightText: 2015-2026 Espressif Systems (Shanghai) CO LTD
```

spi_nand_flash/src/nand_linux_mmap_emul.c
RathiSonika, line 1:
```suggestion
/*
 * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
 *
 * SPDX-License-Identifier: Apache-2.0
 */
```

RathiSonika (commented):
Please bump the component version and update the CHANGELOG.md accordingly.

spi_nand_flash/src/nand_impl_linux.c
```diff
@@ -13,6 +15,22 @@
 
 static const char *TAG = "nand_linux";
 
+/* Start of erase block `block` in the mmap file: ppb slots of (data + OOB) per page. */
+static esp_err_t linux_mmap_block_file_offset(const spi_nand_flash_device_t *h, uint32_t block, size_t *out_offset)
```
RathiSonika, line 19: Prefer using `handle` instead of `h` here for consistency and readability—just a suggestion, feel free to ignore.

omnitex:
Changes in this PR were transferred to https://github.com/espressif/idf-extra-components/pull/727 (including the review nitpicks by @RathiSonika).

This way, the component version will be bumped just once.

Closing this PR as absolete.

PR #727, opened 2026-04-14
fix(spi_nand_flash): Linux mmap emulator fixes, docs, and host tests (v1.0.2) (IEC-522)

omnitex:
## Description
Patch release for `spi_nand_flash` (**1.0.2** in `idf_component.yml`). Focus is **Linux mmap NAND emulation correctness**, **BDL logging fixes**, and **expanded Linux host tests**. Fault-injection work is **not** part of this change set.
### Fixes
- BDL: correct `printf`-style format specifiers for size-related fields on 64-bit Linux (avoids UB and bogus logs).
- Linux NAND emulation: OOB markers and free-page detection aligned with the real hardware path.
- Linux mmap: correct backing file path selection for the emulated image.
- Linux mmap layout: on-disk stride accounts for interleaved OOB vs user-visible erase-block size (bad-block handling, erase, and OOB clearing).
### Documentation
- `nand_file_mmap_emul_config_t`: document how backing file size relates to interleaved OOB and reported user capacity.
- `README.md` / `host_test/README.md`: align with current mmap config API, build/run notes, and BDL prerequisites where relevant.
### Testing
- Linux host: broader **FTL** and **BDL** coverage.
- Shared buffer pattern helpers (including seeded patterns) for host tests and the in-tree test app.
### Release note
See `spi_nand_flash/CHANGELOG.md` section **[1.0.2]** for the authoritative list.

### DIFF
--- a/spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
+++ b/spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
@@ -1,11 +1,13 @@
 /*
- * SPDX-FileCopyrightText: 2023-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2023-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
+#include <cstdlib>
 
 #include "spi_nand_flash.h"
 #include "spi_nand_flash_test_helpers.h"
@@ -14,6 +16,31 @@
 #include "esp_nand_blockdev.h"
 
 #include <catch2/catch_test_macros.hpp>
+
+/* Linux mmap file: each page is k_page data + k_oob, so the file uses k_ppb * (k_page + k_oob)
+ * bytes per erase block. chip.block_size is k_ppb * k_page (same as real chips); num_blocks
+ * is file_size / (k_ppb * (k_page + k_oob)). BDL disk_size must be num_blocks * chip.block_size.
+ * Regression: storing the file stride in chip.block_size broke BDL erase decode. */
+TEST_CASE("BDL geometry matches Linux mmap file and chip.block_size", "[spi_nand_flash][bdl]")
+{
+    constexpr uint32_t k_file_bytes = 16u * 1024u * 1024u;
+    constexpr uint32_t k_page = 2048u;
+    constexpr uint32_t k_oob = 64u;
+    constexpr uint32_t k_ppb = 64u;
+    const uint32_t file_bytes_per_physical_block = k_ppb * (k_page + k_oob);
+    const uint32_t bdl_bytes_per_erase_block = k_ppb * k_page;
+    const uint32_t num_physical_blocks = k_file_bytes / file_bytes_per_physical_block;
+    const uint64_t expected_disk_size = (uint64_t)num_physical_blocks * bdl_bytes_per_erase_block;
+
+    nand_file_mmap_emul_config_t conf = {"", k_file_bytes, false};
+    spi_nand_flash_config_t nand_flash_config = {&conf, 0, SPI_NAND_IO_MODE_SIO, 0};
+    esp_blockdev_handle_t bdl = nullptr;
+    REQUIRE(nand_flash_get_blockdev(&nand_flash_config, &bdl) == ESP_OK);
+    REQUIRE(bdl->geometry.write_size == k_page);
+    REQUIRE(bdl->geometry.erase_size == bdl_bytes_per_erase_block);
+    REQUIRE(bdl->geometry.disk_size == expected_disk_size);
+    bdl->ops->release(bdl);
+}
 
 TEST_CASE("verify mark_bad_block works with bdl interface", "[spi_nand_flash][bdl]")
 {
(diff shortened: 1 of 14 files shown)


RathiSonika (dismissed):
Overall, changes LGTM. Please consider updating the public API documentation for the returned error codes if needed. Thanks for the fix!

spi_nand_flash/host_test/main/nand_mmap_emul_host_test_config.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include "nand_linux_mmap_emul.h"
+
+/**
+ * Build `nand_file_mmap_emul_config_t` for Linux host tests.
+ *
+ * Uses C++20 designated initializers so new fields never trip
+ * `-Werror=missing-field-initializers`. Empty name keeps mkstemp-backed temp file
+ * behavior (same as legacy `""` aggregate init).
+ */
+[[nodiscard]] inline nand_file_mmap_emul_config_t nand_mmap_emul_host_test_config(
```
omnitex, line 21: Add filename argument as well, so it can be named, not just the default `/tmp/idf-nand-XXXXXX`. Can be useful for keeping a specific dump for analysis.

spi_nand_flash/src/nand_impl_linux.c
RathiSonika, line 39: Can we add a pre-check before nand_emul_init() to ensure emul_conf->flash_file_size % dev->chip.block_size == 0? If not, boundary test cases may hit partial-block issues?

spi_nand_flash/src/nand_impl_linux.c
```diff
@@ -106,28 +127,32 @@ esp_err_t nand_is_bad(spi_nand_flash_device_t *handle, uint32_t block, bool *is_
...
 }
 
 esp_err_t nand_mark_bad(spi_nand_flash_device_t *handle, uint32_t block)
 {
     esp_err_t ret = ESP_OK;
+    size_t block_base = 0;
 
-    uint32_t first_block_page = block * (1 << handle->chip.log2_ppb);
+    uint64_t first_block_page = (uint64_t)block * (1ull << handle->chip.log2_ppb);
     uint16_t bad_block_indicator = 0;
```
RathiSonika, line 156: In the HW path (`nand_impl.c`), we write 4-byte markers to avoid alignment issues.
A similar approach could be used here for consistency. This likely applies to `nand_write()` and `nand_copy()` as well when marking a page as used.

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
```diff
@@ -240,3 +268,821 @@ TEST_CASE("Release and no use-after-free: create, release, create again, minimal
...
+TEST_CASE("Linux mmap inject: fail on 1st erase", "[spi_nand_flash][bdl][raw][fault]")
+{
+    nand_file_mmap_emul_config_t conf = nand_mmap_emul_host_test_config(8 * 1024 * 1024, false, 0, 0, 1);
+    spi_nand_flash_config_t nand_flash_config = {&conf, 0, SPI_NAND_IO_MODE_SIO, 0};
+    esp_blockdev_handle_t bdl = nullptr;
+    REQUIRE(nand_flash_get_blockdev(&nand_flash_config, &bdl) == ESP_OK);
+
+    const uint32_t block_size = bdl->geometry.erase_size;
+    REQUIRE(bdl->ops->erase(bdl, 0, block_size) == ESP_ERR_NOT_FINISHED);
+    REQUIRE(bdl->ops->erase(bdl, 0, block_size) == ESP_OK);
```
RathiSonika, line 1055: What is the expected Dhara behavior on read/write/erase failure? Should it simply retry and succeed, or should it treat the block as a bad block and mark it accordingly?

spi_nand_flash/include/nand_linux_mmap_emul.h
```diff
@@ -20,13 +20,25 @@ typedef struct {
...
     bool keep_dump;
+    /**
+     * Host-only fault simulation: 0 = disabled.
+     * On the N-th successful-boundary `nand_emul_read` / `nand_emul_write` / `nand_emul_erase_block`
+     * invocation for this device (counts each API call, including OOB sub-writes from program/copy),
+     * the call returns `ESP_ERR_NOT_FINISHED` without modifying emulated flash (hardware-like fail).
+     */
+    uint32_t inject_fail_read_n;
+    uint32_t inject_fail_write_n;
+    uint32_t inject_fail_erase_n;
```
RathiSonika, line 47: Can we extend this to cover ECC-related read failures as well? For NAND, this would be useful to emulate both correctable and uncorrectable bitflips, and verify that the upper layers handle them as expected.

spi_nand_flash/src/nand_linux_mmap_emul.c
```diff
@@ -29,10 +31,12 @@ static esp_err_t nand_emul_mmap_init(nand_mmap_emul_handle_t *emul_handle)
     }
 
     // Create or open file
-    if (emul_handle->file_mmap_ctrl.flash_file_name[0] != '\0') {
+    // Note: flash_file_name is never empty here - nand_emul_init always fills it
+    // (either with the caller-provided path or with the "/tmp/idf-nand-XXXXXX" template).
+    // When the template was used, mem_file_fd was already opened via mkstemp() there,
+    // so skip open() in that case.
```
RathiSonika:
```suggestion
// When the "/tmp/idf-nand-XXXXXX" template is used, `mem_file_fd` is already
// opened by `mkstemp()` in `nand_emul_init()`, so `open()` should be skipped here.
```
I think there’s no need to keep the first two comment lines. They’re confusing, especially since flash_file_name has already been removed.

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
```diff
@@ -240,3 +268,821 @@ TEST_CASE("Release and no use-after-free: create, release, create again, minimal
...
+        REQUIRE(wl_bdl->ops->read(wl_bdl, r, page_size, (uint64_t)s * page_size, page_size) == ESP_OK);
+        REQUIRE(spi_nand_flash_check_buffer(r, page_size / sizeof(uint32_t)) == 0);
+    }
+
+    free(w);
+    free(r);
+    wl_bdl->ops->release(wl_bdl);
+}
+
+TEST_CASE("Flash BDL second program without erase is bitwise AND (NAND)", "[spi_nand_flash][bdl][raw]")
```
RathiSonika, line 785: In raw NAND, an already programmed page generally cannot be rewritten without first erasing the containing block, except on devices that explicitly allow limited partial-page programming under specific restrictions, right? Given that, what is the purpose of this test case? Dhara will not attempt to write to an already used page anyway, right?

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
```diff
@@ -0,0 +1,786 @@
...
+    uint8_t *wbuf = (uint8_t *)malloc(sz);
+    uint8_t *rbuf = (uint8_t *)malloc(sz);
+    REQUIRE(wbuf != nullptr);
+    REQUIRE(rbuf != nullptr);
+
+    const uint32_t TARGET_SECTOR = 7;
+    const int OVERWRITE_ROUNDS = 20;
+
+    for (int i = 0; i < OVERWRITE_ROUNDS; i++) {
+        spi_nand_flash_fill_buffer(wbuf, sz / sizeof(uint32_t));
```
RathiSonika: The intent of this test is to verify that, from the application’s perspective, the same logical sector can be written multiple times (while Dhara internally writes to different physical sectors and updates the mapping accordingly), right? However, this test currently writes the same buffer each time. Is there a better way to structure this test?

spi_nand_flash/include/nand_linux_mmap_emul.h
omnitex: Deferred for a future host_test extension with focus on more simulation-centric improvements such as these ECC-related read failures. Thank you for the idea.

spi_nand_flash/src/nand_impl_linux.c
omnitex: Added.

spi_nand_flash/src/nand_impl_linux.c
omnitex: Aligned with the HW path. Thank you.

spi_nand_flash/src/nand_linux_mmap_emul.c
omnitex: Yes, you're right, thank you. Removed the first two comment lines and reworded the rest slightly.

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
omnitex: Yes, you're correct. This test was redundant for the purpose exercising Dhara. Removed, thank you.

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
omnitex: Added new seeded variants of both the `fill_buffer` and `check_buffer` functions to allow proper varying contents across tests which run a loop.

spi_nand_flash/host_test/main/test_nand_flash_bdl.cpp
omnitex: Decided to remove the fault injection from this PR completely, to stay true to the initial intention of the fixes, as a good design proved to not be so trivial. Will propose an overall more unified direction for failure injection and other simulator-like behavior to host test, up to discussion in what capacity we actually want it.

spi_nand_flash/include/nand_linux_mmap_emul.h
```diff
@@ -98,8 +114,8 @@ esp_err_t nand_emul_write(spi_nand_flash_device_t *handle, size_t addr, const vo
  * @param handle spi_nand_flash_device_t handle for nand device
  * @param offset Block Address offset to erase
  * @return ESP_OK on success
- *         ESP_ERR_INVALID_STATE if not initialized
- *         ESP_ERR_INVALID_ARG if address is not block-aligned or exceeds flash size
+ *         ESP_ERR_INVALID_STATE if there is no emulation handle or the backing mmap is not ready
+ *         ESP_ERR_INVALID_SIZE if erase range would exceed emulated flash size
```
RathiSonika: I am not sure how much critical is this, but this would be breaking change right? Changing the return code..

spi_nand_flash/src/nand_impl_linux.c
```diff
@@ -13,11 +15,40 @@
 
 static const char *TAG = "nand_linux";
 
+/* OOB marker layout at page data offset `page_size` (matches nand_impl.c HW path):
+ * bytes 0-1: bad-block marker (0xFFFF = good / erased, 0x0000 = bad)
+ * bytes 2-3: page-used marker (0xFFFF = free, 0x0000 = used after program) */
+static const uint8_t s_oob_prog_markers[4] = { 0xFF, 0xFF, 0x00, 0x00 };
```
RathiSonika:
```suggestion
static const uint8_t s_oob_used_page_markers[4] = { 0xFF, 0xFF, 0x00, 0x00 };
```
Nitpick. Feel free to ignore this

spi_nand_flash/src/nand_impl_linux.c
```diff
@@ -13,11 +15,40 @@
...
+    *out_offset = (size_t)off;
+    return ESP_OK;
+}
+
 static esp_err_t detect_chip(spi_nand_flash_device_t *dev)
 {
     esp_err_t ret = ESP_OK;
     spi_nand_flash_config_t *config = &dev->config;
 
+    if (config->emul_conf->flash_file_size % dev->chip.block_size != 0) {
```
RathiSonika: You already have this check after calculating the block size in the same function, so it can be removed.

RathiSonika (commented):
Please consider addressing mentioned comment. Otherwise LGTM!

spi_nand_flash/src/nand_impl_linux.c
omnitex: Check removed, thanks.

spi_nand_flash/src/nand_impl_linux.c
omnitex: Renamed, thanks.

RathiSonika (approved):
Thank you for addressing the comments. LGTM

PR #728, opened 2026-04-14
refactor(dhara): replace Dhara git submodule with vendored upstream snapshot (IEC-523) (IEC-525)

omnitex:
## Summary
Replace the `dhara` git submodule with an in-tree vendored snapshot of upstream **dlbeer/dhara**, so consumers get Dhara sources without submodule checkout. Packaging and docs only intent: **no intentional FTL behavior change**.
## Baseline / provenance
- **Upstream:** https://github.com/dlbeer/dhara  
- **Baseline commit:** `1b166e41b74b4a62ee6001ba5fab7a8805e80ea2` (byte-for-byte snapshot at this SHA)  
- **Tracking / refresh:** `dhara/VENDORED_UPSTREAM.md`
## Changes
- Vendor Dhara tree under `dhara/dhara/` (remove submodule; ordinary tracked files).
- Update `dhara/README.md` and `spi_nand_flash/README.md` (vendored sources, no submodule).
- Bump `dhara` component version to **1.0.0** and refresh `dhara/sbom_dhara.yml`.
- Bump `spi_nand_flash` to **1.0.3** with the Dhara version and git submodule-removal info in changelog

# Checklist
- [x] ~~rename component to `esp_dhara`?~~ we are okay keeping the `dhara` name for now, to make the transition fully transparent and non-breaking
- [x] ~~git checkout with the dhara sources added in this branch conflicts with existing submodule sources, if present -- what's the workflow for end users? would they have to resolve this? migration info necessary?~~ if components defined via dependencies in `idf_component.yml`, fullclean and new build achieves a clean slate


dhara/sbom_dhara.yml
```diff
@@ -1,7 +1,7 @@
 name: dhara
-version: 1b166e41b74b4a62ee6001ba5fab7a8805e80ea2 
+version: 4c85c045e12f121d496989b78fa4cc89a81a32b9
 cpe: cpe:2.3:a:dhara:dhara:{}:*:*:*:*:*:*:*
-supplier: 'Organization: dhara'
-description: NAND Flash translation layer for small MCUs
+supplier: 'Organization: dlbeer'
```
omnitex: should we have `Organization: Espressif Systems` as we're going to be the maintainers of our own Dhara version?

omnitex: switched to Espressif Systems as we will maintain and supply it going forward, for our internal purposes

RathiSonika:
I think a better approach would be to fork dhara into our org/repository and then use that fork as a submodule here?

Since we plan to maintain our own changes and extensions without upstreaming them, keeping it as a separate fork would give us cleaner ownership, versioning, and easier future maintenance. It would also preserve the original history and keep idf-extra-component from carrying vendored third-party code directly. I think ESP-IDF follows a similar pattern for some third-party components like SPIFFS.

RathiSonika:
After the offline discussion, I agree with Martin’s point that if we move dhara into a separate fork, we would also need a separate CI setup for that fork, and changes there would not automatically be validated in this repo until the submodule/dependency is updated here.

To keep things simpler for now, I think we should skip creating a separate fork and use the current approach so we can continue relying on the existing CI infrastructure and keep the build flow straightforward.

pacucha42 (approved):
Approved asi-is, changes and possible updates discussed offline can be solved in follow-up PRs. Thanks @omnitex!

PR #735, opened 2026-04-23
fix(spi_nand_flash): fix incorrect flash geometry parameter for GD5F4GM8xExxG

RathiSonika:
# Change description

This Pull Request resolves an issue where the GD5F4GM8xExxG (4Gb, DID-85h, 95h) NAND flash fails during page copy operations when using the spi_nand_flash driver. While the datasheet does not explicitly label the device as "dual-plane," its architectural constraints for hardware-accelerated data movement necessitate treating it as such. Updated accordingly.

Datasheet Reference (https://download.gigadevice.com/Datasheet/DS-00846-GD5F4GM8UE-Rev1.5.pdf): Added logic to enforce the rule that Internal Data Move "can only be used on blocks with the same parity attribute" and within the same 2Gb partition

### DIFF
--- a/spi_nand_flash/CHANGELOG.md
+++ b/spi_nand_flash/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [1.0.1]
+- fix: fix incorrect flash geometry parameter for flash GD5F4GM8xExxG
+
 ## [1.0.0]
 ### Breaking Changes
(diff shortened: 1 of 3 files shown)


pacucha42 (approved):
LGTM, approved. Thank you @RathiSonika

PR #741, opened 2026-05-05
fix(spi_nand_flash): Disable QIO/QOUT tests on runner

RathiSonika:
# Change description
Disabling QIO and QOUT tests for now, as they are failing on the runner. The NAND flash on the runner is connected via wires, which may be causing signal integrity issues in QIO mode. I’ll work on fixing this as soon as possible. In the meantime, these tests are being disabled to unblock the pipelines.


pacucha42 (approved):
Please, enable the tests as soon as possible. Approved

PR #747, opened 2026-05-15
feat(spi_nand_flash): add fudan microelectronics (FM25S005BI3) NAND flash support

RathiSonika:
# Change description
Added Fudan Microelectronics (FM25S005BI3) NAND flash support


omnitex (approved):
LGTM

Possibly consider making the abbreviation of the manufacturer `FMSH` instead of `FM` (in constants and file names etc.):
- from datasheet `MID is Manufacture ID (A1h for FMSH)...`
- their website is [fmsh.com](https://www.fmsh.com/)
- you already abbreviate it as FMSH in the README, so there's existing precedent for that

PR #748, opened 2026-05-19
feat(spi_nand_flash): add support for GD5F4GM7xExxG NAND device

RathiSonika:
# Change description
Added support for GD5F4GM7xExxG NAND device

### DIFF
--- a/spi_nand_flash/priv_include/nand_flash_devices.h
+++ b/spi_nand_flash/priv_include/nand_flash_devices.h
@@ -49,6 +49,7 @@ extern "C" {
 #define GIGADEVICE_DI_82              0x82
 #define GIGADEVICE_DI_91              0x91
 #define GIGADEVICE_DI_81              0x81
+#define GIGADEVICE_DI_94              0x94   // GD5F4GM7UExxG
 
 // Alliance Memory
 #define ALLIANCE_DI_25                0x25   // AS5F31G04SND-08LIN
--- a/spi_nand_flash/src/devices/nand_gigadevice.c
+++ b/spi_nand_flash/src/devices/nand_gigadevice.c
@@ -55,6 +55,15 @@ esp_err_t spi_nand_gigadevice_init(spi_nand_flash_device_t *dev)
     case GIGADEVICE_DI_95:
     case GIGADEVICE_DI_85:
         dev->chip.num_blocks = 4096;
+        // Not hardware dual-plane: num_planes models odd/even block parity required
+        // for Internal Data Move (GD5F4GM8 datasheet §9.5). See nand_copy().
+        dev->chip.flags = NAND_FLAG_HAS_PROG_PLANE_SELECT | NAND_FLAG_HAS_READ_PLANE_SELECT;
+        dev->chip.num_planes = 2;
+        break;
+    case GIGADEVICE_DI_94:
+        dev->chip.log2_page_size = 12;  // 4096 bytes per page
+        dev->chip.num_blocks = 2048;
+        // Same parity grouping as GD5F4GM8 (GIGADEVICE_DI_95) above.
         dev->chip.flags = NAND_FLAG_HAS_PROG_PLANE_SELECT | NAND_FLAG_HAS_READ_PLANE_SELECT;
         dev->chip.num_planes = 2;
         break;
--- a/spi_nand_flash/CHANGELOG.md
+++ b/spi_nand_flash/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Versioning policy: see [VERSIONING.md](VERSIONING.md). From **v1.0.0** onward this component follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## [1.2.0]
+- feat: added support for Gigadevice GD5F4GM7UExxG NAND flash
+
 ## [1.1.0]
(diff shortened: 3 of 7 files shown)


spi_nand_flash/priv_include/nand_flash_devices.h
```diff
@@ -48,6 +48,7 @@ extern "C" {
 #define GIGADEVICE_DI_82              0x82
 #define GIGADEVICE_DI_91              0x91
 #define GIGADEVICE_DI_81              0x81
+#define GIGADEVICE_DI_94              0x94
```
omnitex: can you please add an entry for `DI 0x84` as well? just like https://github.com/espressif/idf-extra-components/pull/703 added 2 entries, because if we're adding support for `GD5F4GM7xExxG`, it should cover both the 
- `GD5F4GM7UExxG -> DI 0x94`
- `GD5F4GM7RExxG -> DI 0x84`
the datasheets specifies.

spi_nand_flash/src/devices/nand_gigadevice.c
```diff
@@ -58,6 +58,12 @@ esp_err_t spi_nand_gigadevice_init(spi_nand_flash_device_t *dev)
         dev->chip.flags = NAND_FLAG_HAS_PROG_PLANE_SELECT | NAND_FLAG_HAS_READ_PLANE_SELECT;
         dev->chip.num_planes = 2;
         break;
+    case GIGADEVICE_DI_94:
+        dev->chip.log2_page_size = 12;  // 4096 bytes per page
+        dev->chip.num_blocks = 2048;
+        dev->chip.flags = NAND_FLAG_HAS_PROG_PLANE_SELECT | NAND_FLAG_HAS_READ_PLANE_SELECT;
+        dev->chip.num_planes = 2;
+        break;
```
omnitex, line 69: can you please double check this? I can't find any dual plane info in the datasheet for this flash chip. additionally, in the datasheet there is a section for `CASN page` in the OTP area, which lists
```
logical planes per lun | 01h
```

so I believe this flash chip is single plane?

The same issue is afaict existing for the gigadevice IDs just above the ones this PR adds:
```
case GIGADEVICE_DI_95:
case GIGADEVICE_DI_85:
        dev->chip.num_blocks = 4096;
        dev->chip.flags = NAND_FLAG_HAS_PROG_PLANE_SELECT | NAND_FLAG_HAS_READ_PLANE_SELECT;
        dev->chip.num_planes = 2;
```
`0x95, 0x85` belong to `GD5F4GM8UExxG` and `GD5F4GM8RExxG` which this commit https://github.com/espressif/idf-extra-components/commit/7bbbd02720d28d9f194e0299f242de8566e7c740 made dual plane, but I don't see anything about dual plane in their datasheet either?

am I missing something? can you please clarify?

spi_nand_flash/priv_include/nand_flash_devices.h
RathiSonika: I intentionally did not add DI 0x84 (GD5F4GM7RExxG). The RE variant is rated for 1.7–2.0 V, while the UE variant (0x94) is 2.7–3.6 V. This driver targets ESP32 boards which operates at 3.3 V, so only the UE part I have added in this PR.

Older GigaDevice RE device IDs remain in the driver from earlier commits and are not being removed in this PR. I will update the README to clarify the UE/RE voltage distinction and list only GD5F4GM7UExxG as supported for this addition.

Do you have any suggestion on how this should be handled?

spi_nand_flash/src/devices/nand_gigadevice.c
RathiSonika: You’re right that this is not a traditional dual-plane NAND device.

However, according to the datasheet (Internal Data Move, §9.5), the operation is only valid between blocks that share the same odd/even parity attribute; it cannot be performed between an odd block and an even block. That introduces a copy restriction similar to the one we already handle for GD5F4GM8 in #735.

So in this case, num_planes = 2 and the plane-select flags are not meant to indicate a second physical plane. Instead, they are used to model this parity-based grouping in the column-addressing logic so that nand_copy() behaves correctly:
```
same parity → same column address → hardware copy path
different parity → different column address → RAM copy fallback
```
Without this, nand_copy() would treat all blocks uniformly and might attempt an internal data move between odd and even blocks, which the datasheet explicitly forbids. That is the same class of issue previously seen on GD5F4GM8 before #735.

I agree that the current naming (num_planes, plane-select flags) is somewhat misleading in this context. A follow-up refactor with clearer terminology would make sense, but the behavior for GD5F4GM7UE is intentionally aligned with the existing GD5F4GM8 handling.

spi_nand_flash/priv_include/nand_flash_devices.h
omnitex: I see, thank you for the explanation, I did not previously check/consider the voltage ratings. 

Adding just the correct variant with explanation: OK

spi_nand_flash/src/devices/nand_gigadevice.c
omnitex: If this is required for correct Internal Data move operation, then OK.

Please document this fact in an appropriate comment, so the fact that `num_planes` *does not actually mean what is says* for this (and the previously merged case) is clear until the refactor takes place, thank you.

omnitex (approved):

Issue #554, opened 2025-09-02
latest cbor versions can cause build failure (IEC-361)


psilva-dsr:
### Answers checklist.

- [x] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [x] I am using target and esp-idf version as defined in component's idf_component.yml
- [x] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

cbor

### ESP-IDF version.

v5.4.1

### Development Kit.

Custom board

### Used Component version.

0.6.1~1

### More Information.

issue due to this commit:
https://github.com/espressif/idf-extra-components/commit/cc9d1d8a1a8a2f387b3786ccf8c8ee0fa855b74f

and behavior of `target_compile_options`: https://cmake.org/cmake/help/latest/command/target_compile_options.html#option-de-duplication

`option-de-duplication` means that when there are multiple `target_compile_options` using "--include" "-file-", only one "--include" statement will be generated, resulting in: "--include" "-file-" "-file-"

in my project, I have cbor and memfault, with both using `target_compile_options` for "--include"

possible fix:
```
target_compile_options(${COMPONENT_LIB} PRIVATE
                       "-include${CMAKE_CURRENT_SOURCE_DIR}/port/include/unreachable_fix.h")
```

PR #752, opened 2026-05-21
fix(libsodium): fix build failure with platformio (IEC-543)

Ashish285:
# Change description
When `CONFIG_LIBSODIUM_USE_MBEDTLS_SHA` is enabled, `libsodium/CMakeLists.txt` injects two -include flags (for the sha256/sha512 port shims) via a CMake `SHELL:` generator expression. Building this component under PlatformIO fails, its build generator does not honor the `SHELL:` prefix, so the option is then re-split on whitespace and CMake deduplicates one of the two -include entries, leaving the relevant header un-included at compile time.

## Fix

Drop the `SHELL:` prefix and join `-include` to the header path so the whole flag is a single, indivisible token:

### DIFF
--- a/libsodium/CMakeLists.txt
+++ b/libsodium/CMakeLists.txt
@@ -209,8 +209,8 @@ target_compile_options(${COMPONENT_LIB} PRIVATE -Wno-unused-function)
 
 if(CONFIG_LIBSODIUM_USE_MBEDTLS_SHA)
     target_compile_options(${COMPONENT_LIB} PRIVATE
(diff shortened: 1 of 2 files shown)


emlynmac:
Confirmed this fixes up a regression in behaviour in 21 recorded here: https://github.com/rbouteiller/airplay-esp32/issues/72

mahavirj:
@suda-morris @igrr Can you please check this once? Related: https://github.com/espressif/idf-extra-components/issues/554#issuecomment-3244912152

PR #757, opened 2026-06-02
fix(network_prov): Fix BLE provisioning preserving AP mode

### LINKED ISSUE #745: The network_provisioning component will inexplicably change the Wi-Fi mode and will not change it back after provisioning is completed. (IEC-538)
### Answers checklist.

- [x] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [x] I am using target and esp-idf version as defined in component's idf_component.yml
- [x] I have searched the [issue tracker](https://github.com/espressif/idf-extra-components/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

argtable3

### ESP-IDF version.

release/v6.0

### Development Kit.

esp32

### Used Component version.

1.2.4

### More Information.

During device initialization, I set the Wi-Fi mode to APSTA, and then I enabled network provisioning. At this point, the component internally changes the Wi-Fi mode to STA. I understand that this change is intended to reduce coexistence interference between SoftAP and Bluetooth. However, after provisioning is completed, the Wi-Fi mode is not changed back to the user's original setting—it remains in STA mode.

This causes confusion for the application layer, because the user originally set APSTA mode, but after provisioning, the SoftAP is no longer available.

Suggestion: Provide an option for users to choose whether the network provisioning component should disable SoftAP internally. Based on actual testing, using BLE network provisioning in APSTA mode does not cause any noticeable issues. Alternatively, if for some internal reason the mode must be changed to STA during provisioning, the component should at least change the Wi-Fi mode back to the user's original setting after provisioning is completed.

wqx6:
# Checklist

- [ ] CI passing

# Change description
Fixes #745

Preserve application-owned SoftAP mode for network provisioning


shahpiyushv (commented):
Approach looks right: keeping `APSTA` for the whole session (and in `configure_wifi_sta()`, which would otherwise have dropped the AP the moment credentials were applied) is better than restoring the mode afterwards. Inline comments below, plus three things not visible in the diff:

- `idf_component.yml` is still `1.2.4`, same as master — needs a version bump plus a `CHANGELOG.md` entry, otherwise the fix doesn't reach users via the registry.
- `network_prov_scheme_t.wifi_mode` in `include/network_provisioning/manager.h` still documents "WIFI_MODE_STA for BLE transport"; the behaviour is now conditional. Worth documenting there/in the README, including that the SoftAP follows the STA channel on connect and so drops its clients.
- Behaviour change for existing users: an app in APSTA previously got STA-only for the whole BLE session, now BLE and SoftAP coexist throughout and the full-band scan runs with the AP up. #745 reports this works, but it's a deliberate call — a Kconfig opt-in would be the safer shape if there's any concern.

Could you also note the manual verification in the PR description? APSTA + BLE happy path (AP still up after `NETWORK_PROV_END`), failure/reset path, `network_prov_mgr_deinit()` path, softap scheme unaffected, and a Thread-only build.

network_provisioning/src/manager.c
```diff
@@ -2115,11 +2131,23 @@ esp_err_t network_prov_mgr_start_provisioning(network_prov_security_t security,
...
-     * This is necessary for scanning to work */
-    ret = esp_wifi_set_mode(WIFI_MODE_STA);
+    prov_ctx->ap_mode_enabled_before_start = false;
+
+    wifi_mode_t wifi_mode_before_start = WIFI_MODE_NULL;
+    ret = esp_wifi_get_mode(&wifi_mode_before_start);
+    if (ret == ESP_OK &&
+            wifi_mode_before_start == WIFI_MODE_APSTA &&
+            prov_ctx->mgr_config.scheme.wifi_mode == WIFI_MODE_STA) {
+        prov_ctx->ap_mode_enabled_before_start = true;
```
shahpiyushv, line 2141: `WIFI_MODE_AP` misses this check, so an app in AP-only mode (config portal with no STA yet) still gets the exact #745 behaviour: forced to STA, AP gone, never restored. The defect is "app-owned SoftAP is destroyed", not "APSTA is destroyed".

```c
(wifi_mode_before_start == WIFI_MODE_APSTA || wifi_mode_before_start == WIFI_MODE_AP)
```

Explicit comparison rather than `mode & WIFI_MODE_AP` — bit-masking that enum works today but isn't a documented contract. Keeping the `bool` (instead of restoring the previous mode verbatim) is right, since restoring pure `WIFI_MODE_AP` would drop the STA link just established.

network_provisioning/src/manager.c
```diff
@@ -2115,11 +2131,23 @@ esp_err_t network_prov_mgr_start_provisioning(network_prov_security_t security,
     prov_ctx->prov_state = NETWORK_PROV_STATE_STARTING;
 #ifdef CONFIG_NETWORK_PROV_NETWORK_TYPE_WIFI
     uint8_t restore_wifi_flag = 0;
-    /* Start Wi-Fi in Station Mode.
-     * This is necessary for scanning to work */
-    ret = esp_wifi_set_mode(WIFI_MODE_STA);
+    prov_ctx->ap_mode_enabled_before_start = false;
+
+    wifi_mode_t wifi_mode_before_start = WIFI_MODE_NULL;
+    ret = esp_wifi_get_mode(&wifi_mode_before_start);
```
shahpiyushv, line 2137: `esp_wifi_get_mode()`'s error goes into `ret` and is overwritten on line 2148, so if it ever fails the AP preservation is silently skipped. Use a separate `esp_err_t` and `ESP_LOGW` on failure.

network_provisioning/src/manager.c
```diff
@@ -2115,11 +2131,23 @@ esp_err_t network_prov_mgr_start_provisioning(network_prov_security_t security,
...
+    }
+
+    /* Start Wi-Fi with station enabled. This is necessary for scanning to work. */
+    wifi_mode_t wifi_mode = prov_ctx->ap_mode_enabled_before_start ?
+                            WIFI_MODE_APSTA :
+                            WIFI_MODE_STA;
+    ret = esp_wifi_set_mode(wifi_mode);
     if (ret != ESP_OK) {
-        ESP_LOGE(TAG, "Failed to set Wi-Fi mode to STA");
+        ESP_LOGE(TAG, "Failed to set Wi-Fi mode");
```
shahpiyushv, line 2150: Dropping "to STA" loses information now that the mode varies. Suggest: `ESP_LOGE(TAG, "Failed to set Wi-Fi mode %d: %s", wifi_mode, esp_err_to_name(ret));`

network_provisioning/src/manager.c
```diff
@@ -2115,11 +2131,23 @@ esp_err_t network_prov_mgr_start_provisioning(network_prov_security_t security,
...
         goto err;
     }
     ret = esp_wifi_start();
```
shahpiyushv, line 2153: Worth confirming: this now starts Wi-Fi with the AP interface enabled. An app that set `APSTA` but hadn't configured `WIFI_IF_AP` or started Wi-Fi yet previously got a working STA-only start; now `esp_wifi_start()` runs with an empty-SSID SoftAP. If the driver rejects that, provisioning start fails outright where it used to succeed.

network_provisioning/src/manager.c
```diff
@@ -2293,6 +2321,10 @@ esp_err_t network_prov_mgr_start_provisioning(network_prov_security_t security,
 err:
     prov_ctx->prov_state = NETWORK_PROV_STATE_IDLE;
 #ifdef CONFIG_NETWORK_PROV_NETWORK_TYPE_WIFI
+    if (ret != ESP_OK && prov_ctx->ap_mode_enabled_before_start) {
+        esp_wifi_set_mode(WIFI_MODE_APSTA);
+        prov_ctx->ap_mode_enabled_before_start = false;
+    }
```
shahpiyushv, line 2327: This is dead code. When the flag is true the mode is never moved away from `APSTA` — start sets `APSTA`, `configure_wifi_sta()` sets `APSTA` — so there is nothing to restore, and the flag reset is moot because `prov_state` goes to `IDLE` and `prov_stop_and_notify()` never runs. `ret != ESP_OK` is also always true here: every `goto err` sets a failure first, and the success path does `goto exit`. Suggest deleting the block — as written it implies a restore is needed.

network_provisioning/src/manager.c
```diff
@@ -146,6 +146,10 @@ struct network_prov_mgr_ctx {
     /* Handle for delayed wifi connection timer */
     esp_timer_handle_t wifi_connect_timer;
 
+    /* Preserve application-owned SoftAP when a STA-only provisioning scheme
+     * runs while Wi-Fi is already in APSTA mode */
+    bool ap_mode_enabled_before_start;
```
shahpiyushv, line 151: Nit: the flag drives whether the SoftAP is kept up *during* provisioning, not just what happens after, so something like `keep_softap_enabled` reads truer than `ap_mode_enabled_before_start`.

PR #765, opened 2026-06-25
Add callback for packet send function to dump the output packets.

dmitry1945:
# Checklist

- [ ] CI passing

# Change description

SystemView currently sends trace data only via RTT. This callback allows firmware to intercept raw packets—for example, to dump them to an alternate channel, integrate with QEMU, or debug without changing the main RTT output path.

Curretn PR adds an optional callback to intercept outgoing SEGGER SystemView packets before they are written to the RTT buffer. New API SEGGER_SYSVIEW_SetSendPacketCallback() lets you subscribe to packet transmission.
The callback is invoked while the packet is being formed, with access to the packet start/end, event ID, and timestamp.

### DIFF
--- a/esp_sysview/src/SEGGER/SEGGER_SYSVIEW.c
+++ b/esp_sysview/src/SEGGER/SEGGER_SYSVIEW.c
@@ -258,6 +258,7 @@ typedef struct {
     U32                     DisabledEvents;
     const SEGGER_SYSVIEW_OS_API  *pOSAPI;
     SEGGER_SYSVIEW_SEND_SYS_DESC_FUNC   *pfSendSysDesc;
+    SEGGER_SYSVIEW_SendPacket_Dump      *pfSendPacketCallback;
 } SEGGER_SYSVIEW_GLOBALS;
 
 /*********************************************************************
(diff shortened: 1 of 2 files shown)


erhankur:
Thanks for the PR, @dmitry1945

`SEGGER_SYSVIEW.c` is vendored upstream code, and we try to keep it as close to the original as possible. All ESP-specific behavior lives under `src/esp/` instead. Adding a hook directly into `_SendPacket()` would diverge from that.

If you feel this callback belongs in the core SystemView library, the cleanest path would be to propose it to SEGGER upstream first. If it is accepted there, we can pull it in with the next sync and keep our copy unmodified.

For the use cases you describe, such as dumping packets to another channel, QEMU integration, or debugging without disturbing the main output, you likely don’t need to modify SEGGER at all. In the `esp_trace` model, SystemView output already flows through the `SEGGER_RTT_WriteSkipNoLock` override in `SEGGER_RTT_esp.c` into a pluggable transport (`esp_trace_transport_t`).

So “send the packets somewhere else” is exactly what a new transport adapter is for. You can implement your sink there and get destination-CPU filtering, link-type handling, and locking for free, without patching vendored code.

PR #773, opened 2026-07-07
feat(esp_ext_part_tables): Improve MBR generation: alignment policy, layout validation, auto-placement, etc., version bump to 0.4.0

adokitkat:
Overhauls `esp_ext_part_tables` so `esp_mbr_generate` produces valid, verifiable
MBRs (instead of silently-broken ones) and `esp_mbr_parse` reports every recognized
partition, with predicate-based selection for filtering.

## MBR generation

- **Layout validation** - overlapping partitions and partitions running past the end
  of the disk are always rejected. `total_size` bounds the disk; it is auto-filled
  from the block-device geometry in `esp_ext_part_list_bdl_write`.
- **`align_policy`** - `KEEP_SIZE` (default), `REJECT`, `PRESERVE_END` control how a
  partition's size is handled when alignment shifts its start.
- **Automatic placement** - `ESP_EXT_PART_FLAG_AUTO_ADDRESS` computes aligned start
  offsets and `ESP_EXT_PART_FLAG_FILL` sizes the last partition to the disk end.
  Resolved on internal copies, so the caller's list is never mutated.
- **Empty list items** - a `ESP_EXT_PART_TYPE_NONE` item can't be encoded without
  leaving a gap that truncates the parsed table and misplaces auto-partitions, so it
  is now rejected (`ESP_ERR_INVALID_ARG`).
- **Alignment ergonomics** - `ESP_EXT_PART_ALIGN_AUTO` (default -> 1 MiB) and
  `ESP_EXT_PART_ALIGN_NONE` make "default" and "no alignment" explicit;
  `esp_mbr_lba_align` now handles any alignment/sector-size ratio (not just powers of
  two).

## MBR parsing & partition selection

- **All recognized partitions are parsed** - the old binary "supported/unsupported"
  drop behavior is gone. exFAT/NTFS, Linux and GPT-protective partitions are no
  longer silently dropped; only truly unknown/extended type codes are skipped.
- **New type `ESP_EXT_PART_TYPE_RAW_DATA` (0xDA)** for non-filesystem/custom-data
  partitions the application manages itself.
- **Predicate-based selection** - `esp_ext_part_match_t { fn; ctx; }` bundles a
  caller predicate with its context. `esp_ext_part_list_next_matching()` iterates the
  items a predicate accepts, and `esp_mbr_parse_extra_args_t.match` filters at parse
  time (a zero-initialized `match` inserts everything). Mountability depends on which
  filesystem drivers are linked into the firmware, so it is a caller decision rather
  than a fixed library classification.
- **Stock matcher** - `esp_ext_part_match_mountable()` returns a ready-to-use matcher:
  FAT12/16/32 are always mountable; LittleFS is mountable when its component is part
  of the build (auto-detected in CMake, or forced via `ESP_EXT_PART_HAS_LITTLEFS`),
  otherwise reported not-mountable (a safe under-report).
- **`ESP_EXT_PART_LIST_FLAG_LOSSY`** is set when any partition is skipped (unknown/
  extended type, or rejected by the parse `match` predicate), signaling that a
  regenerated MBR would not be functionally equivalent to the source.

## Other

- Reordered `esp_mbr_generate_extra_args_t` to minimize padding; a zero-initialized
  struct still selects all defaults.
- Fixed enum portability and a potential LBA overflow at the top of the address space.
- Upgraded the basic example to showcase auto-placement, `FILL`, the `0xDA` type,
  the mountable matcher, and the `LOSSY` flag.
- Minor example/README/license fixes.

Covered by host-side unit tests (linux target). Version bumped to 0.4.0.

**ABI note (pre-1.0):** `esp_mbr_generate_extra_args_t` and
`esp_mbr_parse_extra_args_t` changed shape; source-compatible via designated
initializers. Behavior change: `esp_mbr_parse` now returns exFAT/NTFS, Linux and
GPT-protective partitions as list items instead of dropping them - filter with a
predicate via `esp_mbr_parse_extra_args_t.match` /
`esp_ext_part_list_next_matching()` (e.g. `esp_ext_part_match_mountable()`).

# Checklist

- [x] CI passing

### DIFF
--- a/esp_ext_part_tables/LICENSE
+++ b/esp_ext_part_tables/LICENSE
@@ -1,3 +1,4 @@
+
                                  Apache License
                            Version 2.0, January 2004
                         http://www.apache.org/licenses/
@@ -178,7 +179,7 @@
    APPENDIX: How to apply the Apache License to your work.
 
       To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "{}"
+      boilerplate notice, with the fields enclosed by brackets "[]"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
@@ -186,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 Thesis projects
+   Copyright [yyyy] [name of copyright owner]
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/esp_ext_part_tables/examples/basic/main/main.c
+++ b/esp_ext_part_tables/examples/basic/main/main.c
@@ -22,6 +22,10 @@
 
 static const char *TAG = "esp_ext_part_tables_example_basic";
 
+// A fixed disk size used for the generation demo so that FILL and the disk-bounds
+// check have a known total to work with (16 MiB).
+#define EXAMPLE_DISK_SIZE_BYTES (16 * 1024 * 1024)
+
 void print_loaded_ext_partitions(esp_ext_part_list_item_t *head)
 {
     esp_ext_part_list_item_t *it = head;
@@ -34,6 +38,26 @@ void print_loaded_ext_partitions(esp_ext_part_list_item_t *head)
                parsed_type_to_str(it->info.type));
         i++;
     } while ((it = esp_ext_part_list_item_next(it)) != NULL);
+    fflush(stdout);
+}
+
+// List only the partitions this build can mount, using the stock mountable predicate.
+static void print_mountable_partitions(esp_ext_part_list_t *part_list)
+{
+    ESP_LOGI(TAG, "Mountable partitions (via esp_ext_part_list_next_matching):");
+    int count = 0;
+    esp_ext_part_match_t matcher = esp_ext_part_match_mountable();
+    for (esp_ext_part_list_item_t *it = esp_ext_part_list_next_matching(NULL, part_list, &matcher);
+            it != NULL;
+            it = esp_ext_part_list_next_matching(it, part_list, &matcher)) {
+        printf("\t- %s at sector %" PRIu64 "\n",
+               parsed_type_to_str(it->info.type),
+               esp_ext_part_bytes_to_sector_count(it->info.address, ESP_EXT_PART_SECTOR_SIZE_512B));
+        count++;
+    }
+    if (count == 0) {
+        printf("\t(none)\n");
+    }
     fflush(stdout);
 }
 
@@ -59,7 +83,8 @@ void esp_ext_part_tables_mbr_parse_example_task(void *pvParameters)
     }
     ESP_LOGI(TAG, "MBR loaded successfully");
 
-    // Parse the MBR to get the partition list
+    // Parse the MBR to get the partition list. With extra_args == NULL, every
+    // recognized partition is inserted (mountable or not).
     esp_ext_part_list_t part_list = {0};
     err = esp_mbr_parse((void *) mbr, &part_list, NULL);
     if (err != ESP_OK) {
@@ -70,6 +95,14 @@ void esp_ext_part_tables_mbr_parse_example_task(void *pvParameters)
     }
     free(mbr); // Free the MBR buffer after parsing as it is no longer needed
     ESP_LOGI(TAG, "MBR parsed successfully");
+
+    // If the parser skipped any partition (unknown/extended type), the list is
+    // marked LOSSY - regenerating it would not reproduce the source table.
+    if (part_list.flags & ESP_EXT_PART_LIST_FLAG_LOSSY) {
+        ESP_LOGW(TAG, "Parsed list is LOSSY: some source partitions were not captured");
+    } else {
+        ESP_LOGI(TAG, "Parsed list is complete (not LOSSY)");
+    }
 
     // Get the first partition
     esp_ext_part_list_item_t *it = esp_ext_part_list_item_head(&part_list);
@@ -101,19 +134,22 @@ void esp_ext_part_tables_mbr_generate_example_task(void *pvParameters)
     esp_err_t err;
     esp_ext_part_list_t part_list = {0};
 
+    // A zero-initialized args struct already selects sensible defaults
+    // (AUTO alignment => 1 MiB, KEEP_SIZE policy, 512 B sectors). Here we set only
+    // what the auto-placement/FILL demo needs: an explicit sector size and the
+    // total disk size (so FILL and the bounds check know where the disk ends).
     esp_mbr_generate_extra_args_t mbr_args = {
         .sector_size = ESP_EXT_PART_SECTOR_SIZE_512B,
-        .alignment = ESP_EXT_PART_ALIGN_1MiB
+        .total_size = EXAMPLE_DISK_SIZE_BYTES,
+        // .alignment left 0 (ESP_EXT_PART_ALIGN_AUTO) -> resolves to 1 MiB
     };
 
-    // 2 FAT12 partitions (random parameters)
+    // Partition 1: auto-placed (library picks the aligned start), fixed 4 MiB FAT32.
     esp_ext_part_list_item_t item1 = {
         .info = {
-            // Original MBR starts at 2048, but we use 8 for testing ->
-            .address = esp_ext_part_sector_count_to_bytes(8, mbr_args.sector_size), // Should be round up to 2048 sectors (aligned to 1MiB) due to defined sector size and alignment in `esp_mbr_generate_extra_args_t args` above
-            .size = esp_ext_part_sector_count_to_bytes(7953, mbr_args.sector_size),
-            .type = ESP_EXT_PART_TYPE_FAT12,
-            .label = NULL,
+            .size = 4 * 1024 * 1024,
+            .type = ESP_EXT_PART_TYPE_FAT32,
+            .flags = ESP_EXT_PART_FLAG_AUTO_ADDRESS,
         }
     };
     err = esp_ext_part_list_insert(&part_list, &item1);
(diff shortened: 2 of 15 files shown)


adokitkat:
@copilot review

esp_ext_part_tables/include/esp_mbr.h
```diff
@@ -75,10 +75,16 @@ typedef struct {
...
+    uint64_t total_size; // Total device size in bytes for the "fits within disk" check; 0 disables the check. The BDL write helper auto-fills this from the device geometry when left 0.
+    uint8_t (*esp_mbr_generate_custom_supported_partition_types)(uint8_t); // Custom function for generating supported MBR partition types, optional
+    esp_ext_part_sector_size_t sector_size; // Sector size for correct LBA alignment. Overrides the list's `sector_size` for this call; 0 (UNKNOWN) falls back to the list's value, then to 512 B.
     esp_ext_part_align_t alignment; // Alignment hint for correct LBA alignment
+    esp_ext_part_align_policy_t align_policy; // Policy applied when alignment moves a partition start (default 0 = KEEP_SIZE)
     bool keep_signature; // If true, the disk signature will be preserved in the generated MBR and not overwritten with a random value
-    uint8_t (*esp_mbr_generate_custom_supported_partition_types)(uint8_t); // Custom function for generating supported MBR partition types, optional
+    bool disable_overlap_check; // If false (default), generation fails when two partitions overlap
```
haberturdeur: Is there a use-case where you wouldn't want to do the check?

esp_ext_part_tables/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "0.3.0"
+version: "0.6.1"
```
haberturdeur: Did we really skip all these minor versions?

esp_ext_part_tables/LICENSE
```diff
@@ -178,15 +179,15 @@
...
+      boilerplate notice, with the fields enclosed by brackets "[]"
       replaced with your own identifying information. (Don't include
       the brackets!)  The text should be enclosed in the appropriate
       comment syntax for the file format. We also recommend that a
       file or class name and description of purpose be included on the
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2020 Thesis projects
+   Copyright [yyyy] [name of copyright owner]
```
haberturdeur, line 190: Needs to be updated

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -34,11 +34,21 @@ typedef enum {
 } esp_ext_part_sector_size_t;
 
 typedef enum {
-    ESP_EXT_PART_ALIGN_NONE = 0, // No alignment
+    ESP_EXT_PART_ALIGN_AUTO = 0, // Use the library default alignment (1 MiB). This is what a zero-initialized extra_args selects.
     ESP_EXT_PART_ALIGN_4KiB = 4096, // 4 KiB alignment
     ESP_EXT_PART_ALIGN_1MiB = (1024 * 1024), // 1 MiB alignment
+    ESP_EXT_PART_ALIGN_NONE = -1, // Explicitly perform no alignment (leave LBAs untouched). A representable-in-int sentinel (not a real alignment value).
```
RathiSonika, line 40: Should we add a short migration note in the README for ESP_EXT_PART_ALIGN_NONE?
On master it was 0, but in now it effectively meant “use the 1 MiB default,” not “no alignment.” In this PR it is -1 and now actually means “no alignment,” while ESP_EXT_PART_ALIGN_AUTO (0) selects the default.

esp_ext_part_tables/idf_component.yml
RathiSonika: Also, should we align the PR title with the actual bump and explain why intermediate minor versions were skipped?

esp_ext_part_tables/include/esp_ext_part_tables.h
adokitkat: Since the library is not v1.0.0 yet and there are also pretty much no users yet according to stats in component registry I just made this change. I will introduce a changelog from version 1.0.0.

esp_ext_part_tables/idf_component.yml
adokitkat: I made a new version after each big change and minor for unrelated fixes. If these all had their own PRs there would be a version bump after each one, but I just put it in a single PR so it can be merged faster.

esp_ext_part_tables/include/esp_mbr.h
adokitkat: If I would want to generate MBR with errors on purpose - e.g. for tests or research.

esp_ext_part_tables/LICENSE
adokitkat: Funnily enough this is how the license looks in other components here...

esp_ext_part_tables/idf_component.yml
haberturdeur: In that case I still there should only be one bump per MR as it represents a set of changes. (in theory there should be a tag for each bump AFAIK)

esp_ext_part_tables/include/esp_mbr.h
haberturdeur: I don't think testing API should be exposed like this, perhaps a second function in `esp_private/` that is called by the public wrapper, which is the one containing the checks?

We should at least document it as a testing feature that should not be used under normal circumstances.

esp_ext_part_tables/LICENSE
haberturdeur: Fair enough, but stilll....

igrr: This does not need to be updated. This line is part of the appendix: how to apply this license to a copyrighted work, PTAL a few lines above.

https://github.com/espressif/esp-idf/issues/9470#issuecomment-1202120785

esp_ext_part_tables/idf_component.yml
adokitkat: Ok will change it to 0.4.0

esp_ext_part_tables/include/esp_mbr.h
adokitkat: I checked this part again an I decided to remove these options:

- `disable_overlap_check` shouldn't be an option because it is not allowed for MBR's primary partitions to have overlapping partitions anyway, so every MBR with overlapping partition is marked as corrupted by tools like `parted` or `fdisk`
- `allow_empty_partitions` was removed as well because it allowed our partition list like [p1, NONE, p2] to translate to [p1, p2] MBR partitions. `0x00` MBR partition type cannot mean an "unknown"/custom type, it always means empty or unallocated + there are no holes allowed as well (as it signifies the end of the partition table as well) -> meaning if there were these entries: (0x01 [first FAT12], 0x00, 0x01 [second FAT12]) then the second FAT would be dropped.

There is a way to customize known partitions parser function by the users and I also added a new recognized MBR partition type (`0xDA`) which can be used for raw/custom data if needed. 

If I missed something and one of these or both are actually needed/useful in some cases they can be always added in a later version.

esp_ext_part_tables/src/esp_mbr_utils.c
```diff
@@ -163,3 +188,23 @@
...
+        return ESP_EXT_PART_USAGE_MOUNTABLE;
+    case ESP_EXT_PART_TYPE_RAW_DATA:
+        return ESP_EXT_PART_USAGE_RAW;
+    case ESP_EXT_PART_TYPE_EXFAT_OR_NTFS:
+    case ESP_EXT_PART_TYPE_LINUX_ANY:
+    case ESP_EXT_PART_TYPE_GPT_PROTECTIVE_MBR:
+        return ESP_EXT_PART_USAGE_UNSUPPORTED;
+    case ESP_EXT_PART_TYPE_NONE:
+    default:
+        return 0; // No class (empty or unknown type)
```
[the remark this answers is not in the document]
adokitkat: fixed

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -49,6 +50,7 @@
     // Set defaults
     part_list->sector_size = ESP_EXT_PART_SECTOR_SIZE_512B; // Default sector size
     bool (*f_parse_supported_partition_types)(uint8_t, uint8_t *) = esp_mbr_parse_default_supported_partition_types;
+    esp_ext_part_usage_t usage_mask = ESP_EXT_PART_USAGE_ALL; // Default: insert all recognized partitions
```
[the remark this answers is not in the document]
adokitkat: fixed

esp_ext_part_tables/examples/basic/main/main.c
```diff
@@ -79,7 +112,7 @@ void esp_ext_part_tables_mbr_parse_example_task(void *pvParameters)
         goto end_task;
     }
 
-    // Print the loaded partition list
+    // Print the loaded partition list (now including usage classification)
```
haberturdeur:
```suggestion
    // Print the loaded partition list (including usage classification)
```

esp_ext_part_tables/include/esp_ext_part_tables.h
```diff
@@ -56,13 +67,31 @@ typedef enum {
...
 
+/**
+ * @brief Usage classification of a partition, based on its type.
+ *
+ * Bit-flags so classes can be OR'd together (e.g. for `usage_filter` or
+ * `esp_ext_part_list_next_by_usage`).
+ */
+typedef enum {
+    ESP_EXT_PART_USAGE_NONE        = 0,      /*!< No usage class (empty or unknown partition type). */
+    ESP_EXT_PART_USAGE_MOUNTABLE   = 1 << 0, /*!< Known filesystem with an ESP-IDF driver (FAT12/16/32, LittleFS). */
```
haberturdeur: LittleFS should not be included unless we are sure that littlefs component is present, instead it should be  under UNSUPPORTED in that case.

Instead of this enum, perhaps the user could pass a predicate to the filter function?

adokitkat: At first I also wanted to do a predicate search but for four partitions I just don't think it is necessary... That's why I settled on this enum - the API should be somewhat easy for the user to use, i.e. get the info and mount. It essentially is just "a tip". If we were to support GPT I can see it making more sense. However then we would probably update the major version so we can make breaking changes. But on the other hand it would be better to not change the API if we don't have to and can do it now. I will think about it.

adokitkat: Changed the API to use `esp_ext_part_list_next_matching`

haberturdeur (approved):
LGTM! Thank you

RathiSonika (approved):
LGTM. Thank you!

PR #776, opened 2026-07-10
fix(esp_cli): register VFS with context-pointer flag in test app

mahavirj:
Since IDF v6.0 (backported to release/v6.0 in ece7bf6af6c), the ops tables returned by esp_vfs_uart_get_vfs() and esp_vfs_usb_serial_jtag_get_vfs() hold context-pointer function variants. Registering them without ESP_VFS_FLAG_CONTEXT_PTR makes the VFS layer call them through the context-less union member, shifting all arguments: uart_open() then receives the open() flags (0) as its path and crashes in ROM strcmp() with a LoadProhibited panic on NULL.

Register with ESP_VFS_FLAG_CONTEXT_PTR on IDF v6.0+ and keep the context-less registration on older versions, which still return context-less ops tables.

# Checklist

- [ ] CI passing

# Change description

### DIFF
--- a/esp_cli/test_apps/main/test_esp_cli.c
+++ b/esp_cli/test_apps/main/test_esp_cli.c
@@ -18,6 +18,7 @@
 #include "esp_cli_commands.h"
 
 #include "sdkconfig.h"
+#include "esp_idf_version.h"
 #include "esp_vfs.h"
 #include "esp_vfs_common.h"
 #include "driver/esp_private/uart_vfs.h"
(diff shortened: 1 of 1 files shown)


esp_cli/test_apps/main/test_esp_cli.c
```diff
@@ -27,6 +28,15 @@
 #include "driver/usb_serial_jtag_vfs.h"
 #include "driver/usb_serial_jtag.h"
 
+/* Since IDF v6.0 the tables returned by esp_vfs_uart_get_vfs() and
+ * esp_vfs_usb_serial_jtag_get_vfs() hold context-pointer function variants,
+ * so they must be registered with ESP_VFS_FLAG_CONTEXT_PTR. */
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(6, 0, 0)
+#define TEST_VFS_REGISTER_FLAGS ESP_VFS_FLAG_CONTEXT_PTR
```
RathiSonika, line 35: Nit: consider (ESP_VFS_FLAG_STATIC | ESP_VFS_FLAG_CONTEXT_PTR) on IDF 6.0+ — the ops tables are already static, so without STATIC VFS flag, it will heap-copy them unnecessarily.

RathiSonika (approved):
LGTM. Thank you!

esp_cli/test_apps/main/test_esp_cli.c
mahavirj: Okay. But given this is a test app, I will keep only the mandatory part. I think IDF test apps also use similar pattern. Maybe we should keep this flag always enabled and let someone override to HEAP version if required ..

PR #783, opened 2026-07-21
fix(esp_schedule): correct date engine and timer overflow (IEC-563)

Junius00:
# Checklist

- [x] CI passing

# Change description

Fixes correctness bugs in the `esp_schedule` date/solar next-occurrence logic and a timer overflow, and adds a `test_app` with unit tests.

## Bug fixes

**Date engine rewrite.** Replaced the per-type ad-hoc next-time calculation with a single unified engine (`esp_schedule_get_next_date_time`) shared by date and solar triggers. It scans forward (up to 370 days) matching all constraints:
- Time of day (`hours`/`minutes`)
- Day-of-week mask (`day.repeat_days`)
- Day-of-month (`date.day`)
- Months-of-year mask (`date.repeat_months`)
- Specific year (`date.year` + `date.repeat_every_year`)

Day-of-week and day-of-month combine with **OR**; month and year act as additional filters. This adds flexibility (combined weekday/monthday/month/year patterns, one-shot specific-date triggers) that the previous per-type paths could not express.

**Timer overflow fix.** Old code computed the timer period as `(diff * 1000) / portTICK_PERIOD_MS` in 32-bit, which overflowed for large diffs (long-horizon schedules fired early). Period is now computed in 64-bit and clamped to `portMAX_DELAY - 1`; `esp_schedule_common_timer_cb` re-arms for the remaining time when a diff exceeds one timer period.

**One-shot handling.** Added `esp_schedule_trigger_fired_and_done` so a fired one-shot trigger is not recomputed to a future occurrence.

**Solar triggers** now honor the same day/month/year filters when computing the next valid sunrise/sunset, plus `solar.offset_minutes`.

## Other

- Header: added `extern "C"` guards, exposed internal functions for unit testing, updated copyright to 2025-2026.
- README: documented date-based and solar trigger patterns with examples.
- `test_app` with 23 unit tests covering the date/solar engine, DST (spring-forward / fall-back compliance), validity windows, and one-shot fired-and-done detection. Registered in `.build-test-rules.yml` and `test_app/CMakeLists.txt`; `pytest_esp_schedule.py` runs them.

### DIFF
--- a/esp_schedule/src/esp_schedule.c
+++ b/esp_schedule/src/esp_schedule.c
@@ -1,9 +1,12 @@
 /*
- * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2025-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 #include <string.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <time.h>
 #include <inttypes.h>
 #include "esp_log.h"
 #include "esp_sntp.h"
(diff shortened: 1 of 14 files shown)


esp_schedule/test_app/main/test_app_main.c
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "unity.h"
+#include "esp_err.h"
+#include "nvs_flash.h"
+#include "esp_schedule_internal.h"
+
```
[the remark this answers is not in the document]
Junius00: Fixed. Added `#include <stdio.h>` to `test_app_main.c`.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,426 @@
...
+    ESP_LOGI(TAG, "Schedule %s will be active on: %s. DST: %s", schedule->name, time_str, schedule_time.tm_isdst ? "Yes" : "No");
 
-    /* Calculate difference */
-    time_diff = difftime((mktime(&schedule_time)), mktime(&current_time));
-
-    /* For one time schedules to check for expiry after a reboot. If NVS is enabled, this should be stored in NVS. */
-    trigger->next_scheduled_time_utc = mktime(&schedule_time);
-
-    return time_diff;
+    return (uint32_t)difftime(schedule->trigger.next_scheduled_time_utc, now);
```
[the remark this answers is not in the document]
Junius00: Fixed defensively. The diff is now clamped to `[0, UINT32_MAX]` before the `uint32_t` cast, since casting an out-of-range `double` is undefined behavior.

In practice the diff cannot exceed `UINT32_MAX` (date/solar targets are bounded to ~370 days by the engine; RELATIVE targets are bounded by the `uint32_t` relative-seconds field), so this is hardening rather than a reachable overflow, but the clamp makes the intent explicit and safe.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,426 @@
...
+    default:
+        return true;
+    }
+}
+
+/*
+ * Ensure trigger->next_scheduled_time_utc is set to the next occurrence.
+ * If it's already set in the future, it is reused and not recomputed.
+ * Returns true if a valid future time is present after this call, false otherwise.
+ */
```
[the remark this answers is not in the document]
Junius00: Good catch — the header comment was stale. Rewrote it to match the implementation: repeating date/day-of-week/solar triggers are always recomputed on every arm (so a timezone change is picked up), RELATIVE keeps its computed absolute target, and already-fired one-shots are left untouched.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,426 @@
...
-        const char *schedule_name)
+/**
+ * @brief Calculate solar time for a given time in UTC
+ * @param is_sunrise: true if sunrise, false if sunset
+ * @param time_utc: time in UTC
+ * @param latitude: latitude
+ * @param longitude: longitude
+ * @param offset_minutes: offset in minutes
+ * @return solar time in UTC, 0 if calculation failed
+ */
```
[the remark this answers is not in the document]
Junius00: Using `localtime_r` here is intentional, not a bug: scheduling is wall-clock based, so the solar event is computed for the *local* calendar date. Clarified the doc comment to say the calendar date is derived in local time by design, and that the returned solar time is UTC.

shahpiyushv (approved):
LGTM! Verified the date engine against a host harness in UTC and America/New_York — DST correct in both directions (including the near-midnight day-walk), day-31/Feb-29 handled, one-shot semantics preserved, far-future validity windows found, and the timer clamp + re-arm covers long delays properly. Also good to see the NULL `schedule_count` contract and the months-mask one-shot behavior covered by tests, and no changes to the API or NVS format.

One non-blocking note: a DATE trigger with a `year` but no months mask (`dd` + `yy`, no `mm`) now fires on that day of every remaining month of the bounded year, where 1.3.3 treated it as one-shot. No sane client sends that combination (one-shot needs only `dd`, and a specific date includes `mm`), and the new reading is the more literal one — but since it's technically a behavior change, a sentence in the README or changelog documenting the intended semantics would be cheap insurance.

Junius00:
I will draft a specification document to make sure the new trigger rules are clear and semantically consistent with what we expect, and then re-open for review once complete

Junius00:
Bumped to `1.4.0`, because this adds/modifies existing functionality with the same API format

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,455 @@
...
+    }
+    switch (trigger->type) {
+    case ESP_SCHEDULE_TYPE_RELATIVE:
+        return true; /* relative schedules fire exactly once */
+    case ESP_SCHEDULE_TYPE_DAYS_OF_WEEK:
+        return trigger->day.repeat_days == ESP_SCHEDULE_DAY_ONCE;
+#if CONFIG_ESP_SCHEDULE_ENABLE_DAYLIGHT
+    /* Solar schedules share DATE's one-shot semantics; only the time-of-day is
+     * replaced by the computed sunrise/sunset instant. */
+    case ESP_SCHEDULE_TYPE_SUNRISE:
```
vikramdattu, line 496: Add /* fall-through */ comments when block does not have anything to avoid compiler crying.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,455 @@
...
+ * once its next_scheduled_time_utc is set (>0) and has passed (<=now).
+ *
+ * Repeating triggers (weekly day-of-week, yearly dates, repeating solar) return
+ * false so they are recomputed and re-armed. Without this guard the date engine
+ * treats an empty day/month mask as "any", so a DAY_ONCE schedule would re-fire
+ * every day and a one-time DATE schedule every month.
+ */
+bool esp_schedule_trigger_fired_and_done(const esp_schedule_trigger_t *trigger, time_t now)
+{
+    if (!(trigger->next_scheduled_time_utc > 0 && trigger->next_scheduled_time_utc <= now)) {
```
vikramdattu: This guard doesn't survive a reboot, and the failure mode is re-firing.

`esp_schedule_nvs_add()` is only called from `esp_schedule_set()` (create/edit), so the *persisted* `next_scheduled_time_utc` stays `0` forever for non-relative schedules — it's never re-written when the schedule arms or fires. After a reboot, a `DAY_ONCE` (or one-shot DATE) schedule that **already fired** comes back from NVS with `next == 0`, this check reads it as "not computed yet", the engine recomputes a fresh future occurrence, and `esp_schedule_init()` happily re-arms it. A bare one-shot DATE (`day` set, no months, no year) is the worst case: the engine matches that day-of-month in *any* month, so every reboot resurrects it for the following month.

v1 had the opposite bug (`next == 0` counted as expired, so a one-shot that had *never* fired got deleted at boot) — this rewrite fixes that direction but opens this one. For a scheduler, firing again is arguably worse than disappearing.

Suggested fix: persist the armed state — e.g. call `esp_schedule_nvs_add(schedule)` from `esp_schedule_start_timer()` after the diff is computed (it re-writes the whole blob, so it's a one-liner). Then a reboot sees the real armed target: past + one-shot ⇒ `fired_and_done` ⇒ deleted at init, exactly as this function intends. Cost is one NVS write per (re)arm, which is negligible at schedule frequencies. A test that simulates the reboot path (recompute from a persisted-as-fired trigger) would lock the behavior in.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,455 @@
...
+            return false;
+        }
+        /* A months mask keeps the schedule firing through the remaining masked
+         * months (matching v1 is_expired behavior: e.g. "16th of Jun+Jul+Aug"
+         * fires each of those months before expiring), so it is a true one-shot
+         * only when no months are set. A specific year is bounded by the engine. */
+        if (trigger->date.repeat_months != 0) {
+            return false;
+        }
+        return trigger->date.year == 0;
```
vikramdattu: Semantics question: with `year` set but `repeat_months == 0`, this returns `false` (not done) — so "day=27, year=2026, no month mask" keeps matching the 27th of **every remaining month of 2026** after it first fires. The year does bound the schedule, but at year *end*, not at one occurrence — users writing "fire once on this specific date" will likely expect one shot.

If the intent of `year != 0` without a months mask is "specific date, once", this should probably be `return trigger->date.repeat_months == 0;` (one-shot whenever there's no recurring pattern, year or not). If the monthly-until-year-end behavior is intentional, it deserves an explicit example in `docs/trigger_rules.md` — it's surprising enough to generate bug reports.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,455 @@
...
-    /* Adjust schedule day */
-    if (trigger->type == ESP_SCHEDULE_TYPE_DAYS_OF_WEEK) {
-        int no_of_days = 0;
-        no_of_days = esp_schedule_get_no_of_days(trigger, &current_time, &schedule_time);
-        schedule_time.tm_sec += no_of_days * SECONDS_IN_DAY;
-    }
+    /* Unified DATE and DAYS_OF_WEEK using date finder */
+    time_t next_time = 0;
+    bool ok = false;
+    uint16_t minutes_since_midnight = (uint16_t)(trigger->hours * 60 + trigger->minutes);
```
vikramdattu, line 597: No range validation on `hours`/`minutes` before this: with `hours = 25`, `mktime()` silently normalizes the candidate into the *following* day, so the returned instant can land on a day that violates the mask the engine just matched. A cheap guard here (or in `esp_schedule_create/edit`) closes it:

```c
if (trigger->hours > 23 || trigger->minutes > 59) {
    return false;
}
```

Pre-existing in v1 too, but the unified engine makes it a one-line fix now.

esp_schedule/src/esp_schedule.c
```diff
@@ -13,539 +16,455 @@
...
-        /* First scheduled day of the next week */
-        next_day = ffs(repeat_days) - 1;
-        if (next_day == today) {
-            /* Same day, next week */
-            return 7;
+    candidate_tm.tm_isdst = -1;
+    time_t candidate_time = mktime(&candidate_tm);
+    localtime_r(&candidate_time, &candidate_tm);
+
+    for (int month_attempts = 0; month_attempts < 25; month_attempts++) {
```
vikramdattu: Nit: the `25` reads arbitrary — it's actually a sound bound (each outer iteration lands on one *masked* month, and 25 candidate months covers the worst "Feb 29 in a leap year" gap of 8 years… via the month-skip do-while below). Worth a one-line comment stating that rationale so the next reader doesn't have to re-derive it.

esp_schedule/src/esp_schedule.c
```diff
@@ -631,8 +554,9 @@ static void esp_schedule_delete_timer(esp_schedule_t *schedule)
 static void esp_schedule_create_timer(esp_schedule_t *schedule)
 {
     if (esp_schedule_nvs_is_enabled()) {
-        /* This is just used for calculating next_scheduled_time_utc for ESP_SCHEDULE_DAY_ONCE (in case of ESP_SCHEDULE_TYPE_DAYS_OF_WEEK) or for ESP_SCHEDULE_MONTH_ONCE (in case of ESP_SCHEDULE_TYPE_DATE), and only used when NVS is enabled. And if NVS is enabled, time will already be synced and the time will be correctly calculated. */
-        schedule->next_scheduled_time_diff = esp_schedule_get_next_schedule_time_diff(schedule->name, &schedule->trigger);
+        /* This is used for calculating next_scheduled_time_utc, and only used when NVS is enabled.
+         * If NVS is enabled, time will already be synced and the time will be correctly calculated. */
+        schedule->next_scheduled_time_diff = esp_schedule_get_next_schedule_time_diff(schedule);
```
vikramdattu, line 772: Nit: this pre-computation looks vestigial — every caller of `esp_schedule_create_timer()` follows up with `esp_schedule_start_timer()`, which recomputes `next_scheduled_time_diff` from scratch (and `esp_schedule_init()` additionally computed it once *before* calling this). Three computations per schedule at boot; dropping this one loses nothing.

vikramdattu (commented):
Nice work overall — the unified date engine is a big step up from the per-type paths, and I went looking for problems in exactly the places rewrites like this usually break and found them handled correctly: the Monday=bit0 ↔ `(tm_wday+6)%7` mapping is consistent, day advancement via `tm_mday++` + `mktime` re-normalization (instead of +86400) is the right DST-safe idiom, the solar path's "23:59 anchor day, then `day_end + 1`" trick is correct, `force_include_first` properly avoids both the month-cap exhaustion and the DST-boundary skip for future validity windows, and the 64-bit tick math with the `portMAX_DELAY - 1` clamp plus the early-fire re-arm guard closes the overflow for real. The test app is a very welcome addition.

Inline comments, by weight:

1. **One-shot schedules resurrect after reboot** (`esp_schedule_trigger_fired_and_done`) — the fired state is never persisted (`esp_schedule_nvs_add` only runs at create/edit), so a fired `DAY_ONCE`/one-shot DATE re-arms from NVS on the next boot. v1 failed in the opposite direction (deleted never-fired one-shots at boot); this direction re-fires, which I'd consider worse. Concrete one-line fix suggested inline. This is the only item I'd consider gating.
2. **Semantics**: DATE with `year` set but no months mask fires monthly through year-end after the first occurrence — intentional? Either tighten `fired_and_done` or document it.
3. Smaller: missing `hours`/`minutes` range validation (mktime normalization can defeat the day mask), a rationale comment for the 25-month cap, and a vestigial recompute in `esp_schedule_create_timer`.

esp_schedule/src/esp_schedule.c
Junius00: Added `/* fall-through */` on the `SUNRISE` / `SUNSET` labels.

For the record, no compiler was actually complaining here: `-Wimplicit-fallthrough` only fires when there are *statements* between case labels, and these are consecutive empty labels (only an `#endif` sits between `SUNSET` and `DATE`). Added anyway since it makes the intent explicit to a reader.

esp_schedule/src/esp_schedule.c
Junius00: Valid, fixed. `hours` / `minutes` are `uint8_t` and `esp_schedule_create()` validated only the name and type, so `hours = 25` reached the engine and `mktime()` normalized the candidate into the following day — off the day/month mask that had just been matched.

Guarded in two places:

1. `esp_schedule_get_next_date_time()` rejects `minutes_since_midnight >= MINUTES_IN_DAY`, so the engine can never return an instant outside the day it matched, whatever the caller passes.
2. `esp_schedule_config_time_of_day_is_valid()` rejects `hours > 23` / `minutes > 59` in both `esp_schedule_create()` (returns `NULL`) and `esp_schedule_edit()` (returns `ESP_ERR_INVALID_ARG`), so a bad config fails loudly at the API boundary rather than silently firing on the wrong day.

The API check skips `RELATIVE`, which does not use those fields.

esp_schedule/src/esp_schedule.c
Junius00: Good nit — the reasoning was worth writing down, and doing so showed 25 was larger than needed. Replaced the literal with a derived constant:

```c
/* Longest gap between two leap years, in years. Normally 4, but a century that
 * is not a leap year stretches it to 8: 2096 and 2104 are leap, 2100 is not. */
#define MAX_LEAP_YEAR_GAP 8
#define MAX_MASKED_MONTH_ATTEMPTS (1 + (MAX_LEAP_YEAR_GAP - 1) + 1)
```

The bound needed is `1 + N + 1`: one attempt for the starting month (which may be partially elapsed), `N` for the longest run of consecutive *masked* months in which `day_of_month` does not exist, and one for the month that matches. Only `day_of_month` can be missing from a month — a `days_of_week_mask` always matches, since every month contains all seven weekdays.

`N` is maximized by the narrowest mask over the rarest day:

| pattern | N |
|---|---|
| `day <= 28`, or any dow mask | 0 |
| `day 31`, no months mask | 1 (Feb, Apr, Jun, Sep, Nov all lack a 31st, but none are adjacent) |
| `day 29`, mask = Feb only | **7** — the non-leap Februaries 2097..2103 inside the 2096 → 2104 gap |

So 9, verified by tracing the worst case: starting late on 29 Feb 2096 with the target time already past, attempt 0 skips today, attempts 1..7 walk Feb 2097..2103, attempt 8 matches 29 Feb 2104.

Unsatisfiable patterns (`day 30` masked to February, `day 31` masked to only 30-day months) have no finite `N` — they exhaust the horizon and fail whatever its value, so the lower bound just halves their wasted work.

Full derivation is in the comment above the macro, including the note that adding an "nth weekday of month" rule would need a new term in the sum rather than a larger `MAX_LEAP_YEAR_GAP`. The docs' "search horizon" wording is updated to match.

esp_schedule/src/esp_schedule.c
Junius00: Partly — it is redundant for the `init` path, as you say, but dropping it outright would change behaviour for `RELATIVE`.

`RELATIVE` is computed once and then reused (`if (trigger->next_scheduled_time_utc == 0)` in `esp_schedule_set_next_scheduled_time_utc`), so computing it here is what anchors the absolute target at **create** time. Remove it and the anchor moves to **enable** time. It also means `esp_schedule_get()` — which returns `next_scheduled_time_utc` for `RELATIVE` — would report `0` for a created-but-not-yet-enabled schedule.

Narrowed it to that one case instead of deleting it:

```c
if (schedule->trigger.type == ESP_SCHEDULE_TYPE_RELATIVE && esp_schedule_nvs_is_enabled()) {
    schedule->next_scheduled_time_diff = esp_schedule_get_next_schedule_time_diff(schedule);
}
```

That removes the duplicate computation for every other type (your point stands there — `esp_schedule_start_timer()` recomputes from scratch) while keeping the `RELATIVE` semantics unchanged.

esp_schedule/src/esp_schedule.c
Junius00: Confirmed and fixed — thanks, this was the real bug in the set. The chain was exactly as you describe: `esp_schedule_nvs_add()` runs inside `esp_schedule_set()`, which is called *before* `esp_schedule_create_timer()` computes the target, so the persisted `next_scheduled_time_utc` was always `0`. A stored `0` is indistinguishable from "never armed", so after a reboot a fired one-shot was recomputed and re-armed. Bare one-shot `DATE` was the worst case, as you note.

I fixed it slightly differently from your suggestion. Persisting from `esp_schedule_start_timer()` has two problems:

- **It writes on every re-arm.** `start_timer()` runs again after every fire, so a daily or weekly schedule would take one flash write per fire, forever — to store a value that repeating triggers never read back (`fired_and_done` ignores the stored value for them).
- **It runs on the FreeRTOS timer daemon task.** `esp_schedule_common_timer_cb()` → `start_timer()`, and `CONFIG_FREERTOS_TIMER_TASK_STACK_DEPTH` defaults to 2048 bytes; a flash write there also blocks every other software timer in the system.

So the write goes in `esp_schedule_enable()` instead — the API call that arms the schedule, on the caller's own thread, the same context as the existing `nvs_add()` in `esp_schedule_set()`:

```c
esp_schedule_start_timer(schedule);
if (schedule->trigger.next_scheduled_time_utc > 0 && esp_schedule_trigger_is_one_shot(&schedule->trigger)) {
    esp_schedule_nvs_add(schedule);
}
```

`esp_schedule_trigger_is_one_shot()` is the configuration-only half of `fired_and_done`, split out so it can be asked "does this fire once?" without "has it fired?". `fired_and_done` is now `(fired) && is_one_shot`, behaviour and signature unchanged.

Reboot path for a `DAY_ONCE` schedule:

1. `enable()` → target computed → stored as `Fri 19:00`.
2. Fires Friday. `start_timer()` re-arms, finds nothing, writes nothing.
3. Reboot. `esp_schedule_init()` reads the past target, `fired_and_done` is true, `has_future` is false → `esp_schedule_delete()` → `esp_schedule_nvs_remove()`.

Not re-armed. One NVS write per one-shot schedule, zero for repeating ones.

I deliberately did **not** add an erase at fire time. It would have to run on the timer daemon task, and `esp_schedule_init()` already removes expired entries from NVS, so a fired one-shot's dead entry just lingers until the next boot — same as before this PR, and nothing consults it in the meantime.

On tests: not added for this path. The existing suite is pure engine logic, whereas this needs NVS plus a faked clock across a simulated reboot, which is closer to an integration test. Happy to add one if you would like it in this PR rather than as a follow-up.

Two adjacent things I found while doing this, both pre-existing — say the word and I will fix either here or separately:

- `esp_schedule_edit()` only clears `next_scheduled_time_utc` for `RELATIVE`, so editing an already-fired one-shot to a new future date leaves the stale past target in place and `fired_and_done` refuses to arm it. An edit is a redefinition, so it should arguably clear the target for all types.
- `esp_schedule_nvs_get()` allocates the *stored* blob size and casts it to `esp_schedule_t *` with no `buf_size != sizeof(esp_schedule_t)` check. Any future change to the struct layout misparses old blobs, and a layout *growth* under-allocates, so `init`'s writes to `timer` / the callbacks would corrupt the heap.

esp_schedule/src/esp_schedule.c
Junius00: Your reading of the code is right — `day = 27, year = 2026, no months mask` does keep firing on the 27th of every remaining month of 2026, bounded at year end rather than at one occurrence.

That is intentional rather than an oversight: it preserves the v1.3.3 `is_expired` semantics, which anchored expiry to `date.year`. It is already documented and covered by a test, so I have left the behaviour alone:

- `docs/trigger_rules.md` §3 — "`year != 0` (day-only) → not marked done here; the engine bounds it and it expires after that year."
- §4 scenario table — `dow 0`, `day 5`, `months 0`, `year 2026` → "Fires on the 5th of each month **in 2026 only**, then dead." Unit test: `date year bounding`.
- The same section spells out the neighbouring case: with `year = 0` a months mask repeats every year, and setting `date.year` is what makes a seasonal pattern expire.

So the "explicit example in the docs" you asked for is there. If you would rather `year != 0` without a months mask mean "specific date, once" — i.e. `return trigger->date.repeat_months == 0;` — that is a one-line change, but it breaks parity with 1.3.3 for anyone relying on the year-bounded monthly behaviour, so I would rather make it deliberately than as part of this fix. Your call.

Junius00:
@shahpiyushv @vikramdattu should be ready for review now; trigger document is updated as discussed

vikramdattu:
LGTM.
@mahavirj can you please run automated review pass?

Junius00:
Fixed both suppressed comments

vikramdattu:
@mahavirj can you please merge this?

vikramdattu (approved):
LGTM

PR #788, opened 2026-07-27
esp_sysview: Fix flush path and events buffer overflow

erhankur:
- check the s_events_buf size for every link type, not only the debug probe one, so that a failed write cannot write past the end of it
- return esp_err_t from the flush functions and keep the events buffered when a write fails, instead of dropping them
- add the encoder flush entry so that esp_trace_flush() works, and call flush_nolock() again only when the transport returns quickly
- take no lock in the panic handler, and do not log from the RTT path
- move the SEGGER_RTT_ESP_* declarations out of the vendored SEGGER header into an Espressif one

### DIFF
--- a/esp_sysview/src/esp/SEGGER_RTT_esp.c
+++ b/esp_sysview/src/esp/SEGGER_RTT_esp.c
@@ -7,10 +7,10 @@
 #include "string.h"
 #include "freertos/FreeRTOS.h"
 #include "SEGGER_RTT.h"
+#include "SEGGER_RTT_esp.h"
 #include "SEGGER_SYSVIEW.h"
 #include "SEGGER_SYSVIEW_Conf.h"
 
-#include "esp_log.h"
 #include "esp_cpu.h"
 #include "esp_trace_port_transport.h"
 #include "esp_trace_port_encoder.h"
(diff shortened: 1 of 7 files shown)


esp_sysview/src/esp/SEGGER_RTT_esp.c
```diff
@@ -7,18 +7,18 @@
...
-#include "esp_log.h"
 #include "esp_cpu.h"
 #include "esp_trace_port_transport.h"
 #include "esp_trace_port_encoder.h"
 #include "esp_trace_types.h"
 #include "adapter_encoder_sysview.h"
 #include "esp_private/startup_internal.h"
 
-const static char *TAG = "segger_rtt";
+/* No logging in this file, these functions can run in ISR context */
```
gerekon: you can use ESP_EARLY_LOG instead, w/o locks. I suppose that any tracing API write/flush etc. can be called from ISR context directly by users or from RTOS, so need to care of it.

gerekon (approved):

esp_sysview/src/esp/SEGGER_RTT_esp.c
erhankur: You're right that ESP_EARLY_LOGx avoids the newlib lock.

The problem is that ESP_EARLY_LOGx writes to both ROM putc channels, so with a UART console the early logs still go out the USB-Serial-JTAG port. 

I fixed the routing on the IDF side with esp_rom_install_channel_putc(2, NULL) in the USJ transport init, but that isn't merged yet.

The two flush messages would also repeat on every retry now that flush_nolock() is called in a loop. Both functions return esp_err_t, so the caller still gets the failure.

erhankur: Comment is updated.

PR #789, opened 2026-07-30
refactor(esp schedule): v1.5.0 (IEC-568)

Junius00:
# Checklist

- [x] Component contains License
- [x] Component contains README.md
- [x] Component contains idf_component.yml file with `url` field defined
- [x] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [x] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [x] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description

PR #790, opened 2026-07-30
refactor(esp_schedule): add porting glue layer and split public header (IEC-569)

Junius00:
# Checklist

- [ ] CI passing

# Change description

Add glue layer to `esp_schedule` component

### DIFF
--- a/esp_schedule/CMakeLists.txt
+++ b/esp_schedule/CMakeLists.txt
@@ -1,8 +1,7 @@
 include(${CMAKE_CURRENT_LIST_DIR}/esp_schedule_variables.cmake)
 
 # Include ESP-IDF specific glue sources and include directories
-list(APPEND ESP_SCHEDULE_SRCS "src/esp_schedule_nvs.c"
-                              "glue/esp/nvs.c"
+list(APPEND ESP_SCHEDULE_SRCS "glue/esp/nvs.c"
                               "glue/esp/time.c"
                               "glue/esp/timer.c")
 list(APPEND ESP_SCHEDULE_INCLUDE_DIRS "include/esp")
--- a/esp_schedule/CMakeLists.txt
+++ b/esp_schedule/CMakeLists.txt
@@ -1,7 +1,15 @@
-set(component_srcs "src/esp_schedule.c"
-                   "src/esp_schedule_nvs.c")
+include(${CMAKE_CURRENT_LIST_DIR}/esp_schedule_variables.cmake)
 
-idf_component_register(SRCS "${component_srcs}"
-                       INCLUDE_DIRS "include"
-                       PRIV_INCLUDE_DIRS "src"
-                       PRIV_REQUIRES nvs_flash esp_netif)
+# Include ESP-IDF specific glue sources and include directories
+list(APPEND ESP_SCHEDULE_SRCS "src/esp_schedule_nvs.c"
+                              "glue/esp/nvs.c"
+                              "glue/esp/time.c"
+                              "glue/esp/timer.c")
+list(APPEND ESP_SCHEDULE_INCLUDE_DIRS "include/esp")
+list(APPEND ESP_SCHEDULE_PRIV_INCLUDE_DIRS "glue/esp")
+list(APPEND ESP_SCHEDULE_PRIV_REQUIRES "nvs_flash" "esp_netif")
+
+idf_component_register(SRCS ${ESP_SCHEDULE_SRCS}
+                       INCLUDE_DIRS ${ESP_SCHEDULE_INCLUDE_DIRS}
+                       PRIV_INCLUDE_DIRS ${ESP_SCHEDULE_PRIV_INCLUDE_DIRS}
+                       PRIV_REQUIRES ${ESP_SCHEDULE_PRIV_REQUIRES})
--- a/esp_schedule/esp_schedule_variables.cmake
+++ b/esp_schedule/esp_schedule_variables.cmake
@@ -1,0 +1,15 @@
+# This file is used to set the common non-glue variables for the esp_schedule component.
+# Include this file and extend the variables to add glue sources and include directories.
+
+# Source files
+set(ESP_SCHEDULE_SRCS "${CMAKE_CURRENT_LIST_DIR}/src/esp_schedule.c")
+
+# Include directories
+set(ESP_SCHEDULE_INCLUDE_DIRS "${CMAKE_CURRENT_LIST_DIR}/include/common")
+
+# Private include directories
+set(ESP_SCHEDULE_PRIV_INCLUDE_DIRS "${CMAKE_CURRENT_LIST_DIR}/src"
+                                   "${CMAKE_CURRENT_LIST_DIR}/glue")
+
+# Private requirements
+set(ESP_SCHEDULE_PRIV_REQUIRES "esp_daylight")
--- a/esp_schedule/esp_schedule_variables.cmake
+++ b/esp_schedule/esp_schedule_variables.cmake
@@ -1,8 +1,11 @@
 # This file is used to set the common non-glue variables for the esp_schedule component.
 # Include this file and extend the variables to add glue sources and include directories.
 
-# Source files
-set(ESP_SCHEDULE_SRCS "${CMAKE_CURRENT_LIST_DIR}/src/esp_schedule.c")
+# Source files. src/esp_schedule_nvs.c is platform-agnostic: it talks to storage
+# only through glue_nvs.h, so it belongs here rather than with the glue sources.
+# Leaving it out would silently build a port with no schedule persistence.
+set(ESP_SCHEDULE_SRCS "${CMAKE_CURRENT_LIST_DIR}/src/esp_schedule.c"
+                      "${CMAKE_CURRENT_LIST_DIR}/src/esp_schedule_nvs.c")
 
 # Include directories
 set(ESP_SCHEDULE_INCLUDE_DIRS "${CMAKE_CURRENT_LIST_DIR}/include/common")
--- a/esp_schedule/glue/esp/nvs.c
+++ b/esp_schedule/glue/esp/nvs.c
@@ -1,0 +1,121 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file esp_nvs.c
+ * @brief ESP NVS implementation.
+ */
+
+#include "nvs_flash.h"
+#include "glue_nvs.h"
+
+#include <string.h>
+
+static nvs_open_mode_t to_nvs_open_mode(esp_schedule_nvs_open_mode_t mode)
+{
+    switch (mode) {
+    case ESP_SCHEDULE_NVS_OPEN_READONLY:
+        return NVS_READONLY;
+    case ESP_SCHEDULE_NVS_OPEN_READWRITE:
+        return NVS_READWRITE;
+    default:
+        return NVS_READWRITE;
+    }
+}
+
+static esp_schedule_nvs_error_t to_esp_schedule_nvs_error(esp_err_t err)
+{
+    switch (err) {
+    case ESP_OK:
+        return ESP_SCHEDULE_NVS_OK;
+    case ESP_ERR_NVS_NOT_FOUND:
+        return ESP_SCHEDULE_NVS_NOT_FOUND;
+    case ESP_ERR_NO_MEM:
+    case ESP_ERR_NVS_NOT_ENOUGH_SPACE:
+        /* Preserve the out-of-memory / out-of-space signal so callers can tell
+         * resource exhaustion apart from a generic failure and take the correct
+         * recovery path (e.g. free cached data and retry the save). */
+        return ESP_SCHEDULE_NVS_NO_MEM;
+    default:
+        return ESP_SCHEDULE_NVS_ERROR;
+    }
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_open_from_partition(const char *partition_label, const char *name_space, esp_schedule_nvs_open_mode_t mode, esp_schedule_nvs_handle_t *p_handle)
+{
+    return to_esp_schedule_nvs_error(nvs_open_from_partition(partition_label, name_space, to_nvs_open_mode(mode), (nvs_handle_t *) p_handle));
+}
+
+void esp_schedule_nvs_close(esp_schedule_nvs_handle_t handle)
+{
+    nvs_close((nvs_handle_t) handle);
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_commit(esp_schedule_nvs_handle_t handle)
+{
+    return to_esp_schedule_nvs_error(nvs_commit((nvs_handle_t) handle));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_erase_key(esp_schedule_nvs_handle_t handle, const char *key)
+{
+    return to_esp_schedule_nvs_error(nvs_erase_key((nvs_handle_t) handle, key));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_erase_all(esp_schedule_nvs_handle_t handle)
+{
+    return to_esp_schedule_nvs_error(nvs_erase_all((nvs_handle_t) handle));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_set_blob(esp_schedule_nvs_handle_t handle, const char *key, const void *value, size_t value_len)
+{
+    return to_esp_schedule_nvs_error(nvs_set_blob((nvs_handle_t) handle, key, value, value_len));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_get_blob(esp_schedule_nvs_handle_t handle, const char *key, void *value, size_t *p_value_len)
+{
+    return to_esp_schedule_nvs_error(nvs_get_blob((nvs_handle_t) handle, key, value, p_value_len));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_set_u8(esp_schedule_nvs_handle_t handle, const char *key, uint8_t value)
+{
+    return to_esp_schedule_nvs_error(nvs_set_u8((nvs_handle_t) handle, key, value));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_get_u8(esp_schedule_nvs_handle_t handle, const char *key, uint8_t *value)
+{
+    return to_esp_schedule_nvs_error(nvs_get_u8((nvs_handle_t) handle, key, value));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_entry_find_blobs(const char *partition_label, const char *name_space, esp_schedule_nvs_iterator_t *iterator)
+{
+    return to_esp_schedule_nvs_error(nvs_entry_find(partition_label, name_space, NVS_TYPE_BLOB, (nvs_iterator_t *) iterator));
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_entry_get_key(esp_schedule_nvs_iterator_t iterator, char **p_key)
+{
+    nvs_entry_info_t nvs_entry;
+    esp_err_t err = nvs_entry_info((nvs_iterator_t) iterator, &nvs_entry);
+    if (err != ESP_OK) {
+        return to_esp_schedule_nvs_error(err);
+    }
+    size_t key_len = strlen(nvs_entry.key);
+    *p_key = (char *)malloc(key_len + 1);
+    if (*p_key == NULL) {
+        return ESP_SCHEDULE_NVS_NO_MEM;
+    }
+    memcpy(*p_key, nvs_entry.key, key_len + 1);
+    return ESP_SCHEDULE_NVS_OK;
+}
+
+esp_schedule_nvs_error_t esp_schedule_nvs_entry_next(esp_schedule_nvs_iterator_t *iterator)
+{
+    return to_esp_schedule_nvs_error(nvs_entry_next((nvs_iterator_t *) iterator));
+}
+
+void esp_schedule_nvs_release_iterator(esp_schedule_nvs_iterator_t iterator)
+{
+    nvs_release_iterator((nvs_iterator_t) iterator);
+}
--- a/esp_schedule/glue/esp/nvs.c
+++ b/esp_schedule/glue/esp/nvs.c
@@ -5,12 +5,13 @@
  */
 
 /**
- * @file esp_nvs.c
+ * @file nvs.c
  * @brief ESP NVS implementation.
  */
 
 #include "nvs_flash.h"
 #include "glue_nvs.h"
+#include "glue_mem.h"
 
 #include <string.h>
 
@@ -102,7 +103,9 @@ esp_schedule_nvs_error_t esp_schedule_nvs_entry_get_key(esp_schedule_nvs_iterato
         return to_esp_schedule_nvs_error(err);
     }
     size_t key_len = strlen(nvs_entry.key);
-    *p_key = (char *)malloc(key_len + 1);
+    /* Freed by the caller with ESP_SCHEDULE_FREE, so it must come from the
+     * same allocator. */
+    *p_key = (char *)ESP_SCHEDULE_MALLOC(key_len + 1);
     if (*p_key == NULL) {
         return ESP_SCHEDULE_NVS_NO_MEM;
     }
--- a/esp_schedule/glue/esp/timer.c
+++ b/esp_schedule/glue/esp/timer.c
@@ -5,12 +5,13 @@
  */
 
 /**
- * @file esp_timer.c
+ * @file timer.c
  * @brief ESP Timer implementation.
  */
 
 #include "glue_timer.h"
 #include "glue_log.h"
+#include "glue_mem.h"
 
 #include <stddef.h>
 #include <stdint.h>
@@ -40,7 +41,7 @@ typedef struct {
     esp_schedule_timer_cb_t cb;
     void *priv_data;
     uint64_t remaining_ticks; /* ticks still to wait after the current period */
-} __timer_priv_data_t;
+} esp_schedule_timer_priv_data_t;
 
 /* Callback serialization ******************************************************
  *
@@ -48,7 +49,7 @@ typedef struct {
  * while esp_schedule_timer_cancel() runs in the caller's (app) task. xTimerStop
  * only prevents FUTURE expiries; it does not wait for a callback that is already
  * executing. Without synchronization, cancel() would free() the priv_data (and
- * the caller then frees the schedule) while __timer_common_cb is still
+ * the caller then frees the schedule) while esp_schedule_timer_common_cb is still
  * dereferencing them in the daemon task -> use-after-free.
  *
  * A single global mutex serializes the callback body against cancel(). The
@@ -70,7 +71,7 @@ static portMUX_TYPE s_cb_mutex_init_lock = portMUX_INITIALIZER_UNLOCKED;
 static StaticSemaphore_t s_cb_mutex_buf;
 static SemaphoreHandle_t s_cb_mutex = NULL;
 
-static void __ensure_cb_mutex(void)
+static void esp_schedule_ensure_cb_mutex(void)
 {
     if (s_cb_mutex != NULL) {
         return;
@@ -87,7 +88,7 @@ static void __ensure_cb_mutex(void)
 /**
  * @brief Whether the caller is running in the FreeRTOS timer daemon task.
  */
-static bool __in_timer_daemon(void)
+static bool esp_schedule_in_timer_daemon(void)
 {
     return xTaskGetCurrentTaskHandle() == xTimerGetTimerDaemonTaskHandle();
 }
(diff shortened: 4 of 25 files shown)


shahpiyushv (commented):
Reviewed the glue seam + header split on its own — splitting this out of #789 was the right call, the mechanical part is much easier to reason about in isolation.

The seams themselves look good: `src/` no longer references `esp_log.h`, `esp_sntp.h`, `freertos/*` or `esp_heap_caps.h`, and no ESP types leak through the `glue_*.h` interfaces. Two things worth stating explicitly in the PR description because they're the first questions a reviewer asks:

- **The NVS blob layout is unchanged.** `TimerHandle_t` → `esp_schedule_timer_handle_t` is pointer-to-pointer, so `sizeof(esp_schedule_t)` and every field offset hold and existing flash contents still load.
- **Lazy timer creation is strictly better than master**, which called `xTimerCreate()` at `create()` time and ignored failure, so a failed creation meant `xTimerStop(NULL)` and a FreeRTOS assert on the first enable.

Also nice: `esp_schedule_timer_cancel()` nulls the caller's handle (master left `schedule->timer` dangling before `free(schedule)`), dropping the dead `esp_schedule_timer_reset()` that #789 still carries, and the new deadlock canary test — its comment correctly explains why asserting on `next_scheduled_time_utc` would *not* catch the bug.

Most findings are in `glue/esp/timer.c`, the one file that's new logic rather than a rename. Four I'd consider blocking, all commented inline:

1. **`esp_schedule_timer_cancel()` self-deadlocks when reached from a trigger callback.** `esp_schedule_timer_start()` guards for the daemon-task case with `__in_timer_daemon()`; `cancel()` takes the non-recursive mutex unconditionally. `esp_schedule_delete()` (or `esp_schedule_edit()`) called from inside `trigger_cb` freezes the timer daemon, and with it every software timer in the system.
2. **`portMAX_DELAY` on timer-command-queue calls that are reachable from the daemon task** (`timer_stop()`, and `cancel()`'s stop/delete). timer.c states this rule for the re-arm path but doesn't apply it here.
3. **Two dropped `xTimerChangePeriod()` return values**, one of which silently loses a slice of a long delay.
4. **`__`-prefixed identifiers** are reserved for the implementation by C11 §7.1.3 — worth avoiding specifically in the file whose purpose is toolchain portability.

Plus two layering points inline: `src/esp_schedule_nvs.c` sitting in the ESP-specific CMake list, and `esp_schedule_untyped.h` dropping `#include "sdkconfig.h"` while still gating public struct members on `CONFIG_ESP_SCHEDULE_ENABLE_DAYLIGHT`.

### Smaller things, not worth inline comments

- `esp_schedule_prepare_timer()` (renamed from `create_timer`, `src/esp_schedule.c`) no longer touches a timer at all — it only computes the RELATIVE diff. Worth renaming to match.
- `esp_schedule_enable()` still returns `ESP_OK` when the arm actually failed. `esp_schedule_arm_timer()` now *detects* the failure (logs it, zeroes the timestamp) but nothing propagates. #789 fixes this by threading return codes; if this lands first, it may be worth doing here.
- `to_esp_schedule_return_type()` in `src/esp_schedule_nvs.c` collapses every NVS error into four buckets and maps `NOT_FOUND` onto the same `INVALID_STATE` used for "NVS not enabled". No live impact since callers ignore these returns, but the specific `ESP_ERR_NVS_*` codes master propagated are gone.
- `glue_log.h` has the same `#include`-inside-`extern "C"` issue as `glue_mem.h`.
- `@file` names are wrong in all three glue implementations: `timer.c` says `@file esp_timer.c`, `nvs.c` says `esp_nvs.c`, `time.c` says `esp_time.c`.
- `examples/get_started/README.md` now points "API Reference" at `esp_schedule_untyped.h`; `include/esp/esp_schedule.h` is the actual entry point.
- Version 1.4.0 → 1.4.1 for a public-header reorganization: `#include "esp_schedule.h"` still resolves through the new `include/esp` public dir so it isn't source-breaking, and a patch bump is defensible — but a minor bump would signal the layout change more clearly.

### Pre-existing, out of scope here — just don't lose #789's fixes

`esp_schedule_nvs_get()` `malloc(buf_size)` then casting to `esp_schedule_t *` with no size/version validation; no `handle_count < *schedule_count` bound in `esp_schedule_nvs_get_all()`; count underflow in `esp_schedule_nvs_remove()`; `esp_schedule_common_timer_cb` using `schedule` after `trigger_cb` returns (the use-after-free behind finding 1); `init_done` set and never read.

esp_schedule/glue/esp/timer.c
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /* Detach the private data under the callback mutex. Taking the mutex is the
+     * barrier: any callback already executing runs to completion before we
+     * proceed, and any callback dispatched afterwards will read a NULL timer ID
+     * and return without touching the freed data. The mutex is released before
+     * the blocking xTimerDelete()/free() so a full timer-command queue cannot
+     * deadlock the daemon against this task. */
+    __timer_priv_data_t *priv_data = __timer_priv_data_get(timer_handle);
+    xSemaphoreTake(s_cb_mutex, portMAX_DELAY);
```
shahpiyushv: **Blocking: this self-deadlocks when `cancel()` is reached from a trigger callback.**

The mutex is non-recursive and `__timer_common_cb` holds it across the entire callback body, including the user's `trigger_cb`. `esp_schedule_timer_start()` handles exactly this case (`__in_timer_daemon()` on line 215, skipping the take), but `cancel()` takes it unconditionally:

```
__timer_common_cb  [holds s_cb_mutex, timer daemon task]
  -> trigger_cb
       -> esp_schedule_delete(handle)
            -> esp_schedule_delete_timer        src/esp_schedule.c:746
                 -> esp_schedule_timer_cancel   timer.c:265
                      -> xSemaphoreTake(s_cb_mutex, portMAX_DELAY)   <-- already held
```

The timer daemon task blocks forever, freezing **every** software timer in the system, not just this schedule's. `esp_schedule_edit()` from a callback reaches the same place via `esp_schedule_set()`.

A one-shot schedule that deletes itself when it fires is a natural pattern, so I'd expect this to be hit.

Options: apply the same `__in_timer_daemon()` guard here; use a recursive mutex; or defer the destroy off the daemon task. Whichever way, the "repeating schedule re-arms without deadlock" canary test generalises nicely — the same pattern around `esp_schedule_delete()` called from inside `trigger_cb` would cover this.

esp_schedule/glue/esp/timer.c
```diff
@@ -0,0 +1,296 @@
...
+    *p_timer_handle = (esp_schedule_timer_handle_t) timer_handle;
+    return true;
+}
+
+void esp_schedule_timer_stop(esp_schedule_timer_handle_t timer_handle)
+{
+    if (timer_handle == NULL) {
+        return;
+    }
+    xTimerStop((TimerHandle_t) timer_handle, portMAX_DELAY);
```
shahpiyushv: **Blocking: `portMAX_DELAY` here is reachable from the timer daemon task.**

```
esp_schedule_common_timer_cb            [daemon task]
  -> esp_schedule_start_timer
       -> esp_schedule_stop_timer       src/esp_schedule.c:683 (time not updated)
                                        src/esp_schedule.c:694 (calculation failed)
            -> esp_schedule_timer_stop  <-- xTimerStop(..., portMAX_DELAY)
```

FreeRTOS forbids a non-zero block time on the timer command queue from inside a callback: with a full queue the daemon ends up waiting for itself to drain it. The comments on lines 170-171 and 235-237 state this rule for the re-arm path — it needs to apply here too.

Same applies to the `xTimerStop()` on line 275 and `xTimerDelete()` on line 291 in `cancel()`, which are reachable from the daemon via the path in my other comment.

esp_schedule/glue/esp/timer.c
```diff
@@ -0,0 +1,296 @@
...
+    if (timer_priv_data == NULL) {
+        xSemaphoreGive(s_cb_mutex);
+        return;
+    }
+    if (timer_priv_data->remaining_ticks > 0) {
+        /* A long delay split across multiple periods: keep counting down and
+         * re-arm without invoking the user callback yet. A timer callback must
+         * never block on the timer command queue, so use a zero block time. */
+        TickType_t period = __period_ticks(timer_priv_data->remaining_ticks);
+        timer_priv_data->remaining_ticks -= period;
```
shahpiyushv: **Blocking: `remaining_ticks` is decremented before the arm, and the arm's return value is dropped.**

`xTimerChangePeriod()` on the next line uses a zero block time (correctly — we're on the daemon task), so it *can* fail if the command queue is full. When it does, this subtraction has already happened: that slice of the delay is gone and the timer is never re-armed, so a long schedule silently never fires.

Suggest checking the result and only committing the subtraction on success, e.g.

```c
TickType_t period = __period_ticks(timer_priv_data->remaining_ticks);
if (xTimerChangePeriod(timer_handle, period, 0) == pdPASS) {
    timer_priv_data->remaining_ticks -= period;
} else {
    /* log; the schedule is now disarmed */
}
```

esp_schedule/glue/esp/timer.c
```diff
@@ -0,0 +1,296 @@
...
+        timer_priv_data->priv_data = priv_data;
+        timer_priv_data->remaining_ticks = remaining_ticks;
+        if (!on_daemon) {
+            xSemaphoreGive(s_cb_mutex);
+        }
+
+        /* xTimerChangePeriod also (re)starts a dormant/expired timer. On the
+         * daemon task the block time MUST be zero; from an app task it may
+         * block on the command queue. */
+        xTimerChangePeriod(timer_handle, period, on_daemon ? 0 : portMAX_DELAY);
```
shahpiyushv: `xTimerChangePeriod()`'s result is dropped and the function returns `true` unconditionally on the reuse path.

That means `esp_schedule_arm_timer()`'s failure branch (`src/esp_schedule.c:665-668`, which logs and zeroes `next_scheduled_time_utc`) can never fire for a re-arm — only for a first-time creation. A dropped command here leaves the schedule silently disarmed while the caller believes it is armed.

Worth propagating: `return xTimerChangePeriod(timer_handle, period, on_daemon ? 0 : portMAX_DELAY) == pdPASS;`

esp_schedule/glue/esp/timer.c
```diff
@@ -0,0 +1,296 @@
...
+#define ESP_SCHEDULE_TIMER_MAX_TICKS ((TickType_t)(portMAX_DELAY / 2))
+
+/**
+ * @brief Timer private data.
+ */
+typedef struct {
+    esp_schedule_timer_cb_t cb;
+    void *priv_data;
+    uint64_t remaining_ticks; /* ticks still to wait after the current period */
+} __timer_priv_data_t;
```
shahpiyushv: **Identifiers beginning with two underscores are reserved for the implementation** (C11 §7.1.3), so `__timer_priv_data_t`, `__timer_common_cb`, `__ensure_cb_mutex`, `__in_timer_daemon`, `__ticks_from_seconds`, `__period_ticks` and `__timer_priv_data_create`/`_get` are all technically UB and can collide with compiler or libc internals.

Usually harmless in practice, but this is specifically the file whose purpose is portability to other toolchains, so it seems worth not relying on. Suggest `s_` for statics and `esp_schedule_` otherwise, matching the rest of the component.

Unrelated nit while here: the header comment on line 8 says `@file esp_timer.c`, but the file is `timer.c`. Same in `glue/esp/nvs.c` (`esp_nvs.c`) and `glue/esp/time.c` (`esp_time.c`).

esp_schedule/CMakeLists.txt
```diff
@@ -1,7 +1,15 @@
-set(component_srcs "src/esp_schedule.c"
-                   "src/esp_schedule_nvs.c")
+include(${CMAKE_CURRENT_LIST_DIR}/esp_schedule_variables.cmake)
 
-idf_component_register(SRCS "${component_srcs}"
-                       INCLUDE_DIRS "include"
-                       PRIV_INCLUDE_DIRS "src"
-                       PRIV_REQUIRES nvs_flash esp_netif)
+# Include ESP-IDF specific glue sources and include directories
+list(APPEND ESP_SCHEDULE_SRCS "src/esp_schedule_nvs.c"
```
shahpiyushv: `src/esp_schedule_nvs.c` isn't ESP-IDF specific, so it shouldn't be under the "ESP-IDF specific glue sources" comment.

It includes only `esp_schedule_internal.h`, `glue_mem.h`, `glue_nvs.h` and `glue_log.h` — it's platform-agnostic by construction; `glue/esp/nvs.c` is the ESP-specific half. It belongs in `esp_schedule_variables.cmake` next to `src/esp_schedule.c`.

As written, a port that follows the documented recipe in the README (include `esp_schedule_variables.cmake`, append your own glue) silently builds without any NVS persistence at all — `esp_schedule_nvs_add()` and friends just won't be linked in.

esp_schedule/include/common/esp_schedule_untyped.h
```diff
@@ -4,13 +4,22 @@
...
+ *       return-value macros, then includes this file:
+ *       - ESP_SCHEDULE_RETURN_TYPE            : the API return type (e.g. esp_err_t)
+ *       - ESP_SCHEDULE_RET_OK / _FAIL / _NO_MEM / _INVALID_ARG / _INVALID_STATE
+ */
+
 #pragma once
 
 #include <stdint.h>
 #include <stdbool.h>
 #include <time.h>
```
shahpiyushv, line 22: `#include "sdkconfig.h"` was dropped here, but this header still gates public types on `#if CONFIG_ESP_SCHEDULE_ENABLE_DAYLIGHT` (lines 60 and 121) — including the `solar` member **inside `esp_schedule_trigger_t`**, and therefore `sizeof(esp_schedule_config_t)`.

It does work today, but only through an incidental chain:

```
include/esp/esp_schedule.h -> esp_err.h -> esp_compiler.h -> sdkconfig.h
```

Nothing documents that dependency, and if it ever breaks — or for any non-IDF typed binding that doesn't happen to define the macro, which is the whole point of this split — the result is a **silent** struct-layout divergence between application and library. No compile error, no link error; `esp_schedule_get()` writes past the caller's struct.

One-line fix that keeps the untyped header platform-agnostic: add `#include "sdkconfig.h"` to `include/esp/esp_schedule.h` (the ESP-specific binding) before it includes this file.

Separately, lines 60/121 use the bare `#if CONFIG_...` form while other conditionals elsewhere use `#if defined(X) && X`. Worth picking one — the bare form also warns under `-Wundef`.

esp_schedule/glue/esp/nvs.c
```diff
@@ -0,0 +1,121 @@
...
+
+esp_schedule_nvs_error_t esp_schedule_nvs_entry_get_key(esp_schedule_nvs_iterator_t iterator, char **p_key)
+{
+    nvs_entry_info_t nvs_entry;
+    esp_err_t err = nvs_entry_info((nvs_iterator_t) iterator, &nvs_entry);
+    if (err != ESP_OK) {
+        return to_esp_schedule_nvs_error(err);
+    }
+    size_t key_len = strlen(nvs_entry.key);
+    *p_key = (char *)malloc(key_len + 1);
```
shahpiyushv: This allocates with raw `malloc()`, but the caller frees it with `ESP_SCHEDULE_FREE` (`src/esp_schedule_nvs.c:258`) — an allocator mismatch straight across the glue boundary.

Harmless today because `ESP_SCHEDULE_FREE` is `free`, but the point of `glue_mem.h` is that the allocator is swappable, and a custom `ESP_SCHEDULE_MALLOC` would make this a cross-allocator free. Same pattern in `glue/esp/timer.c` (raw `calloc`/`free` on lines 129, 249, 293) and `src/esp_schedule_nvs.c:288` (`strndup`).

Suggest routing these through `glue_mem.h` too. Minor related nit: `strndup(p, strlen(p))` is just `strdup(p)`, and the partition string it allocates is never freed (there's no deinit).

esp_schedule/glue/glue_mem.h
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * The default "glue_mem_impl.h" is provided for ESP-IDF under glue/esp.
+ * If you wish to use a different implementation, you can create your own "glue_mem_impl.h" file and not include the default one.
+ * - The following memory allocation macros MUST be defined:
+ *   - ESP_SCHEDULE_MALLOC(size) : malloc implementation.
+ *   - ESP_SCHEDULE_CALLOC(num, size) : calloc implementation.
+ *   - ESP_SCHEDULE_REALLOC(ptr, size) : realloc implementation.
+ *   - ESP_SCHEDULE_FREE(ptr) : free implementation.
+ */
+#include "glue_mem_impl.h"
```
shahpiyushv, line 28: This `#include` sits inside the `extern "C" { }` block, so `esp_heap_caps.h` gets pulled in with C linkage forced. It happens to work because that header has its own guards, but it's the classic way to break a C++ consumer — better above the `extern "C"`. Same in `glue_log.h:29` with `esp_log.h`.

Also, `glue/esp/glue_mem_impl.h` uses `malloc`/`calloc`/`realloc`/`free` without including `<stdlib.h>`; it only compiles because every current includer happens to pull it in first.

esp_schedule/esp_schedule_variables.cmake
```diff
@@ -0,0 +1,15 @@
...
+
+# Include directories
+set(ESP_SCHEDULE_INCLUDE_DIRS "${CMAKE_CURRENT_LIST_DIR}/include/common")
+
+# Private include directories
+set(ESP_SCHEDULE_PRIV_INCLUDE_DIRS "${CMAKE_CURRENT_LIST_DIR}/src"
+                                   "${CMAKE_CURRENT_LIST_DIR}/glue")
+
+# Private requirements
+set(ESP_SCHEDULE_PRIV_REQUIRES "esp_daylight")
```
shahpiyushv: Missing trailing newline at end of file.

esp_schedule/test_app/main/CMakeLists.txt
```diff
@@ -1,9 +1,7 @@
-# The tests exercise the component's internal helpers (esp_schedule_internal.h),
-# so the component's private src/ directory is added to the include path. The
-# esp_schedule dependency itself is pulled in via main/idf_component.yml.
+include(${CMAKE_CURRENT_LIST_DIR}/../../esp_schedule_variables.cmake)
 set(_srcs "test_app_main.c" "test_esp_schedule.c")
 idf_component_register(SRCS ${_srcs}
-                    PRIV_INCLUDE_DIRS "${CMAKE_CURRENT_LIST_DIR}/../../src"
-                    PRIV_REQUIRES "esp_daylight" "unity" "nvs_flash" "esp_netif"
+                    PRIV_INCLUDE_DIRS ${ESP_SCHEDULE_PRIV_INCLUDE_DIRS}
```
shahpiyushv, line 7: `ESP_SCHEDULE_PRIV_INCLUDE_DIRS` is `src` + `glue`, but not `glue/esp` (the component's own `CMakeLists.txt` appends that separately).

Fine today, since the tests only reach `glue_timer.h` transitively via `esp_schedule_internal.h`. But any future test that includes `glue_log.h` or `glue_mem.h` will fail to find the corresponding `*_impl.h`, with a confusing error. Might be worth appending `glue/esp` here now.

esp_schedule/glue/esp/timer.c
Junius00: Confirmed and fixed in 1010b91 — the mutex is now recursive (`xSemaphoreCreateRecursiveMutexStatic`), which covers both re-entry paths at once: `cancel()` from `trigger_cb` and the re-arm from `esp_schedule_timer_start()`. The `__in_timer_daemon()` skip in `start()` is gone as a result; the take is unconditional again and just resolves recursively on the owning task.

I went with the recursive mutex over a `__in_timer_daemon()` guard in `cancel()` because the guard only fixes the paths we happen to enumerate, whereas recursion is correct for any re-entry from the callback. The barrier semantics still hold: from an app task the take blocks until an in-flight callback completes; from the callback itself there is nothing to barrier against, since we *are* the thing being waited on.

Took the canary suggestion — new test `schedule deletes itself from its callback without deadlock`. It does earn its keep: with the fix reverted it fails, canary never fires and the NVS removal never completes.

One thing worth flagging, because it changed the scope of the fix. Once the deadlock is gone, the path immediately hits the use-after-free you listed under "pre-existing, out of scope": `esp_schedule_delete()` frees the schedule and then `esp_schedule_common_timer_cb` calls `esp_schedule_start_timer()` on it. So fixing this one alone trades a hang for heap corruption, and the new test would have caught that instead. I included the minimal fix — the free is deferred to the dispatching callback:

```c
if (schedule == s_dispatching) {
    s_dispatch_deleted = true;
    return ESP_SCHEDULE_RET_OK;   /* callback completes the free */
}
```

The `s_dispatching` check sits *after* `esp_schedule_delete_timer()`, which is what makes it safe without a lock: that call already barriers against a callback running on another task, so reaching the check with `s_dispatching` still equal to this schedule can only mean the caller is its own running callback.

esp_schedule/glue/esp/timer.c
Junius00: Fixed in 1010b91. All command-queue block times now come from one helper rather than being hardcoded:

```c
static TickType_t esp_schedule_cmd_block_ticks(void)
{
    return esp_schedule_in_timer_daemon() ? 0 : portMAX_DELAY;
}
```

Applied to `timer_stop()`, and to both the `xTimerStop()` and `xTimerDelete()` in `cancel()`.

Also caught one you did not list: `xTimerStart()` on the creation path in `esp_schedule_timer_start()` had the same `portMAX_DELAY`. It is daemon-reachable too — a `trigger_cb` can create and enable a brand new schedule, which lands on first-time creation rather than the reuse path. Same helper applied there.

esp_schedule/glue/esp/timer.c
Junius00: Fixed in 1010b91, taking your suggestion:

```c
TickType_t period = esp_schedule_period_ticks(timer_priv_data->remaining_ticks);
if (xTimerChangePeriod(timer_handle, period, 0) == pdPASS) {
    timer_priv_data->remaining_ticks -= period;
} else {
    ESP_SCHEDULE_LOGE(TAG, "Failed to re-arm timer; schedule is now disarmed");
}
```

This is what pulled `glue_log.h` into `timer.c`, which it did not include before.

esp_schedule/glue/esp/timer.c
Junius00: Fixed in 1010b91, as suggested:

```c
return xTimerChangePeriod(timer_handle, period, esp_schedule_cmd_block_ticks()) == pdPASS;
```

So `esp_schedule_arm_timer()`'s failure branch is now reachable for re-arms, not just first-time creation.

The creation path had a related hole: `xTimerStart()`'s result was dropped too. It now returns `false` on failure, and detaches the timer ID before deleting — if the delete command itself cannot be queued the timer outlives the call, and a later expiry has to find a NULL ID rather than the freed private data.

Note this only makes `enable()` *able* to see the failure. `esp_schedule_enable()` still returns `ESP_OK` regardless, per your point in the review body — left alone here, since #789 threads the return codes properly.

esp_schedule/glue/esp/timer.c
Junius00: Fixed in ee7be26. All seven renamed: `esp_schedule_timer_priv_data_t`, `esp_schedule_timer_common_cb`, `esp_schedule_ensure_cb_mutex`, `esp_schedule_in_timer_daemon`, `esp_schedule_ticks_from_seconds`, `esp_schedule_period_ticks`, `esp_schedule_timer_priv_data_create`/`_get`, plus the new `esp_schedule_cmd_block_ticks`.

Extended it to the glue include guards, which were the same class of violation — `__GLUE_MEM_H__`, `__GLUE_LOG_H__`, `__GLUE_TIMER_H__`, `__GLUE_TIME_H__`, `__GLUE_NVS_H__`. Switched to `#pragma once`, matching `esp_schedule_internal.h` and the public headers.

`@file` names fixed in all three (`timer.c`, `nvs.c`, `time.c`).

esp_schedule/CMakeLists.txt
Junius00: Agreed and fixed in dc4b64a — `src/esp_schedule_nvs.c` moved into `esp_schedule_variables.cmake` alongside `src/esp_schedule.c`, and dropped from the ESP-specific list in `CMakeLists.txt`.

The silent-no-persistence failure mode is the convincing part: it links, it runs, schedules just never survive a reboot. Added a comment at the new location so it does not drift back.

esp_schedule/include/common/esp_schedule_untyped.h
Junius00: Fixed in dc4b64a — `#include "sdkconfig.h"` added to `include/esp/esp_schedule.h`, above the `esp_err.h` include, with a comment explaining why it is there so nobody prunes it as unused.

On the conditional form: I checked before changing anything, and the component is already consistent — all 13 sites in `src/`, `include/` and `test_app/main/` use the bare `#if CONFIG_ESP_SCHEDULE_ENABLE_DAYLIGHT`. The only `defined(X) && X`-style test is `glue/esp/glue_mem_impl.h:17` on `CONFIG_SPIRAM`/`CONFIG_SPIRAM_SUPPORT`, which is a different case — those genuinely may be undefined, whereas the daylight macro now always has `sdkconfig.h` in scope. So there is nothing to unify, and I left the bare form rather than churn 13 sites. Happy to switch the lot if you would rather be `-Wundef`-clean regardless.

esp_schedule/glue/esp/nvs.c
Junius00: Fixed in ee7be26. `glue/esp/nvs.c:105` now uses `ESP_SCHEDULE_MALLOC`, and `glue/esp/timer.c` uses `ESP_SCHEDULE_CALLOC`/`ESP_SCHEDULE_FREE` at all three sites.

`src/esp_schedule_nvs.c:288` — replaced the `strndup(p, strlen(p))` with an explicit `ESP_SCHEDULE_MALLOC` + `memcpy`, which drops the redundant-`strlen` oddity at the same time.

The unfreed partition string I left as is: there is no deinit entry point to hang it off, `esp_schedule_nvs_init()` early-returns when already enabled so it cannot accumulate, and it is one small allocation for the process lifetime. Adding a deinit API felt out of scope for this PR — say the word if you would rather it went in here.

esp_schedule/glue/glue_mem.h
Junius00: Fixed in ee7be26 for both. Moving the include above the `extern "C"` left the block empty in each case — `glue_mem.h` and `glue_log.h` contribute macros only — so I dropped the block rather than leave a no-op wrapper, and noted in the comment that the impl header must not be wrapped either.

`<stdlib.h>` added to `glue/esp/glue_mem_impl.h`.

esp_schedule/esp_schedule_variables.cmake
Junius00: Fixed in dc4b64a.

esp_schedule/test_app/main/CMakeLists.txt
Junius00: Fixed in dc4b64a — `glue/esp` appended in `test_app/main/CMakeLists.txt` with a comment on why the variables file cannot cover it (it is the portable half; the `*_impl.h` headers are ESP-specific).

Kept it as a `list(APPEND ...)` at the test-app level rather than moving it into `esp_schedule_variables.cmake`, since that file must stay glue-free for the porting recipe to mean anything.

Junius00:
Thanks — all four blocking findings and both layering points are addressed, in three commits on top of the branch. Replies are on the individual threads; this covers the non-inline items and what I deliberately left.

**1010b91** — timer glue re-entrancy (findings 1-4)
**dc4b64a** — build seam (`esp_schedule_nvs.c` placement, `sdkconfig.h`, test include dirs, trailing newline)
**ee7be26** — naming and includes (reserved identifiers, allocator routing, `extern "C"`, `@file`, renames)

### One scope change worth calling out

Fixing finding 1 makes the use-after-free you filed under "pre-existing, out of scope" immediately reachable: with the daemon no longer hanging, `esp_schedule_delete()` frees the schedule and `esp_schedule_common_timer_cb` then re-arms it. So the deadlock fix on its own would have converted a hang into heap corruption. I included the minimal fix — the free is deferred to the dispatching callback — rather than ship the halfway state. Details on that thread.

### Smaller things from your list

- `esp_schedule_prepare_timer()` → `esp_schedule_prepare_relative_target()`. It computes the RELATIVE anchor and touches no timer.
- `examples/get_started/README.md` API Reference now points at `include/esp/esp_schedule.h`.
- `@file` names fixed in all three glue implementations.
- `glue_log.h` got the same `extern "C"` treatment as `glue_mem.h`.

### Left alone, on purpose

- **`esp_schedule_enable()` returning `ESP_OK` on a failed arm.** The failure is now *detectable* — both `xTimerChangePeriod()` and `xTimerStart()` propagate — but nothing threads it up. #789 does this properly with return codes and I would rather not build a second, conflicting version of it here. If this lands first I will port that part across.
- **`to_esp_schedule_return_type()` collapsing the `ESP_ERR_NVS_*` codes.** Agreed it loses information versus master. Same reasoning: it is a return-code change, and it belongs with the #789 work rather than split across both PRs.
- **Version bump.** Still 1.4.1. Happy to make it 1.5.0 to signal the layout change, but #789 is titled v1.5.0, so I would rather you pick which PR owns that number than have the two collide.
- **The pre-existing NVS items** (`nvs_get()` size/version validation, the missing `handle_count` bound, the `nvs_remove()` count underflow, `init_done` set-and-never-read). Untouched here, and noted so they do not get lost with #789.
- **Bare `#if CONFIG_...`.** Checked rather than assumed: the component is already consistent at all 13 sites. Detail on that thread.

### Verification

test_app and `examples/get_started` both build clean on IDF v6.0 / esp32. Full suite is 36/36 under QEMU after each of the three commits. The new self-delete canary fails as expected when the fix is reverted — the daemon hangs and the canary never fires.

Junius00:
Favouring runtime function pointers instead - see #791

PR #791, opened 2026-07-31
refactor(esp_schedule): replace compile-time glue with a runtime port (IEC-570)

### LINKED ISSUE #733: Make `esp_schedule` sntp initialization optional (IEC-529)
### Answers checklist.

- [x] I have read the documentation of the component in question and the issue is not addressed there.

### General issue report

esp_schedule currently initialize the sntp module if it is not initialized.

Users might use an RTC clock, or update the local time without having access to internet.

The esp_schedule is not usable for such cases. Would it be possible to disable this from the init function ?

Junius00:
# Checklist

- [x] CI passing

# Change description
Platform access - timer, storage, time, memory, log - now goes through function-pointer tables installed at init instead of substituted headers. esp_schedule_init() installs the ESP-IDF ones; esp_schedule_init_with_config() takes your own, and because esp_schedule_init() is the only referrer of the defaults, they never reach the link.

Replaces #790. Also closes #733

### DIFF
--- a/esp_schedule/port/esp/timer.c
+++ b/esp_schedule/port/esp/timer.c
@@ -1,0 +1,344 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @file timer.c
+ * @brief FreeRTOS software-timer implementation of esp_schedule_timer_ops_t.
+ *
+ * @note Reached only through esp_schedule_esp_timer_ops, which only
+ *       esp_schedule_init() references. A build that installs its own port
+ *       never links this file, and with it never pulls in FreeRTOS timers.
+ */
+
+#include "esp_schedule.h"
+#include "esp_schedule_esp_port.h"
+
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include "esp_log.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "freertos/timers.h"
+#include "freertos/semphr.h"
+
+static const char *TAG = "esp_schedule_timer";
+
+/* Types **********************************************************************/
+
+/* xTimerCreate()/xTimerChangePeriod() take a 32-bit TickType_t, so a period
+ * derived from a delay beyond ~49.7 days (at CONFIG_FREERTOS_HZ=1000) would
+ * truncate and fire early. Cap a single hardware period to a safe value and
+ * split longer delays across several periods, re-arming without invoking the
+ * user callback until the full delay has elapsed. portMAX_DELAY/2 leaves
+ * headroom and avoids the portMAX_DELAY "block forever" sentinel. */
+#define ESP_SCHEDULE_TIMER_MAX_TICKS ((TickType_t)(portMAX_DELAY / 2))
+
+/**
+ * @brief Timer private data.
+ */
+typedef struct {
+    esp_schedule_timer_cb_t cb;
+    void *priv_data;
+    uint64_t remaining_ticks; /* ticks still to wait after the current period */
+} esp_schedule_timer_priv_data_t;
+
+/* Callback serialization ******************************************************
+ *
+ * FreeRTOS software-timer callbacks run in the timer service (daemon) task,
+ * while esp_schedule_timer_cancel() runs in the caller's (app) task. xTimerStop
+ * only prevents FUTURE expiries; it does not wait for a callback that is already
+ * executing. Without synchronization, cancel() would free() the priv_data (and
+ * the caller then frees the schedule) while esp_schedule_timer_common_cb is still
+ * dereferencing them in the daemon task -> use-after-free.
+ *
+ * A single global mutex serializes the callback body against cancel(). The
+ * callback holds it while running (including across the user trigger callback);
+ * cancel() waits on it as a barrier so any in-flight callback has fully
+ * completed before priv_data is freed. cancel() never holds the mutex while
+ * calling a blocking timer API, so a full timer-command queue cannot deadlock
+ * the daemon against the app task.
+ *
+ * The mutex is RECURSIVE. The user trigger callback runs with it held, and from
+ * there the user may legitimately re-enter this layer on the same task: a
+ * repeating schedule re-arms itself via esp_schedule_timer_start(), a brand new
+ * schedule may be created and enabled, and a one-shot schedule that deletes
+ * itself reaches esp_schedule_timer_cancel(). With a non-recursive mutex either
+ * re-entry would block the daemon task on a mutex the daemon itself holds,
+ * freezing every software timer in the system. Recursive ownership makes the
+ * nested take a no-op for the owning task while still serializing app tasks
+ * against an in-flight callback.
+ *
+ * Those same re-entry paths make every entry point below reachable from the
+ * daemon task, which is why none of them may hardcode a block time on the timer
+ * command queue - see esp_schedule_cmd_block_ticks(). */
+static portMUX_TYPE s_cb_mutex_init_lock = portMUX_INITIALIZER_UNLOCKED;
+static StaticSemaphore_t s_cb_mutex_buf;
+static SemaphoreHandle_t s_cb_mutex = NULL;
+
+static void esp_schedule_ensure_cb_mutex(void)
+{
+    if (s_cb_mutex != NULL) {
+        return;
+    }
+    /* Static allocation performs no heap work, so it is safe inside a critical
+     * section; the lock makes first-time creation race-free. */
+    portENTER_CRITICAL(&s_cb_mutex_init_lock);
+    if (s_cb_mutex == NULL) {
+        s_cb_mutex = xSemaphoreCreateRecursiveMutexStatic(&s_cb_mutex_buf);
+    }
+    portEXIT_CRITICAL(&s_cb_mutex_init_lock);
+}
+
+/**
+ * @brief Whether the caller is running in the FreeRTOS timer daemon task.
+ */
+static bool esp_schedule_in_timer_daemon(void)
+{
+    return xTaskGetCurrentTaskHandle() == xTimerGetTimerDaemonTaskHandle();
+}
+
+/**
+ * @brief Block time to use for a timer command queue operation.
+ *
+ * FreeRTOS forbids a non-zero block time on the timer command queue from within
+ * a timer callback: if the queue is full the daemon task would end up waiting
+ * for itself to drain it. Hence the block time is chosen from the calling
+ * context rather than hardcoded.
+ */
+static TickType_t esp_schedule_cmd_block_ticks(void)
+{
+    return esp_schedule_in_timer_daemon() ? 0 : portMAX_DELAY;
+}
+
+/* Private functions **********************************************************/
+
+/**
+ * @brief Convert a delay in seconds to timer ticks (64-bit to avoid overflow).
+ */
+static uint64_t esp_schedule_ticks_from_seconds(uint32_t delay_seconds)
+{
+    return (((uint64_t) delay_seconds) * 1000) / portTICK_PERIOD_MS;
+}
+
+/**
+ * @brief Clamp a 64-bit tick count to a valid single timer period (>0).
+ */
+static TickType_t esp_schedule_period_ticks(uint64_t total_ticks)
+{
+    if (total_ticks == 0) {
+        return 1; /* a timer period must be strictly greater than 0 */
+    }
+    if (total_ticks > ESP_SCHEDULE_TIMER_MAX_TICKS) {
+        return ESP_SCHEDULE_TIMER_MAX_TICKS;
+    }
+    return (TickType_t) total_ticks;
+}
+
+/**
+ * @brief Create a new timer private data.
+ *
+ * @param[in] cb Callback function.
+ * @param[in] priv_data Private data.
+ *
+ * @return Pointer to the timer private data.
+ */
+static esp_schedule_timer_priv_data_t *esp_schedule_timer_priv_data_create(esp_schedule_timer_cb_t cb, void *priv_data)
+{
+    esp_schedule_timer_priv_data_t *data = calloc(1, sizeof(esp_schedule_timer_priv_data_t));
+    if (data == NULL) {
+        return NULL;
+    }
+    data->cb = cb;
+    data->priv_data = priv_data;
+    return data;
+}
+
+/**
+ * @brief Get the timer private data from the timer handle.
+ *
+ * @param[in] timer_handle Timer handle.
+ *
+ * @return Pointer to the timer private data.
+ */
+static esp_schedule_timer_priv_data_t *esp_schedule_timer_priv_data_get(esp_schedule_timer_handle_t timer_handle)
+{
+    return (esp_schedule_timer_priv_data_t *) pvTimerGetTimerID(timer_handle);
+}
+
+/**
+ * @brief Common timer callback. Runs in the FreeRTOS timer daemon task.
+ *
+ * @param[in] timer_handle Timer handle.
+ */
+static void esp_schedule_timer_common_cb(TimerHandle_t timer_handle)
+{
+    esp_schedule_ensure_cb_mutex();
+    /* Hold the mutex for the whole callback so cancel() can barrier against it.
+     * priv_data is read AFTER acquiring the mutex: cancel() clears the timer ID
+     * under the same mutex before freeing, so we either see a valid pointer
+     * (cancel has not started) or NULL (cancel already ran). */
+    xSemaphoreTakeRecursive(s_cb_mutex, portMAX_DELAY);
+    esp_schedule_timer_priv_data_t *timer_priv_data = esp_schedule_timer_priv_data_get(timer_handle);
+    if (timer_priv_data == NULL) {
+        xSemaphoreGiveRecursive(s_cb_mutex);
+        return;
+    }
+    if (timer_priv_data->remaining_ticks > 0) {
+        /* A long delay split across multiple periods: keep counting down and
+         * re-arm without invoking the user callback yet. A timer callback must
+         * never block on the timer command queue, so use a zero block time.
+         *
+         * Only commit the countdown once the re-arm has actually been queued: a
+         * full command queue would otherwise consume this slice of the delay
+         * with no timer left running to serve it, and the schedule would
+         * silently never fire. */
+        TickType_t period = esp_schedule_period_ticks(timer_priv_data->remaining_ticks);
+        if (xTimerChangePeriod(timer_handle, period, 0) == pdPASS) {
+            timer_priv_data->remaining_ticks -= period;
+        } else {
+            ESP_LOGE(TAG, "Failed to re-arm timer; schedule is now disarmed");
+        }
+        xSemaphoreGiveRecursive(s_cb_mutex);
+        return;
+    }
+    timer_priv_data->cb(timer_priv_data->priv_data);
+    xSemaphoreGiveRecursive(s_cb_mutex);
+}
+
+/* Public functions ************************************************************/
+
+static bool esp_timer_start(esp_schedule_timer_handle_t *p_timer_handle, uint32_t delay_seconds, esp_schedule_timer_cb_t cb, void *priv_data)
+{
+    esp_schedule_ensure_cb_mutex();
+    TimerHandle_t timer_handle = (TimerHandle_t) * p_timer_handle;
+
+    uint64_t total_ticks = esp_schedule_ticks_from_seconds(delay_seconds);
+    TickType_t period = esp_schedule_period_ticks(total_ticks);
+    uint64_t remaining_ticks = (total_ticks > period) ? (total_ticks - period) : 0;
+
+    if (timer_handle != NULL) {
+        /* Reuse the existing timer rather than cancel + recreate. A repeating
+         * schedule re-arms from inside its own fired callback (daemon task);
+         * routing that through esp_schedule_timer_cancel() would delete the
+         * timer that is currently executing and free the priv_data the callback
+         * is still using.
+         *
+         * Instead, rebind the private data in place: this honors the start()
+         * contract (the caller-provided cb and priv_data are set explicitly,
+         * not assumed unchanged) without freeing anything the running callback
+         * references, and avoids per-fire create/delete churn.
+         *
+         * The mutex serializes the rebind against a callback running
+         * concurrently on the daemon, which could otherwise read a half-updated
+         * (cb, priv_data) pair. Re-arming from inside the fired callback takes
+         * it recursively on the task that already owns it. */
+        xSemaphoreTakeRecursive(s_cb_mutex, portMAX_DELAY);
+        esp_schedule_timer_priv_data_t *timer_priv_data = esp_schedule_timer_priv_data_get(timer_handle);
+        if (timer_priv_data == NULL) {
+            /* A concurrent cancel() detached and is deleting this timer; there
+             * is nothing safe to reuse. */
+            xSemaphoreGiveRecursive(s_cb_mutex);
+            return false;
+        }
+        timer_priv_data->cb = cb;
+        timer_priv_data->priv_data = priv_data;
+        timer_priv_data->remaining_ticks = remaining_ticks;
+        xSemaphoreGiveRecursive(s_cb_mutex);
+
+        /* xTimerChangePeriod also (re)starts a dormant/expired timer. A dropped
+         * command leaves the schedule disarmed, so report it rather than
+         * claiming success. */
+        return xTimerChangePeriod(timer_handle, period, esp_schedule_cmd_block_ticks()) == pdPASS;
+    }
+
+    esp_schedule_timer_priv_data_t *timer_priv_data = esp_schedule_timer_priv_data_create(cb, priv_data);
+    if (timer_priv_data == NULL) {
+        return false;
+    }
+    timer_priv_data->remaining_ticks = remaining_ticks;
+    timer_handle = xTimerCreate("schedule", period, pdFALSE, timer_priv_data, esp_schedule_timer_common_cb);
+    if (timer_handle == NULL) {
+        free(timer_priv_data);
+        return false;
+    }
+    if (xTimerStart(timer_handle, esp_schedule_cmd_block_ticks()) != pdPASS) {
+        /* Detach before deleting: if the delete command itself cannot be queued
+         * the timer outlives this call, and a later expiry must find a NULL ID
+         * rather than the freed private data. */
+        vTimerSetTimerID(timer_handle, NULL);
+        xTimerDelete(timer_handle, esp_schedule_cmd_block_ticks());
+        free(timer_priv_data);
+        return false;
+    }
+    *p_timer_handle = (esp_schedule_timer_handle_t) timer_handle;
+    return true;
+}
+
+static void esp_timer_stop(esp_schedule_timer_handle_t timer_handle)
+{
+    if (timer_handle == NULL) {
+        return;
+    }
+    xTimerStop((TimerHandle_t) timer_handle, esp_schedule_cmd_block_ticks());
+}
+
+static void esp_timer_cancel(esp_schedule_timer_handle_t *p_timer_handle)
+{
+    if (p_timer_handle == NULL || *p_timer_handle == NULL) {
+        return;
+    }
+    TimerHandle_t timer_handle = (TimerHandle_t) * p_timer_handle;
+    esp_schedule_ensure_cb_mutex();
+
+    TickType_t block_ticks = esp_schedule_cmd_block_ticks();
+
+    /* Stop the timer so no new expiry is dispatched. */
+    if (xTimerIsTimerActive(timer_handle) == pdTRUE) {
+        xTimerStop(timer_handle, block_ticks);
+    }
+
+    /* Detach the private data under the callback mutex. Taking the mutex is the
+     * barrier: any callback already executing on another task runs to
+     * completion before we proceed, and any callback dispatched afterwards will
+     * read a NULL timer ID and return without touching the freed data. When the
+     * caller IS the running callback the mutex is taken recursively and there is
+     * nothing to barrier against. The mutex is released before the
+     * xTimerDelete()/free() so a full timer-command queue cannot deadlock the
+     * daemon against this task.
+     *
+     * The read must happen INSIDE the mutex, not before it: two tasks cancelling
+     * the same handle would otherwise both capture a non-NULL pointer and both go
+     * on to delete the timer and free the data. Clearing the ID under the lock
+     * makes the non-NULL read the single claim on the delete. */
+    xSemaphoreTakeRecursive(s_cb_mutex, portMAX_DELAY);
+    esp_schedule_timer_priv_data_t *priv_data = esp_schedule_timer_priv_data_get(timer_handle);
+    if (priv_data != NULL) {
+        vTimerSetTimerID(timer_handle, NULL);
+    }
+    xSemaphoreGiveRecursive(s_cb_mutex);
+
+    if (priv_data == NULL) {
+        /* Another cancel already claimed this timer and owns its deletion. */
+        *p_timer_handle = NULL;
+        return;
+    }
+
+    xTimerDelete(timer_handle, block_ticks);
+    free(priv_data);
+    *p_timer_handle = NULL;
+}
+
+/* Operations table ************************************************************
+ *
+ * The only exported symbol in this file. esp_schedule_init() is its sole
+ * referrer, so nothing here is linked into a build that installs its own port. */
+const esp_schedule_timer_ops_t esp_schedule_esp_timer_ops = {
+    .start = esp_timer_start,
+    .stop = esp_timer_stop,
+    .cancel = esp_timer_cancel,
+};
--- a/esp_schedule/src/esp_schedule.c
+++ b/esp_schedule/src/esp_schedule.c
@@ -8,8 +8,6 @@
 #include <stdbool.h>
 #include <time.h>
 #include <inttypes.h>
-#include "esp_log.h"
-#include "esp_sntp.h"
 #include "esp_daylight.h"
 #include "esp_schedule_internal.h"
 
@@ -63,10 +61,8 @@ static const char *TAG = "esp_schedule";
  */
 #define MAX_MASKED_MONTH_ATTEMPTS (1 + (MAX_LEAP_YEAR_GAP - 1) + 1)
 
-static bool init_done = false;
-
 // Forward declarations for static functions
-static void esp_schedule_common_timer_cb(TimerHandle_t timer);
+static void esp_schedule_common_timer_cb(void *priv_data);
 
 /*
  * Unified date-based next occurrence calculation.
@@ -309,10 +305,10 @@ time_t esp_schedule_calc_solar_time_for_time_utc(bool is_sunrise, time_t time_ut
 
     bool calc_ok = esp_daylight_calc_sunrise_sunset_utc(year, month, day, latitude, longitude, &sunrise_utc, &sunset_utc);
     if (!calc_ok) {
-        ESP_LOGW(TAG, "Failed to calculate %s for date %04d-%02d-%02d at latitude %.5f, longitude %.5f (likely polar night/day condition)",
-                 is_sunrise ? "sunrise" : "sunset",
-                 year, month, day, latitude, longitude
-                );
+        ESP_SCHEDULE_LOGW(TAG, "Failed to calculate %s for date %04d-%02d-%02d at latitude %.5f, longitude %.5f (likely polar night/day condition)",
+                          is_sunrise ? "sunrise" : "sunset",
+                          year, month, day, latitude, longitude
+                         );
         return 0;
     }
     time_t solar_time = is_sunrise ? sunrise_utc : sunset_utc;
@@ -344,7 +340,7 @@ time_t esp_schedule_get_next_valid_solar_time(time_t now, const esp_schedule_tri
             // Outside validity window or not in the future -> advance to next valid day
         } else if (validity && validity->end_time && solar_time > validity->end_time) {
             // Past validity window -> return 0
-            ESP_LOGD(TAG, "Schedule %s: next solar event is past validity end_time.", schedule_name);
+            ESP_SCHEDULE_LOGD(TAG, "Schedule %s: next solar event is past validity end_time.", schedule_name);
             return 0;
         } else {
             return solar_time;
@@ -352,11 +348,11 @@ time_t esp_schedule_get_next_valid_solar_time(time_t now, const esp_schedule_tri
 
         // Advance anchor to next day
         if (!esp_schedule_get_next_date_time(day_end + 1, MINUTES_IN_DAY - 1, trigger->day.repeat_days, trigger->date.day, trigger->date.repeat_months, match_year, validity, &day_end)) {
-            ESP_LOGD(TAG, "Schedule %s: no further day matches the date/day-of-week arm.", schedule_name);
+            ESP_SCHEDULE_LOGD(TAG, "Schedule %s: no further day matches the date/day-of-week arm.", schedule_name);
             return 0;
         }
     }
-    ESP_LOGD(TAG, "Schedule %s: no solar event found within 370 candidate days.", schedule_name);
+    ESP_SCHEDULE_LOGD(TAG, "Schedule %s: no solar event found within 370 candidate days.", schedule_name);
     return 0;
 }
 #endif /* CONFIG_ESP_SCHEDULE_ENABLE_DAYLIGHT */
@@ -380,24 +376,24 @@ static bool esp_schedule_date_arm_is_valid(const esp_schedule_trigger_t *trigger
     /* V3: a months mask with no day-of-month would mean every day of those
      * months, which is never the intended schedule. */
     if (trigger->date.day == 0 && trigger->date.repeat_months != 0) {
-        ESP_LOGE(TAG, "Schedule %s: date.repeat_months is set but date.day is 0. Set date.day.", schedule_name);
+        ESP_SCHEDULE_LOGE(TAG, "Schedule %s: date.repeat_months is set but date.day is 0. Set date.day.", schedule_name);
         return false;
     }
     /* V4: a recurrence or a year bound with no date pattern to apply it to. */
     if (trigger->date.day == 0 && trigger->date.repeat_months == 0 &&
             (trigger->date.year != 0 || trigger->date.repeat_every_year)) {
-        ESP_LOGE(TAG, "Schedule %s: date.year/date.repeat_every_year set with no date.day or date.repeat_months to apply it to.", schedule_name);
+        ESP_SCHEDULE_LOGE(TAG, "Schedule %s: date.year/date.repeat_every_year set with no date.day or date.repeat_months to apply it to.", schedule_name);
         return false;
     }
     /* V5: "only year N" and "every year" are contradictory. */
     if (trigger->date.year != 0 && trigger->date.repeat_every_year) {
-        ESP_LOGE(TAG, "Schedule %s: date.year and date.repeat_every_year are mutually exclusive.", schedule_name);
+        ESP_SCHEDULE_LOGE(TAG, "Schedule %s: date.year and date.repeat_every_year are mutually exclusive.", schedule_name);
         return false;
     }
     /* V6: repeat_every_year recurs *over the month set*, so with no mask it
      * would be inert. date.year is exempt: it still constrains the arm. */
     if (trigger->date.repeat_every_year && trigger->date.repeat_months == 0) {
-        ESP_LOGE(TAG, "Schedule %s: date.repeat_every_year needs date.repeat_months to recur over (use ESP_SCHEDULE_MONTH_ALL for every month).", schedule_name);
+        ESP_SCHEDULE_LOGE(TAG, "Schedule %s: date.repeat_every_year needs date.repeat_months to recur over (use ESP_SCHEDULE_MONTH_ALL for every month).", schedule_name);
         return false;
     }
     return true;
@@ -426,21 +422,21 @@ bool esp_schedule_trigger_is_valid(const esp_schedule_trigger_t *trigger, const 
          * or before the base time, so the arm path would find nothing to fire and
          * the schedule would silently never run. */
         if (trigger->relative_seconds <= 0) {
-            ESP_LOGE(TAG, "Schedule %s: relative_seconds must be > 0, got %d.", schedule_name, trigger->relative_seconds);
+            ESP_SCHEDULE_LOGE(TAG, "Schedule %s: relative_seconds must be > 0, got %d.", schedule_name, trigger->relative_seconds);
             return false;
         }
         return true;
     case ESP_SCHEDULE_TYPE_DAYS_OF_WEEK:
         /* V1: this type reads only day.repeat_days. */
         if (esp_schedule_date_arm_present(trigger)) {
-            ESP_LOGE(TAG, "Schedule %s: DAYS_OF_WEEK reads only day.repeat_days, but a date.* field is set.", schedule_name);
+            ESP_SCHEDULE_LOGE(TAG, "Schedule %s: DAYS_OF_WEEK reads only day.repeat_days, but a date.* field is set.", schedule_name);
             return false;
         }
         return true;
     case ESP_SCHEDULE_TYPE_DATE:
         /* V2: DATE has no day-of-week arm. */
         if (trigger->day.repeat_days != 0) {
-            ESP_LOGE(TAG, "Schedule %s: DATE reads only date.*, but day.repeat_days is set. Use DAYS_OF_WEEK instead.", schedule_name);
+            ESP_SCHEDULE_LOGE(TAG, "Schedule %s: DATE reads only date.*, but day.repeat_days is set. Use DAYS_OF_WEEK instead.", schedule_name);
             return false;
         }
         return esp_schedule_date_arm_is_valid(trigger, schedule_name);
@@ -452,7 +448,7 @@ bool esp_schedule_trigger_is_valid(const esp_schedule_trigger_t *trigger, const 
          * populated is ambiguous, not a union. */
         if (trigger->day.repeat_days != 0) {
             if (esp_schedule_date_arm_present(trigger)) {
-                ESP_LOGE(TAG, "Schedule %s: solar day.repeat_days and date.* are mutually exclusive arms; only one may be set.", schedule_name);
+                ESP_SCHEDULE_LOGE(TAG, "Schedule %s: solar day.repeat_days and date.* are mutually exclusive arms; only one may be set.", schedule_name);
                 return false;
             }
             return true; /* day-of-week arm */
@@ -541,7 +537,7 @@ static bool esp_schedule_set_next_scheduled_time_utc(const char *schedule_name, 
     time_t now;
 
     /* Get current time */
-    time(&now);
+    ESP_SCHEDULE_GET_TIME(&now);
     /* Always recompute the next occurrence for repeating date/day-of-week/solar
      * triggers instead of reusing a stored next_scheduled_time_utc. This keeps
      * the fire time correct after a timezone change (picked up on the next arm)
@@ -582,7 +578,9 @@ static bool esp_schedule_set_next_scheduled_time_utc(const char *schedule_name, 
     if (trigger->type == ESP_SCHEDULE_TYPE_SUNRISE || trigger->type == ESP_SCHEDULE_TYPE_SUNSET) {
         time_t solar_time = esp_schedule_get_next_valid_solar_time(now, trigger, validity, schedule_name);
         if (solar_time == 0) {
-            ESP_LOGW(TAG, "Solar schedule %s has no next occurrence (no sunrise/sunset at this location/date, no remaining matching day, or past the validity window). Enable debug logs for the cause.", schedule_name);
+            /* Kept within ESP_SCHEDULE_LOG_BUF_LEN even with a full-length name;
+             * see the buffer comment in esp_schedule_port.c. */
+            ESP_SCHEDULE_LOGW(TAG, "Solar schedule %s has no next occurrence: no solar event at this location/date, no matching day left, or past validity. Enable debug logs.", schedule_name);
             return false;
         }
 
@@ -617,7 +615,7 @@ static bool esp_schedule_set_next_scheduled_time_utc(const char *schedule_name, 
 static uint32_t esp_schedule_get_next_schedule_time_diff(esp_schedule_t *schedule)
 {
     time_t now;
-    time(&now);
+    ESP_SCHEDULE_GET_TIME(&now);
 
     if (!esp_schedule_set_next_scheduled_time_utc(schedule->name, &schedule->trigger, &schedule->validity)) {
         schedule->trigger.next_scheduled_time_utc = 0;
@@ -633,7 +631,7 @@ static uint32_t esp_schedule_get_next_schedule_time_diff(esp_schedule_t *schedul
     localtime_r(&schedule->trigger.next_scheduled_time_utc, &schedule_time);
     memset(time_str, 0, sizeof(time_str));
     strftime(time_str, sizeof(time_str), "%c %z[%Z]", &schedule_time);
-    ESP_LOGI(TAG, "Schedule %s will be active on: %s. DST: %s", schedule->name, time_str, schedule_time.tm_isdst ? "Yes" : "No");
+    ESP_SCHEDULE_LOGI(TAG, "Schedule %s will be active on: %s. DST: %s", schedule->name, time_str, schedule_time.tm_isdst ? "Yes" : "No");
 
     /* Clamp before the uint32_t cast: casting a double outside uint32_t range is
      * undefined behavior. */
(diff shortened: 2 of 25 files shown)


esp_schedule/src/esp_schedule_port.c
```diff
@@ -0,0 +1,117 @@
...
+
+void esp_schedule_log(esp_schedule_log_level_t level, const char *tag, const char *format, ...)
+{
+    if (g_esp_schedule_port.log == NULL) {
+        return;
+    }
+    va_list args;
+    va_start(args, format);
+    g_esp_schedule_port.log(level, tag, format, args);
+    va_end(args);
```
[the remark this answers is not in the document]
Junius00: Fixed in e08dcc6, though not the way the alert reads at first glance.

The `va_end()` itself is correct: C11 7.16.1 says a callee that receives a `va_list` may consume it, but the caller that did `va_start()` still owns the matching `va_end()`. Both in-tree implementations (`vsnprintf` in the ESP port, `vprintf` in the custom-port test app) consume without ending, so there was no live defect. clang-tidy cannot see through `g_esp_schedule_port.log`, so it conservatively assumes the callee may have ended the list and reports our `va_end()` as operating on a released one.

It is a fair thing to be nervous about, though, and pointed at a real design flaw rather than a false positive to suppress: `esp_schedule_log_fn_t` is implemented by integrators, and nothing in the contract stopped a port from calling `va_end()` on the list it was handed. That would double-end the caller's list, which is genuinely undefined. The interface was one careless port away from real UB, and a `NOLINT` would have preserved that.

So instead of suppressing it, the `va_list` no longer crosses the boundary. `esp_schedule_log()` formats into a buffer and hands the port a plain NUL-terminated string:

```c
typedef void (*esp_schedule_log_fn_t)(esp_schedule_log_level_t level, const char *tag,
                                      const char *message);
```

The failure mode is now structurally impossible rather than merely unreported, and it pays for itself elsewhere: a port no longer needs a varargs printf, so `port/esp/log.c` loses its buffer and `vsnprintf` entirely (they moved up one level, so there is no extra copy) and is now just a switch over five `ESP_LOGx` macros. Ports to platforms with no `vsnprintf` get easier too.

Verified: test_app 36/36 under QEMU, the custom-port app still runs end to end, and the linker still strips the whole default port from it.

esp_schedule/examples/get_started/README.md
```diff
@@ -210,5 +210,5 @@ If you run out of memory when creating many schedules:
 
 ## Further Reading
 
-- [ESP Schedule API Reference](../../include/esp_schedule.h)
+- [ESP Schedule API Reference](../../include/esp/esp_schedule.h)
```
vikramdattu: Confirmed broken: `esp_schedule/include/` holds only `esp_schedule.h` and `esp_schedule_port.h` at this head — there is no `include/esp/` directory, so this link 404s. The original path was right.

Looks like the `esp/` from the new `port/esp/` layout got applied to the header path too.

esp_schedule/include/esp_schedule_port.h
```diff
@@ -0,0 +1,218 @@
...
+typedef struct esp_schedule_port_config {
+    esp_schedule_timer_ops_t timer;  /**< Required. */
+    esp_schedule_nvs_ops_t nvs;      /**< Optional; all-NULL disables persistence. */
+    esp_schedule_time_ops_t time;    /**< Required. */
+    esp_schedule_mem_ops_t mem;      /**< Required. */
+    esp_schedule_log_fn_t log;       /**< Optional; NULL discards all output. */
+} esp_schedule_port_config_t;
+
+/** Initialize ESP Schedule against a caller-supplied port.
+ *
```
vikramdattu:
```suggestion
/** 
 * @brief Initialize ESP Schedule against a caller-supplied port.
 *
```

esp_schedule/port/esp/esp_schedule_esp_port.h
```diff
@@ -0,0 +1,35 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern const esp_schedule_timer_ops_t esp_schedule_esp_timer_ops;
+extern const esp_schedule_nvs_ops_t esp_schedule_esp_nvs_ops;
+extern const esp_schedule_time_ops_t esp_schedule_esp_time_ops;
+extern const esp_schedule_mem_ops_t esp_schedule_esp_mem_ops;
+
+void esp_schedule_esp_log(esp_schedule_log_level_t level, const char *tag,
```
vikramdattu: API doc not needed?

esp_schedule/src/esp_schedule_port.c
```diff
@@ -0,0 +1,132 @@
...
+    /* Format here rather than handing the va_list to the port. Ownership of a
+     * va_list is easy to get wrong across an interface boundary - an
+     * implementation that called va_end() on it would double-end this one,
+     * which is undefined - and doing it here also spares every port a varargs
+     * printf. vsnprintf always NUL-terminates within the buffer. */
+    char message[ESP_SCHEDULE_LOG_BUF_LEN];
+    va_list args;
+    va_start(args, format);
+    vsnprintf(message, sizeof(message), format, args);
+    va_end(args);
```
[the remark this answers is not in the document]
Junius00: Correcting my previous reply: the restructure was worth doing on its own merit, but it was never going to clear this alert. Suppressed in 5c979ee instead, after actually reproducing it.

I pulled `warnings.txt` from the clang-tidy run for e08dcc6 and the reported path is the giveaway:

```
esp_schedule_port.c:129:5: warning: va_end() is called on an uninitialized va_list
:117:9: note: Assuming 'log' is not equal to NULL
:117:5: note: Taking false branch
:129:5: note: va_end() is called on an uninitialized va_list
```

It jumps from line 117 to line 129 and never models the `va_start()` at 127. It is not reasoning about the control flow in this function at all.

The reason I got it wrong the first time is that I could not reproduce locally: CI runs esp-clang **21.1.3**, the toolchain pinned by IDF v6.0 is **19.1.2**, and 19 does not report this. Once I built a clang compile database and ran the same checker locally, the file came back clean — which is what misled me into thinking the redesign had fixed it. Rebuilding the CI artifact is what settled it.

So this is a checker regression in clang 21, not a defect. The `va_start`/`va_end` pair is unconditional and adjacent, with only `vsnprintf()` in between, which consumes the list without ending it. Rather than contort the code around a false positive I have suppressed the single line with the reasoning inline:

```c
/* ... clang 19 does not report it. The pairing is unconditional and
 * immediate, with only vsnprintf() - which consumes but does not end the
 * list - in between. */
// NOLINTNEXTLINE(clang-analyzer-valist.Uninitialized)
va_end(args);
```

Worth flagging two other things from the same artifact, neither of which becomes an alert because `.github/filter_sarif.py` strips them by design (newlib has no `_s` variants):

- `esp_schedule_port.c:128` — `vsnprintf` insecureAPI, newly introduced by this PR
- `esp_schedule.c:630` `memset` and `esp_schedule_nvs.c:260` `memcpy` — same category

The va_list one was the only unfiltered warning this component produces.

esp_schedule/port/esp/timer.c
```diff
@@ -0,0 +1,335 @@
...
+
+    /* Detach the private data under the callback mutex. Taking the mutex is the
+     * barrier: any callback already executing on another task runs to
+     * completion before we proceed, and any callback dispatched afterwards will
+     * read a NULL timer ID and return without touching the freed data. When the
+     * caller IS the running callback the mutex is taken recursively and there is
+     * nothing to barrier against. The mutex is released before the
+     * xTimerDelete()/free() so a full timer-command queue cannot deadlock the
+     * daemon against this task. */
+    esp_schedule_timer_priv_data_t *priv_data = esp_schedule_timer_priv_data_get(timer_handle);
```
vikramdattu: `priv_data` is read on the line above the `xSemaphoreTakeRecursive()`, so two concurrent (or duplicate) cancels both capture a non-NULL pointer and both go on to `vTimerSetTimerID()` / `xTimerDelete()` / `free()` — double delete, double free.

Reading it *inside* the mutex and bailing out on NULL closes it for free, and matches the rule the callback comment already states ("priv_data is read AFTER acquiring the mutex").

esp_schedule/include/esp_schedule_port.h
```diff
@@ -0,0 +1,218 @@
...
+ * implementations out of the binary.
+ *
+ * @param[in] port Operations to install. Copied; need not outlive the call.
+ *                 Persistence is unavailable if @c port->nvs is all-NULL,
+ *                 whatever @c enable_nvs says.
+ * @param[in] enable_nvs Whether to enable persistence.
+ * @param[in] nvs_partition (Optional) Partition to use, NULL for the default.
+ * @param[out] schedule_count Number of schedules restored from storage.
+ *
+ * @return Array of restored schedule handles, or NULL if there are none. The
```
vikramdattu: Worth saying which `free`: this array now comes from `port->mem.malloc` (it was libc `malloc` before), so a port with its own allocator must release it with the matching `mem.free`. Only safe with plain `free()` because the default port happens to wrap malloc.

esp_schedule/src/esp_schedule_port.c
```diff
@@ -0,0 +1,140 @@
...
+
+    g_esp_schedule_port = *port;
+    s_port_installed = true;
+    return ESP_OK;
+}
+
+/* Long enough for the longest line this component emits; anything beyond that
+ * is truncated rather than grown, because this frequently runs on the timer
+ * task's stack. */
+#define ESP_SCHEDULE_LOG_BUF_LEN 160
```
vikramdattu, line 114: Note this makes logging format twice — here into 160 bytes, then `ESP_LOGx(tag, "%s", ...)` formats again — and both frames land on the timer daemon stack, where the trigger callback also runs (`FREERTOS_TIMER_TASK_STACK_DEPTH`, 2048 by default).

Not a bug, but it's the one measurable cost of the indirection and nothing in the README or Kconfig mentions the daemon stack.

esp_schedule/port/esp/timer.c
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * @brief Block time to use for a timer command queue operation.
+ *
+ * FreeRTOS forbids a non-zero block time on the timer command queue from within
+ * a timer callback: if the queue is full the daemon task would end up waiting
+ * for itself to drain it. Every entry point here is reachable from the daemon
+ * task via the user trigger callback, so the block time must be chosen from the
+ * calling context rather than hardcoded.
+ */
```
vikramdattu, line 112: "Reachable from the daemon task via the trigger callback" is now stated here, at the `xTimerStart()` call, and again in `cancel()` — on top of the file-header block. One of the three is enough; the header already carries the design.

vikramdattu (commented):
Nice shape overall — the port table plus "the defaults are only referenced by `esp_schedule_init()`, so they never reach the link" is a clean way to get a POSIX build without `#ifdef`s, and `check_port_strip.py` proving it rather than asserting it is the right instinct.

Two things worth fixing:

- `port/esp/timer.c` — `esp_timer_cancel()` reads `priv_data` just outside the mutex, which lets a duplicate/concurrent cancel double-free. The callback comment already states the correct rule.
- `esp_schedule_port.h` — the returned handle array now comes from `port->mem.malloc` rather than libc `malloc`; the doc should say it must be released with the matching `mem.free`.

Plus a broken README link and two comment nits.

On the recursive mutex: agreed, and I don't think it can be anything else. Both re-entry paths are real — a repeating schedule re-arms from inside its own fired callback, and a one-shot that deletes itself lands in `cancel()` — and both run on the daemon task while the callback holds the lock, so a plain mutex would have the daemon block on itself and stall every software timer in the system. Choosing the command-queue block time from the calling context, and never holding the lock across a blocking timer call, are the two details that make it safe; both are right here. Worth keeping the file-header block that explains it, so nobody "simplifies" it later.

vikramdattu:
@Junius00 looks good overall. The PR is neat. Added a few comments for you to ponder upon. Did you test this with real RainMaker device?

@shahpiyushv PTAL from your end and help merge the same.

esp_schedule/port/esp/timer.c
Junius00: Confirmed and fixed in `2799761`. The diagnosis is exact: both cancels capture a non-NULL `priv_data`, serialize on the mutex, and only the first one's `vTimerSetTimerID(NULL)` has any effect — the second still holds its stale copy and goes on to `xTimerDelete()` + `free()`.

The read now happens inside the mutex, and a NULL result is the signal that someone else owns the deletion:

```c
    xSemaphoreTakeRecursive(s_cb_mutex, portMAX_DELAY);
    esp_schedule_timer_priv_data_t *priv_data = esp_schedule_timer_priv_data_get(timer_handle);
    if (priv_data != NULL) {
        vTimerSetTimerID(timer_handle, NULL);
    }
    xSemaphoreGiveRecursive(s_cb_mutex);

    if (priv_data == NULL) {
        /* Another cancel already claimed this timer and owns its deletion. */
        *p_timer_handle = NULL;
        return;
    }
```

Clearing the ID under the lock makes the non-NULL read the single claim on the delete — the same rule `esp_schedule_timer_common_cb()` and the reuse path in `esp_timer_start()` already follow. This was the one entry point that didn't.

On reachability, for the record: the trigger is two app tasks calling `esp_schedule_delete()` on the same handle. Nothing in `esp_schedule.c` serializes that either (`esp_schedule_delete_timer()` is unguarded), so it is caller misuse rather than an internal race. But the port layer pays for exactly this class of defense everywhere else, and the inconsistency here is what made the window easy to miss.

esp_schedule/examples/get_started/README.md
Junius00: Fixed in `7039e27`. Exactly as you read it — the `esp/` from `port/esp/` had been applied to the header path. Corrected to `../../include/esp_schedule.h`, and I added a link to `esp_schedule_port.h` next to it since the porting interface is now public API too.

esp_schedule/include/esp_schedule_port.h
Junius00: Applied in `7039e27`, and swept the rest of the file for the same thing — it was mixing bare `/** Text` with `@brief` blocks.

esp_schedule/port/esp/esp_schedule_esp_port.h
Junius00: Needed, yes. Added in `7039e27`, for the four ops tables as well as the function:

```c
/** @brief Timer operations backed by FreeRTOS software timers. See port/esp/timer.c. */
extern const esp_schedule_timer_ops_t esp_schedule_esp_timer_ops;
...
/**
 * @brief Emit one pre-formatted log line through esp_log.
 *
 * Maps @c level onto the matching ESP_LOGx macro. A plain function rather than
 * an ops table because esp_schedule_log_fn_t is a bare function pointer.
 */
```

The tables were bare too, which mattered more than the function: which subsystem each one drags into the link is the whole reason this header carries a warning at the top.

esp_schedule/include/esp_schedule_port.h
Junius00: Right, and worth being precise about. Confirmed the allocation is `ESP_SCHEDULE_MALLOC` (`esp_schedule_nvs.c:207`), i.e. `port->mem.malloc`. Documented in `7039e27`:

```c
 * @note The caller owns the returned array and must release it with the same
 *       allocator it supplied - @c port->mem.free, not necessarily libc
 *       free(). The array is allocated with @c port->mem.malloc. Only the array
 *       itself is the caller's; the handles in it stay owned by the component.
```

The last sentence is the other half of the ambiguity — "the caller owns the array and must free it" didn't say whether the handles come with it.

The README snippet now shows `my_free(restored)` rather than `free()` for the same reason. This also turned up a live instance of the mistake: the custom-port test app was calling plain `free()` on that array, which only worked because its port wrapped the libc allocator. That app has since been removed for unrelated reasons (see the PR comment), but it was a good demonstration that the contract needed spelling out.

esp_schedule/port/esp/timer.c
Junius00: Trimmed in `2799761` — removed from both the `xTimerStart()` path and `esp_timer_cancel()`, keeping it in the header block only. I folded the two concrete cases the local comments carried into that block so nothing is lost:

```c
 * repeating schedule re-arms itself via esp_schedule_timer_start(), a brand new
 * schedule may be created and enabled, and a one-shot schedule that deletes
 * itself reaches esp_schedule_timer_cancel(). Every entry point below is
 * therefore reachable from the daemon task, which is also why none of them may
 * hardcode a block time on the timer command queue - see
 * esp_schedule_cmd_block_ticks().
```

`esp_schedule_cmd_block_ticks()` keeps the FreeRTOS rule it exists to satisfy, minus the restatement of reachability.

esp_schedule/src/esp_schedule_port.c
Junius00: The double-format cost is real and now documented, though one part of the premise needs correcting: the README does already carry the daemon stack, under "Timer Task Stack Requirement" — it asks for `CONFIG_FREERTOS_TIMER_TASK_STACK_DEPTH=3072` and explains that `localtime_r`/`strftime` on the re-arm path land on the daemon stack. What was missing is that logging adds two more frames on top of that, which is your actual point. Added in `7039e27`:

> Logging costs one more frame than it looks: `esp_schedule_log()` formats into a 160-byte stack buffer before handing the finished string to `port->log`, and the default port's `ESP_LOGx` then formats a second time. Both frames land on the daemon stack alongside the work above. Lowering the log ceiling (`CONFIG_LOG_MAXIMUM_LEVEL`, or `CONFIG_ESP_SCHEDULE_LOG_LEVEL` if you define one) compiles those calls out and reclaims the stack with them.

Kconfig now mentions it too, in the `ESP_SCHEDULE_DISABLE_DEFAULT_PORT` help — that is where the FreeRTOS-timer dependency is described, so it is the honest place for the consequence:

```
            Note that the default port runs schedule callbacks on the shared
            FreeRTOS timer daemon task, which needs
            CONFIG_FREERTOS_TIMER_TASK_STACK_DEPTH raised to at least 3072; see
            "Timer Task Stack Requirement" in the component README. A port of
            your own that dispatches on its own task avoids that.
```

On the cost itself: it is not quite an extra copy relative to `master`. Before this PR the same `vsnprintf` ran inside `port/esp/log.c` with its own buffer; the refactor moved that frame up one level rather than adding it. What is genuinely new is that the pre-formatted string then goes through `ESP_LOGx(tag, "%s", message)`, so the second pass is a `%s` copy of an already-finished string instead of a full format. That is the price of not passing a `va_list` across the port boundary, and I would rather pay it than hand integrators a list whose ownership rules are easy to get wrong.

160 bytes is deliberate for the same stack reason — the buffer truncates rather than grows.

Junius00:
Review comments addressed across three commits. Two notes on things that came out of the review but were not asked for.

**The README's "Glue Layers" section was stale.** Fixing the broken API link led me into it: the whole section still described the compile-time glue design this PR deletes — `glue_log.h`/`glue_mem.h`/`glue_timer.h`, sources under `glue/esp/`, and a recipe telling integrators to write a custom `CMakeLists.txt` and append glue sources via `esp_schedule_variables.cmake`. None of that exists any more, and the recipe is actively misleading now that a custom port is just a struct passed to `esp_schedule_init_with_config()`. Replaced with a "Port Layer" section covering the actual interface, which group is required, and the link-stripping guarantee. Reviewing `7039e27` on its own is probably easier than reading the diff.

**`port_test_app` is removed.** It was never wired into CI, which I did not notice when adding it. The artifact upload globs in `build_and_run_apps.yml` are `*/examples/*/build_esp*` and `*/test_app*/**/build_esp*`, and `esp_schedule/port_test_app` matches neither — `test_app*` needs the directory name to *start* with `test_app`. It also had no `pytest_*.py` and no `.build-test-rules.yml` entry. So `recursive = true` built it and nothing ever ran it, which means `check_port_strip.py` — the regression test for the link-strip guarantee, per its own docstring — asserted nothing on any PR.

Rather than leave a test that does not test, I dropped it. If the link-strip guarantee is worth enforcing in CI (I think it is — it breaks silently, with no compiler warning, if anything outside `esp_schedule_default.c` ever names an `esp_schedule_esp_*_ops` table), it should come back as `test_app_port` with a pytest file and a rules entry, in its own PR. Happy to do that if you agree.

Verification: `test_app` builds clean for esp32 with no warnings.

Junius00:
> Did you test this with real RainMaker device?

Yes, works as-is, by just replacing the version number

esp_schedule/src/esp_schedule.c
```diff
@@ -935,97 +959,127 @@ esp_err_t esp_schedule_edit(esp_schedule_handle_t handle, esp_schedule_config_t
...
     }
 
     /* Wait for time to be updated here */
 
     /* Below this is initialising schedules from NVS */
-    esp_schedule_nvs_init(nvs_partition);
+    if (esp_schedule_nvs_init(nvs_partition) != ESP_OK) {
+        /* Either the port has no storage or the partition name could not be
+         * copied. Schedules still work, they just will not persist. */
+        return NULL;
```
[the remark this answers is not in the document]
Junius00: Correct, and the same gap exists on the `!enable_nvs` return two blocks up (see the other comment). Note that `schedule_count` is already known non-NULL here — it is checked at the top of the NVS branch — so this really is just a missed assignment, not a NULL-guard question.

Fixed in `bd6dfb6` by hoisting it rather than patching each return. The port-install failure path was already zeroing it, so there were three returns and one of them got it right:

```c
    /* Clear the out-parameter up front so every early return below reports zero
     * restored schedules rather than leaving the caller's variable untouched.
     * Doing it here rather than per-return is what keeps them all consistent. */
    if (schedule_count != NULL) {
        *schedule_count = 0;
    }
```

The `*schedule_count = 0;` that used to sit just above `esp_schedule_nvs_get_all()` is now redundant and gone.

esp_schedule/src/esp_schedule.c
```diff
@@ -935,97 +959,127 @@ esp_err_t esp_schedule_edit(esp_schedule_handle_t handle, esp_schedule_config_t
...
+    }
+
+    if (g_esp_schedule_port.time.timesync_init != NULL) {
+        g_esp_schedule_port.time.timesync_init();
     }
 
     if (!enable_nvs) {
         return NULL;
     }
```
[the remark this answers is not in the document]
Junius00: Agreed — same root cause as the `esp_schedule_nvs_init()` return below, fixed together in `bd6dfb6` by clearing the out-parameter once at the top of the function instead of per-return. Three returns could reach the caller and only the port-install one zeroed it.

esp_schedule/src/esp_schedule.c
```diff
@@ -648,39 +644,30 @@ static uint32_t esp_schedule_get_next_schedule_time_diff(esp_schedule_t *schedul
...
-    if (ticks > max_ticks) {
-        ticks = max_ticks;
-    }
-    xTimerStop(schedule->timer, portMAX_DELAY);
-    xTimerChangePeriod(schedule->timer, (TickType_t)ticks, portMAX_DELAY);
+    if (!g_esp_schedule_port.timer.start(&schedule->timer, seconds, esp_schedule_common_timer_cb, (void *)schedule)) {
+        ESP_SCHEDULE_LOGE(TAG, "Failed to arm timer for schedule %s", schedule->name);
+        schedule->trigger.next_scheduled_time_utc = 0;
+    }
 }
```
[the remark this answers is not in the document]
Junius00: Confirmed, and it is worse than "may still fire" — it fires the user's trigger.

Tracing the ESP port: the reuse path in `esp_timer_start()` rebinds `(cb, priv_data, remaining_ticks)` and then returns `xTimerChangePeriod(...) == pdPASS`. If that command cannot be queued, the timer keeps its previous period *and its previous armed state*. So the old expiry survives. Meanwhile we have just set `next_scheduled_time_utc = 0`, and the early-expiry guard in `esp_schedule_common_timer_cb()` compares `now` against that field — with 0 it reads as "the instant has passed", so the guard waves it through and `trigger_cb` runs at the stale time.

Fixed in `bd6dfb6`:

```c
        /* A failed re-arm of an existing timer does not disarm it: start() may
         * have been unable to apply the new period while the old one is still
         * running, and it would then fire on a time we have just discarded.
         * Stop it explicitly, the same way the two bail-outs in
         * esp_schedule_start_timer() do. Best-effort: whatever made start()
         * fail may well make this fail too, but leaving the stale period armed
         * is strictly worse. */
        if (schedule->timer != NULL) {
            esp_schedule_stop_timer(schedule);
        }
```

The `timer != NULL` guard keeps the first-arm case correct: there, a `start()` failure leaves the handle NULL and there is nothing to stop.

Being honest about the limit: this is best-effort. A full timer command queue is the most likely reason `start()` failed, and `stop()` goes through the same queue, so it can fail too. It closes the common case (`xTimerChangePeriod` rejected while the daemon is busy) rather than making the operation atomic, and it now matches how the other two disarm paths in `esp_schedule_start_timer()` already behave.

esp_schedule/src/esp_schedule.c
```diff
@@ -714,12 +701,25 @@ static void esp_schedule_start_timer(esp_schedule_t *schedule)
...
+ * below -> use-after-free. A one-shot schedule that deletes itself when it fires
+ * is a natural pattern, so the free is deferred instead: delete() tears down the
+ * timer and the NVS entry as usual but leaves the allocation to the callback,
+ * which returns without re-arming.
+ *
+ * The two flags below need no lock. The timer glue serializes every callback body
+ * under one mutex, so at most one dispatch is in flight at a time, and delete()
+ * only consults them AFTER esp_schedule_delete_timer() has barriered against any
+ * callback running on another task. Reaching that point with s_dispatching still
+ * equal to this schedule therefore means the caller IS its running callback. */
```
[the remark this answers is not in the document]
Junius00: This one is right, and it is the most interesting comment in the batch — the code states the assumption in its own comment but the header never required it.

The concrete failure, for the record. Schedules A and B, a port that dispatches on two tasks:

1. task 1 enters A's callback → `s_dispatching = A`
2. task 2 enters B's callback → `s_dispatching = B`
3. A's `trigger_cb` calls `esp_schedule_delete(A)`. `esp_schedule_delete_timer()` barriers, but task 1 *is* A's callback so it returns immediately. The deferral check then reads `s_dispatching == B`, not `A`, so the free is **not** deferred and A is freed.
4. task 1 returns into `esp_schedule_common_timer_cb`, which reads its flags and re-arms — on freed memory.

Addressed in `bd6dfb6` by making the requirement explicit on `esp_schedule_timer_ops_t` rather than leaving it implied:

```c
 * @note Callback bodies must be serialized against each other: at most one
 *       esp_schedule_timer_cb_t may be executing at any time, across all
 *       timers. esp_schedule relies on this to detect a schedule deleting
 *       itself from its own trigger callback, where the deletion has to be
 *       deferred until the callback returns. ...
 *
 *       Serializing under a single mutex held for the whole callback satisfies
 *       both this and the @c cancel barrier below. It must be recursive, or
 *       taken in a way that tolerates re-entry, because the callback may call
 *       back into @c start or @c cancel on the same task.
```

That is a contract change, not a code fix, so worth being explicit about why.

The stronger fix is to move the two flags into `esp_schedule_t`, which removes the requirement entirely: the only concurrent access to schedule X's flag would then be X's own dispatch and a `delete(X)`, and `cancel()` is already required to barrier between those. I did not do it because **`esp_schedule_t` is persisted to NVS verbatim** and `esp_schedule_nvs_read_one()` rejects any blob whose size differs from `sizeof(esp_schedule_t)` — deliberately, since the blob is cast straight to the struct. Two extra `bool`s grow the struct past that check, so every schedule stored by an earlier build would be logged and dropped on upgrade. I checked that the current struct is layout-compatible with `master` (`TimerHandle_t` → `esp_schedule_timer_handle_t`, both pointers), and I would rather not spend that compatibility here.

The requirement is also not onerous: `cancel()` already has to barrier against a running callback, so a port needs a callback mutex regardless. Serializing all bodies under one is the obvious implementation and what `port/esp/timer.c` does.

If you would rather have the per-schedule version, the clean way is to stop persisting the runtime fields at all — the stored blob currently contains `timer`, `trigger_cb`, `timestamp_cb` and `priv_data`, all of which are meaningless across a reboot and one of which this PR already has to NULL out after reading. That is a storage-format change and belongs in its own PR; happy to write it up as a follow-up issue.

esp_schedule/test_app/main/test_esp_schedule.c
```diff
@@ -1575,6 +1578,151 @@ TEST_CASE("nvs restore drops invalid config", "[esp_schedule]")
...
+    TEST_ASSERT_EQUAL_UINT32_MESSAGE(1, s_rearm_fire_count, "repeating trigger must have fired exactly once");
+    TEST_ASSERT_TRUE_MESSAGE(s_canary_fired,
+                             "timer daemon must survive the re-arm (canary fired) -> no deadlock");
+    /* Fixed path also re-arms for the next occurrence (next day). */
+    TEST_ASSERT_TRUE_MESSAGE(sched->trigger.next_scheduled_time_utc >= slot + 23 * 3600,
+                             "schedule must have re-armed for the next day");
+
+    xTimerDelete(canary, portMAX_DELAY);
+    esp_schedule_delete(handle);
+    settimeofday(&tv, NULL);
```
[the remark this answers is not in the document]
Junius00: Right — the trailing `settimeofday(&tv, NULL)` is dead as written. The intent was clearly to restore, and it re-applies the pinned value instead.

Fixed in `3210d5b`, in both tests (this one and the self-delete test at the line you flagged as a duplicate — it was the same copy-paste):

```c
    struct timeval saved = {0};
    gettimeofday(&saved, NULL);
    struct timeval tv = { .tv_sec = slot - 2, .tv_usec = 0 };
    settimeofday(&tv, NULL);
    ...
    /* Restore the clock this test pinned, so a later test that reads it is not
     * silently running in June 2025. */
    settimeofday(&saved, NULL);
```

On impact: no test is affected today. These two are the only ones that touch the system clock — the other cases drive the date engine with an explicit `now` — so nothing is currently reading the leaked value. Worth fixing anyway, since the next test that does read the clock would fail for a reason with no visible connection to its own code.

shahpiyushv (commented):
Thanks for the turnaround on Vikram's and the bots' comments. The `va_list` thread in particular — where you pulled the CI `warnings.txt` and corrected your own first reading rather than leaving it — is the right way to handle a static-analysis alert.

The design is sound and I'm happy with the direction. Four inline comments below, plus two general points.

### Please fold `esp_schedule_port.h` into `esp_schedule.h`

I'd like one public header for this component rather than two. The split doesn't really hold up:

- `esp_schedule_port.h` already `#include "esp_schedule.h"` for `esp_schedule_handle_t`, so it isn't an independent header — it's a second half of one.
- `esp_schedule_init()` is declared in `esp_schedule.h` and `esp_schedule_init_with_config()` in `esp_schedule_port.h`. Two init entry points of the same component in two different headers is the part I most want fixed — anyone comparing them has to already know both files exist.
- Integrators now have to be told which header to include for what, and `examples/get_started/README.md` already has to link both.

Keep the section banners; they read well and will carry the structure inside one file just as clearly. `port/esp/esp_schedule_esp_port.h` should stay separate — that one genuinely is ESP-only and private.

### Issue #733

This gets most of the way to #733 (make SNTP init optional): `time_ops::timesync_init` is explicitly NULL-able and documented as such. But `esp_schedule_init()` still starts SNTP unconditionally, so the default path is unchanged and #733 stays open.

Closing it from here looks cheap — either a README snippet showing a port that copies `esp_schedule_esp_time_ops` with `timesync_init = NULL`, or a Kconfig guard on `esp_time_sync_init()`. Worth doing while the port layer is fresh, and it would give `esp_schedule_init_with_config()` its first in-tree caller, which it currently doesn't have at all. The test app is the natural place: `init_nvs_for_tests()` now starts SNTP on every run while most tests pin the clock with `settimeofday()`, which is latent flakiness the day a DUT has network.

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -14,243 +14,219 @@
...
-    } else {
-        ESP_LOGI(TAG, "Updating the existing schedule %s", schedule->name);
-    }
+    /* The in-memory timer handle means nothing in the next boot, and storing a
+     * live pointer would leave the restored schedule referencing freed memory.
+     * Persist the struct with that one field cleared. */
+    esp_schedule_timer_handle_t saved_timer = schedule->timer;
+    schedule->timer = NULL;
+    err = NVS_OPS()->write(handle, schedule->name, schedule, sizeof(esp_schedule_t));
+    schedule->timer = saved_timer;
```
shahpiyushv: `nvs_add()` clears `schedule->timer` on the **live** object for the duration of the NVS write. If the timer expires inside that window, the daemon runs `esp_schedule_common_timer_cb` -> `esp_schedule_start_timer` -> `esp_schedule_arm_timer` -> `port.timer.start(&schedule->timer, ...)`, which sees `*p_timer_handle == NULL` and takes the **create** branch: a second FreeRTOS timer is created and written to `schedule->timer`. Line 59 then overwrites that with `saved_timer`.

Result: the new timer and its `priv_data` leak, and two live timers now dispatch the same schedule — `esp_schedule_delete()` only cancels the one it can still see.

Reachable from `esp_schedule_edit()` on an enabled schedule (`esp_schedule_set()` -> `nvs_add`) and from `esp_schedule_enable()`, which calls `nvs_add` *after* arming. The window is one NVS write: usually a few ms, but 100 ms+ when a page needs erase/compaction.

Please write from a copy rather than mutating the live object:

```c
esp_schedule_t stored = *schedule;
stored.timer = NULL;
err = NVS_OPS()->write(handle, schedule->name, &stored, sizeof(stored));
```

~110 bytes of stack against an NVS write that already uses far more. Note the read side already clears `timer` in two places — the restore loop in `esp_schedule.c` and `esp_schedule_nvs_read_one()` below — so this write-side clearing is belt-and-braces anyway. It shouldn't be buying a race.

esp_schedule/src/esp_schedule_default.c
```diff
@@ -0,0 +1,45 @@
...
+ *          and the FreeRTOS timer, nvs_flash, esp_sntp and esp_log
+ *          dependencies come with them. This works at archive-member
+ *          granularity and therefore
+ *          does not depend on -ffunction-sections or --gc-sections.
+ *
+ *          Merging this function into esp_schedule.c, or referencing any
+ *          esp_schedule_esp_*_ops from another file, silently undoes all of
+ *          that: the tables become reachable from an object every caller needs,
+ *          and the default port is linked into every build. test_app checks
+ *          this property with nm; see test_app/main/CMakeLists.txt.
```
shahpiyushv: This is no longer true. `a66604d` removed `port_test_app` and `check_port_strip.py`, and there is no `nm` invocation left anywhere in the component — `test_app/main/CMakeLists.txt` has no POST_BUILD step.

Your reasoning for dropping the app is correct, and I checked it: the artifact glob needs a directory name that *starts* with `test_app`, and there was no pytest file or `.build-test-rules.yml` entry, so `check_port_strip.py` asserted nothing on any PR. Removing a test that doesn't test was the right call.

But it leaves three files — this one, `esp_schedule_esp_port.h`, and `esp_schedule_variables.cmake` — carrying "this is what keeps it out of the link" warnings with nothing enforcing them, and this comment pointing at a check that isn't there.

The link-strip guarantee is the headline claim of the PR, and it breaks silently with no compiler warning if anyone ever names an `esp_schedule_esp_*_ops` table outside `esp_schedule_default.c`. I'd prefer it came back here rather than as a follow-up: a POST_BUILD `nm` check on the existing `test_app` is a few lines and needs no new app, no pytest file and no rules entry. If you'd rather split it out, please fix this comment now and open the issue — Vikram called that test out as the right instinct in his review, so let's not lose it to a promise in a thread.

esp_schedule/src/esp_schedule_port.c
shahpiyushv: The first line of this comment isn't accurate — one message exceeds the buffer and is truncated unconditionally:

> `Solar schedule %s has no next occurrence (no sunrise/sunset at this location/date, no remaining matching day, or past the validity window). Enable debug logs for the cause.`

172 chars of literal, plus up to 16 for the name, minus 2 for `%s` = 186, against a 159-char payload. It truncates even with an empty name, and what gets cut is precisely the actionable half:

```
...t the validity window). Enabl
                               ^ lost: "e debug logs for the cause." (27 chars)
```

Every other call site in the component fits — the next largest expands to 134 — so it really is just this one.

Keeping 160 for the daemon-stack reason is a fine policy and I have no objection to it. But then please shorten that message so it fits, and drop "Long enough for the longest line this component emits" from the comment. If you'd rather keep the wording as-is, 192 covers it.

esp_schedule/include/esp_schedule_port.h
```diff
@@ -0,0 +1,237 @@
...
+ * @c message is NUL-terminated and owned by the caller; it is only valid for
+ * the duration of the call, so an implementation that defers must copy it.
+ *
+ * Formatting happens before this is called, so a port needs no varargs printf
+ * of its own. It also keeps va_list handling entirely inside the component:
+ * were the list passed across this boundary, an implementation that called
+ * va_end() on it would double-end the caller's list, which is undefined.
+ *
+ * Calls below the compile-time ceiling (CONFIG_ESP_SCHEDULE_LOG_LEVEL) are
+ * removed by the preprocessor and never reach this function.
```
shahpiyushv: `CONFIG_ESP_SCHEDULE_LOG_LEVEL` doesn't exist. It is in no Kconfig in the repo — I checked every one at this head — so an IDF user has no way to set it. `esp_schedule_internal.h:57` guards it with `#if defined(...)`, so it silently falls through to `CONFIG_LOG_MAXIMUM_LEVEL` and nobody notices.

The README hedges with "or `CONFIG_ESP_SCHEDULE_LOG_LEVEL` if you define one", which suggests you're aware — but this is a public header and it names the option unconditionally, as *the* ceiling.

Either way out is fine by me:
- add the option to `Kconfig`. I'd lean this way: it's the documented route to reclaiming daemon stack, and a per-component log ceiling is something people do ask for.
- or reference only `CONFIG_LOG_MAXIMUM_LEVEL` here and drop the other name.

The `esp_log_level_t` -1 offset in `esp_schedule_internal.h` is correct, including `NONE=0` -> ceiling `-1` -> silence. No issue there.

esp_schedule/src/esp_schedule_nvs.c
Junius00: Confirmed, and fixed exactly as suggested in `8906973`:

```c
    esp_schedule_t stored = *schedule;
    stored.timer = NULL;
    err = NVS_OPS()->write(handle, schedule->name, &stored, sizeof(stored));
```

Your trace is right, including the part that makes it nasty: the create branch writes the new handle into `schedule->timer`, and then line 59 puts `saved_timer` back over the top. So the second timer is not merely created, it becomes unreachable in the same breath — `esp_schedule_delete()` cancels `saved_timer` and the other one keeps dispatching a schedule that no longer exists.

And you are right that the write-side clearing was never load-bearing. The read side clears `timer` in the restore loop in `esp_schedule.c` and again in `esp_schedule_nvs_read_one()`, the latter explicitly commented as defensive cover for blobs written before the store began clearing it. So this was belt-and-braces buying a race, as you put it.

On reachability, agreed it is `esp_schedule_edit()` on an enabled schedule that makes it realistic rather than theoretical. The `esp_schedule_enable()` path needs the arm to be for an imminent expiry, and `esp_schedule_get_next_schedule_time_diff()` returns 0 — skipping timer creation — for anything already due, so that one is a narrower window than edit-while-armed.

Stack cost is `sizeof(esp_schedule_t)`, which is well under the NVS write's own frame, and this can run on the timer daemon so I checked that before taking it.

esp_schedule/src/esp_schedule_port.c
Junius00: Confirmed, with one correction to sharpen it: it does not just truncate with a name, it truncates unconditionally. The literal is 172 chars, so 170 after `%s` is removed, against 159 usable — over by 11 with an empty name before any substitution.

Brought it back under the limit in `6572f6e` rather than growing the buffer, since you have no objection to the 160 policy and the daemon-stack reason for it stands:

```c
"Solar schedule %s has no next occurrence: no solar event at this location/date, no matching day left, or past validity. Enable debug logs."
```

152 worst-case with a full 16-char name, so the actionable clause survives. I confirmed your other figure too — the next largest call site expands to 134 — so this was the only one over.

The comment you objected to is gone. It now states the constraint instead of asserting the property:

```c
/* Deliberately small: this frequently runs on the timer daemon's stack, so the
 * buffer truncates rather than grows. Every call site in the component is
 * written to fit, worst-case name length included - the longest expands to 152
 * of the 159 usable bytes. A new message longer than that will be silently cut,
 * so keep them within it rather than raising this. */
```

There is also a pointer at the call site itself, since that is where someone editing the wording will be looking.

Worth noting what this does not do: nothing enforces the 152. A future message can quietly go over and truncate again. A `_Static_assert` cannot help — the expansion length is not a compile-time constant — so the comment is the only guard. If it recurs I would rather add a `-Wformat-truncation` style check or a test that greps the literals than keep raising the buffer.

esp_schedule/include/esp_schedule_port.h
Junius00: Correct — it exists in no Kconfig, and I had confirmed that myself when writing the README line, which is exactly why it got hedged there. The hedge was the wrong call: this header states it unconditionally, as you say, and hedging in one place while asserting in another is worse than either.

Took your first option in `2790249` — the option now exists:

```
    config ESP_SCHEDULE_LOG_LEVEL_OVERRIDE
        bool "Set a log level for esp_schedule separate from the global one"
        default n

    config ESP_SCHEDULE_LOG_LEVEL
        int "esp_schedule log level"
        depends on ESP_SCHEDULE_LOG_LEVEL_OVERRIDE
        range -1 4
        default 2
```

The `depends on` is what makes this safe to add rather than behaviour-changing. A bare `config ESP_SCHEDULE_LOG_LEVEL` with `default 2` would always be defined, so `#if defined(CONFIG_ESP_SCHEDULE_LOG_LEVEL)` in `esp_schedule_internal.h` would start winning for everybody and silently detach the component from `CONFIG_LOG_MAXIMUM_LEVEL`. Gated behind the override it stays undefined unless someone opts in, so the fallback chain you traced is untouched by default.

The header now names both, in the right relationship:

```c
 * Calls above the compile-time ceiling are removed by the preprocessor and
 * never reach this function. The ceiling is CONFIG_LOG_MAXIMUM_LEVEL, or
 * CONFIG_ESP_SCHEDULE_LOG_LEVEL when CONFIG_ESP_SCHEDULE_LOG_LEVEL_OVERRIDE is
 * enabled.
```

Since the two enums are easy to confuse — which is the whole reason for that `-1` in `esp_schedule_internal.h` you checked — the help text says it outright: the value is `esp_schedule_log_level_t`, where error is 0, not `esp_log_level_t`, which reserves 0 for none. `range -1 4` admits -1 for silence.

README updated to match, and `check-kconfig-files` passes.

esp_schedule/src/esp_schedule_default.c
Junius00: Agreed on the principle — the check should exist in this PR, not in a promise. Done in `02e4923`. But your suggested mechanism cannot work, and it is worth saying why before the diff makes sense.

**`nm` on `test_app` has nothing to assert.** `test_app/main/test_app_main.c:34` calls `esp_schedule_init(true, "nvs", &restored)`. It uses the default port deliberately — the tests exercise the real FreeRTOS timer, NVS and SNTP paths. So every `esp_schedule_esp_*_ops` symbol is legitimately in that ELF, and a POST_BUILD check asserting their absence would fail on a correct build.

Demonstrating the *strip* fundamentally needs an application that installs a **non-ESP** port, because naming the ESP tables at all is what links them. That is what `port_test_app` was for, and it is why the check cannot be hung off an existing app. Re-adding it as an ELF check means re-adding an app — the thing we just agreed was not carrying its weight.

**One more correction:** the comment was never accurate. `git log -S nm -- esp_schedule/test_app/main/CMakeLists.txt` returns nothing, so no `nm` step ever existed there. The claim was wrong when it was written in `a79989c`, pointing at a check that only ever lived as a standalone script in `port_test_app`. My removal exposed it rather than broke it.

**What I did instead.** The invariant is a *source* property, so it is checked at source level — `esp_schedule/check_port_isolation.py` asserts that nothing outside `src/esp_schedule_default.c`, `port/esp/` and the port header names an `esp_schedule_esp_*_ops` table. That covers both regressions you named:

- a second referrer anywhere → the reference appears in a disallowed file
- `esp_schedule_init()` merged into `esp_schedule.c` → the tables get named in `esp_schedule.c`

Verified both ways: passes on the tree as it stands (13 sources checked), and planting `const void *leak = &esp_schedule_esp_timer_ops;` in `esp_schedule.c` fails it with the file and line.

It runs from `pre-commit` — already in this repo's CI — so no new app, no pytest file, no `.build-test-rules.yml` entry, and no dependence on artifact globs, which is what silently disabled the last attempt.

This is also strictly stricter than `nm` would have been: it catches the mistake in review rather than after a link, and it does not depend on `--gc-sections`, archive-member granularity or any toolchain behaviour.

The comment at line 28 now points at the check that exists. The design warnings in `esp_schedule_esp_port.h` and `esp_schedule_variables.cmake` I left alone — they describe the property accurately; it was only this file that named a nonexistent test.

Junius00:
Both review-body asks are done, and both gaps you found in the new check are fixed. Taking the check first, since it was my code that was wrong.

## The two check gaps — confirmed and fixed (`4695d06`)

Both reproduce exactly as you describe. I had tested the check only in the direction I expected it to fire, which is how a filter wrong in *both* directions got past me.

**The comment filter.** `line.lstrip().startswith(('*', '//'))` was wrong twice over, and you are right that neither case is contrived:

```c
*pp = &esp_schedule_esp_timer_ops;   /* was skipped as a comment continuation */
/* esp_schedule_esp_timer_ops ... */ /* was NOT skipped, so prose failed the check */
```

Replaced with a scanner that blanks comments *and* string literals while preserving line structure, so reported line numbers still match the file. Your `^\s*\*(?!\w)` would have fixed the first half; I went with real stripping because it fixes both halves and the string case at the same time — a table name quoted in a literal is not a reference either.

**Half of the invariant unenforced.** This is the better catch, and your `esp_schedule_version()` test is exactly the right probe. The guarantee is not "nobody else names the tables", it is "`esp_schedule_default.o` has exactly one external symbol". Any second export gives its callers a reason to pull the member out of the archive, and the port rides along. The file's own `@warning` says "must contain `esp_schedule_init()` and nothing else" and the check only tested the first clause.

Now checked for both functions and initialised data, with `static` definitions allowed. Test matrix, all eight passing:

| case | want | got |
|---|---|---|
| baseline | pass | pass |
| `*pp = &table;` | fail | fail |
| single-line `/* table */` | pass | pass |
| table name in a string literal | pass | pass |
| exported `esp_schedule_version()` | fail | fail |
| `static` helper in default.c | pass | pass |
| non-static `int g_counter = 0;` | fail | fail |
| restored baseline | pass | pass |

And I have written the limits into the docstring rather than leaving them for the next person to discover: half 2 is a lint that leans on the repo's astyle layout, not a C parser, and neither half sees through macros.

## Folding the headers — done (`1970700`)

Agreed on every point, particularly the two-init-entry-points one. `esp_schedule.h` is now the scheduling API followed by the porting interface, banners intact, and `esp_schedule_port.h` is gone. `port/esp/esp_schedule_esp_port.h` stays separate as you asked.

No compatibility shim, deliberately: `esp_schedule_port.h` is new in this unreleased version, so nothing outside the tree can be including it yet. Better to have never shipped two headers than to ship two and deprecate one.

The `examples/get_started/README.md` link that had to name both is now one link, which was your point about integrators needing to be told which header to include.

## Issue #733 — done (`ace05e8`)

Took the Kconfig route rather than the README snippet, because the snippet still requires the reporter to write a port, and the issue asks for it "from the init function".

`CONFIG_ESP_SCHEDULE_ENABLE_SNTP`, default `y` so existing builds are untouched. Disabled, it compiles `esp_time_sync_init()` out and leaves `timesync_init` NULL, which the core already handles as "this platform has no time sync".

The reason the `esp_sntp.h` include is guarded too is that I wanted the dependency to actually leave rather than sit there unused. Verified:

```
component time.c.obj, undefined sntp references
  CONFIG_ESP_SCHEDULE_ENABLE_SNTP=y   4
  CONFIG_ESP_SCHEDULE_ENABLE_SNTP=n   0
```

Both configurations build clean. The help text notes the constraint that remains: esp_schedule still will not arm until the clock reads past 2020, so whatever sets it must do so before schedules are enabled.

## The test app — done (`0bbeadd`)

You were right that this was the natural place, and right about the flakiness. `test_app_main.c` now installs the port through `esp_schedule_init_with_config()` — the same tables the default port uses, with `timesync_init` replaced by NULL. So the entry point has its first in-tree caller, SNTP no longer starts under tests that pin the clock, and `esp_netif_init()` is no longer needed at all.

Two smaller things that fell out of it:

- It was calling libc `free()` on the restored-handle array, which comes from `port->mem.malloc`. The same mistake the doc note two threads up exists to prevent, in the file closest to being a worked example. Now goes through the port allocator.
- `test_app/main/CMakeLists.txt` claimed `port/esp` was on the include path "for the tests that install the ESP-IDF port tables explicitly through `esp_schedule_init_with_config()`" — which no test did. That comment is now true.

## Not verified

The unit tests still have not been run against any of this. Everything builds clean for esp32 and esp32c3, and the `test_app_main.c` change in particular is a behavioural change to the harness — dropping `esp_netif_init()` — that wants a real run before you trust it. Should be run by the pipeline.

vikramdattu:
@Junius00 please add Closes #733 to PR description

Junius00:
### Note on history

The branch is now a single squashed commit. That means every SHA I quoted in the threads above (`2799761`, `bd6dfb6`, `ace05e8`, and the rest) no longer resolves — the changes are all present, but in one commit rather than the thirteen those replies were written against. Apologies for the broken links; the descriptions in each thread still match what landed.

Base is `origin/master` (`093d9b8`), deliberately, not the raw merge-base: the latter reached back past commits already in master, including an unrelated coap fix, and squashing from there would have absorbed them.

---

Beyond the review comments themselves, here is what else went in.

### From your last suggestion

**The default tables are public.** `esp_schedule_esp_port.h` moved from `port/esp/` to `include/`, so a port can be assembled from the defaults instead of written from scratch. The header's own doc block used to warn readers off it; it now shows the pattern.

The evidence this was needed was already in the tree: `test_app` had to add `port/esp` to its `PRIV_INCLUDE_DIRS` to do exactly this mix-and-match. That hack is gone, along with the matching one in the component's own `CMakeLists.txt`.

Worth being precise about one thing, since it affects how this gets documented: **a zeroed `time_sync` table is rejected.** `get_time` is required (`esp_schedule_port.c`), so the working form is to copy the default table and null the optional member:

```c
esp_schedule_port_config_t port = {
    .timer = esp_schedule_esp_timer_ops,
    .nvs   = esp_schedule_esp_nvs_ops,
    .time_sync = esp_schedule_esp_time_sync_ops,
    .mem   = esp_schedule_esp_mem_ops,
    .log   = esp_schedule_esp_log,
};
port.time_sync.timesync_init = NULL;   /* keep time(), do not start SNTP */
```

**The wall-clock group is `time_sync` throughout**, as you asked. All three names were one character from their `timer` counterparts, so all three moved together:

| | before | after |
| --- | --- | --- |
| type | `esp_schedule_time_ops_t` | `esp_schedule_time_sync_ops_t` |
| default instance | `esp_schedule_esp_time_ops` | `esp_schedule_esp_time_sync_ops` |
| `port_config` field | `.time` | `.time_sync` |

**Both SNTP mechanisms are kept**, because they are not interchangeable:

| | SNTP started | `esp_sntp` linked |
| --- | --- | --- |
| `CONFIG_ESP_SCHEDULE_ENABLE_SNTP=n` | no | **no** — 0 refs in `time.c.obj` |
| `port.time_sync.timesync_init = NULL` | no | yes — 4 refs |

Naming `esp_schedule_esp_time_sync_ops` links `port/esp/time.c`, so nulling the member stops the call but not the dependency. The Kconfig option is also the zero-code answer #733 actually asked for ("from the init function"), so it stays as the route for "no SNTP in my binary" while the NULL member covers "I'm supplying my own sync".

### Not asked for by anyone

- **`check_port_isolation.py`.** The `@warning` in `esp_schedule_default.c` pointed at an `nm` check in `test_app/main/CMakeLists.txt` that does not exist and, per `git log -S`, never did. Replaced with a check that runs from pre-commit and enforces both halves of the invariant — no second referrer in the component, and no second external symbol in `esp_schedule_default.c`. It caught the header move above on its own, which was a fair test of it.
- **The README's "Glue Layers" section was stale.** It still described `glue_*.h`, `glue/esp/` sources and a custom-`CMakeLists.txt` recipe — the design this PR deletes. Rewritten as "Port Layer", now including the mix-and-match section.
- **`port_test_app` removed.** Never wired into CI: the artifact globs match `*/examples/*` and `*/test_app*`, and `port_test_app` matches neither, so `check_port_strip.py` asserted nothing on any PR. Rationale in the thread above.
- **`port/esp/*.c` now include the header that declares their own tables.** They previously only included `esp_schedule.h`, so a signature mismatch between a table and its declaration would not have been caught in the defining translation unit. That matters more now the header is public API.
- **`CONFIG_ESP_SCHEDULE_LOG_LEVEL`** exists, gated behind an override option so the default still follows `CONFIG_LOG_MAXIMUM_LEVEL`.
- **Two `free()` sites** were calling libc `free()` on arrays allocated through `port->mem.malloc` — one in the removed test app, one in `test_app_main.c`. Both worked only because the default port wraps `malloc`.
- **A stale comment in `test_app/main/CMakeLists.txt`** claimed tests installed the port through `esp_schedule_init_with_config()`. None did, until now.

### Correction to something I posted earlier

The 4-vs-0 `esp_sntp` figure above was originally measured with a contaminated config: running `SDKCONFIG_DEFAULTS=... idf.py -B <other-dir>` writes the *project* `sdkconfig`, so a later "default" build was silently `SNTP=n` and briefly measured 0-vs-0. Re-measured with isolated `-D SDKCONFIG=` files, the original numbers hold. The figure was right; the first measurement of it was not sound.

### Not verified

The 36 unit tests still have not been run against any of this. Both `CONFIG_ESP_SCHEDULE_ENABLE_SNTP` settings build clean for esp32 and the default builds clean for esp32c3, but `test_app_main.c` now assembles its own port and no longer calls `esp_netif_init()` — a harness behaviour change that deserves a hardware run before it is trusted. I will post results when I have one.

shahpiyushv (approved):
LGTM. All review comments addressed, and the target tests are green on all six IDF versions — including the `nvs restore drops invalid config` failure that the SNTP/port-equality fix cleared up.

Nice work on the extras that weren't asked for: `check_port_isolation.py` enforcing both halves of the strip invariant, and the `#error` guard in `test_app_main.c` so the SNTP coupling can't break silently.

esp_schedule/src/esp_schedule_nvs.c
```diff
@@ -14,243 +14,225 @@
...
+     * Persist the struct with that one field cleared.
+     *
+     * Write from a copy rather than clearing the field on the live object for
+     * the duration of the write. A timer expiring inside that window reaches
+     * esp_schedule_arm_timer() -> timer.start(&schedule->timer, ...), which
+     * would see the NULL handle, take the create branch and install a second
+     * timer - then be overwritten when the field was restored, leaking that
+     * timer and leaving two of them dispatching the same schedule. */
+    esp_schedule_t stored = *schedule;
+    stored.timer = NULL;
```
vikramdattu, line 69: While you're zeroing `timer` here: `priv_data` has the same problem and isn't cleared on either side. The restore path nulls `trigger_cb`, `timestamp_cb` and `timer` (`esp_schedule.c:1113-1115`) but leaves `priv_data`, and nothing else resets it — so after a reboot `esp_schedule_get()` hands the app a pointer from the previous boot's heap.

It can't be *called* (both callbacks are nulled and guarded), but an integrator that frees or dereferences what it considers its own private data gets a wild pointer.

Pre-existing — master does exactly the same, so not something this PR introduced, and fine as a follow-up. Just cheap to fold in while the `stored` copy is right here: `stored.priv_data = NULL;` keeps it out of flash entirely.

LGTM otherwise.

vikramdattu (approved):
One small non-blocking note inline about `priv_data` (pre-existing, not from this PR).

Everything from my earlier round and Piyush's is addressed — and two of them came out better than what was suggested: the cancel fix added the early-return I'd missed, and the port-isolation check as a pre-commit hook actually runs on every PR, which the `nm`-on-`test_app` route couldn't have done since `test_app` legitimately links the default port.

esp_schedule/src/esp_schedule_nvs.c
Junius00: Confirmed, and folded in — with one addition, because the write-side clear alone does not close the case you describe.

`stored.priv_data = NULL` keeps the pointer out of blobs written **from now on**. But the failure you name is "after a reboot `esp_schedule_get()` hands the app a pointer from the previous boot's heap", and the blobs that do that are the ones already in flash — written by master, or by an earlier build of this branch. Nothing on the write side can reach those, and the restore path does not clear the field, so the first boot after an upgrade still hands out the stale pointer.

The precedent is directly above the line:

```c
/* Defensive: blobs written before the store began clearing this field carry
 * a pointer from a previous boot. */
schedule->timer = NULL;
```

`timer` needed exactly this, for exactly this reason, and the comment now applies word for word to `priv_data`. So both sides are cleared:

```c
    /* Defensive: blobs written before the store began clearing these fields
     * carry pointers from a previous boot. Clearing on the write side only
     * protects blobs written after the upgrade; anything already in flash still
     * has to be cleaned up here. priv_data matters most, since esp_schedule_get()
     * hands it straight back to the application. */
    schedule->timer = NULL;
    schedule->priv_data = NULL;
```

`esp_schedule_nvs_read_one()` is the choke point every stored blob passes through, so that covers the restore loop in `esp_schedule.c` as well without duplicating the nulling there.

Confirmed the pointer really does reach the caller — `esp_schedule.c` copies it out verbatim:

```c
schedule_config->priv_data = schedule->priv_data;
```

Both callbacks being nulled means it cannot be *invoked*, as you say, but an integrator that treats `priv_data` as owned memory and frees it gets a wild pointer, which is the worse of the two.

Agreed it is pre-existing rather than introduced here — master does the same — but it is a two-line fix sitting on top of the copy this PR just added, so it goes in now rather than as a follow-up.

PR #795, opened 2026-08-06
fix(esp_sysview): send TRACE_STOP in a separate transport block

erhankur:
Apptrace/JTAG sends data to the host when the transport is flushed. Sending buffered events and `TRACE_STOP` together can affect the final block transfer and cause apptrace down-channel errors when tracing stops.

Additionally, after a failed write, non-debug transports previously sent only the newest event. This could cause older events still waiting in the buffer to be lost.

### DIFF
--- a/esp_sysview/src/esp/SEGGER_RTT_esp.c
+++ b/esp_sysview/src/esp/SEGGER_RTT_esp.c
@@ -192,6 +192,16 @@ unsigned SEGGER_RTT_WriteSkipNoLock(unsigned BufferIndex, const void *pBuffer, u
     if (NumBytes > SYSVIEW_EVENTS_BUF_SZ) {
         return 0;  /* the caller sees the event was dropped */
     }
+
+    // Apptrace/JTAG uses each flush to exchange a transport block with the host.
+    // Flush pending events first so TRACE_STOP is sent in the next block.
+    if (event_id == SYSVIEW_EVTID_TRACE_STOP && s_events_buf_filled > 0) {
+        esp_err_t res = SEGGER_RTT_ESP_FlushNoLock();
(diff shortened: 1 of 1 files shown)


esp_sysview/src/esp/SEGGER_RTT_esp.c
```diff
@@ -192,6 +192,16 @@ unsigned SEGGER_RTT_WriteSkipNoLock(unsigned BufferIndex, const void *pBuffer, u
     if (NumBytes > SYSVIEW_EVENTS_BUF_SZ) {
         return 0;  /* the caller sees the event was dropped */
     }
+
+    // Apptrace/JTAG uses each flush to exchange a transport block with the host.
+    // Flush pending events first so TRACE_STOP is sent in the next block.
+    if (event_id == SYSVIEW_EVTID_TRACE_STOP && s_events_buf_filled > 0) {
+        esp_err_t res = SEGGER_RTT_ESP_FlushNoLock();
```
gerekon, line 199: Actually this looks like a hack. Is not it enough to flush buffer after `SYSVIEW_EVTID_TRACE_STOP`  is put into it? Why do we need to flush pending data, put `SYSVIEW_EVTID_TRACE_STOP` into buffer and flush it again?
Looks like problem with parsing `SYSVIEW_EVTID_TRACE_STOP` along with other data on host side.

gerekon (approved):

esp_sysview/src/esp/SEGGER_RTT_esp.c
erhankur: It is already fixed in the OpenOCD MR. This was additional workaround but no need now. Closing this PR.

PR #796, opened 2026-08-07
fix(coap): Support SPIRAM (IEC-577)

mrdeep1:
# Checklist

- [ ] CI passing

# Change description
Include SPIRAM support for memory allocations.


coap/port/coap_mem.c
```diff
@@ -0,0 +1,151 @@
...
+{
+}
+
+void *
+coap_malloc_type(coap_memory_tag_t type, size_t size)
+{
+    void *ptr;
+
+    (void)type;
+    ptr = MEM_ALLOC_SPIRAM(size);
```
mahavirj, line 48: Would it be possible to expose only this part to platform? Otherwise platform must track any upstream updates to this entire layer.

mrdeep1: Not entirely sure what you mean here.  Do you mean use (and define) `COAP_MEM_ALLOC_SPIRAM()` instead of `MEM_ALLOC_SPIRAM()` etc.?

Or (unlikely) do you mean this should be a part of the general release libcoap code?

mahavirj: My recommendations is to keep this layer very thin and just make the provision to override `malloc/free` calls. E.g., `COAP_MEMORY_TYPE_TRACK` can be part of the libcoap itself. This way the upstream changes won't break this layer in future. .

mrdeep1:
@mahavirj Is there anything else that needs to be done here?

mahavirj (approved):

PR #797, opened 2026-08-11
fix(catch2): detect console component via target instead of BUILD_COM… (IEC-581)

sudeep-mohanty:
Build system v2 does not support the `BUILD_COMPONENTS` variable. Instead, detect the console component through its target using generator expressions.

### DIFF
--- a/catch2/idf_component.yml
+++ b/catch2/idf_component.yml
@@ -1,4 +1,4 @@
-version: "3.7.0"
+version: "3.7.0~1"
 description: A modern, C++-native, test framework for unit-tests, TDD and BDD - using C++14, C++17 and later
 url: https://github.com/espressif/idf-extra-components/tree/master/catch2
 repository: https://github.com/espressif/idf-extra-components.git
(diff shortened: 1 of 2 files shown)


catch2/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.7.0"
+version: "3.7.0~1"
```
ESP-Marius, line 1: nitpick: I think the revision change is intended if only change is downstream packages?  This should probably be a patch change

ESP-Marius (commented):
LGTM

catch2/idf_component.yml
sudeep-mohanty: At least as per the [versioning scheme](https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/versioning.html) from the component manager, this cmake change should be classified as a downstream package change and we should update the `revision` field and not the `patch` field.

ESP-Marius: Resolved offline, current approach is correct.

suda-morris (approved):

SoucheSouche (approved):

PR #798, opened 2026-08-11
fix(spi_nand_flash): free SPI bus in Unity tearDown on test failure

RathiSonika:
# Change description
Unity aborts on assertion failure and skips in-test deinit, leaving the SPI bus initialized and causing chained failures. Share SPI setup/teardown helpers and always free the bus from tearDown .


adokitkat (approved):
lgtm

PR #800, opened 2026-08-13
fix(spi_nand_flash): handle GigaDevice Internal Data Move parity correctly

RathiSonika:
# Change description
Some GigaDevice SPI NAND parts are single-plane but require **same odd/even block parity** for Internal Data Move (IDM). The driver previously modeled that by setting `num_planes = 2` and plane-select flags, which incorrectly looks like hardware dual-plane and did not cover all affected parts.

This change introduces `NAND_FLAG_IDM_SAME_PARITY_REQUIRED` and uses it in `nand_copy()` to fall back to a RAM copy when src/dst blocks have different parity. Hardware dual-plane handling (e.g. Micron) is unchanged.

## Test plan
- [x] Build `spi_nand_flash` test app
- [x] On GD5F2GM7 / GD5F2GQ5 / GD5F4GQ6 (if available): run copy / Dhara GC across same-parity and different-parity blocks; different-parity should succeed via RAM path
- [x] On GD5F4GM8 / GD5F4GM7 (if available): same parity checks; confirm normal read/write on odd blocks still works
- [x] Micron dual-plane : confirm cross-plane copy still uses RAM path
- [x] CI green

### DIFF
--- a/spi_nand_flash/CHANGELOG.md
+++ b/spi_nand_flash/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Versioning policy: see [VERSIONING.md](VERSIONING.md). From **v1.0.0** onward this component follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.2]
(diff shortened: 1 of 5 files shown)


pacucha42 (approved):
LGTM, thanks @RathiSonika

PR #802, opened 2026-08-17
feat(freetype): enable PNG and ZLIB support

suda-morris:
## Summary
- expose ESP-IDF zlib, libpng, and FreeType components through CMake `find_package` targets
- allow FreeType to use external zlib and libpng when projects declare the dependencies
- add a color emoji example and pytest that validate PNG-backed BGRA glyph decoding

##

Now the build log will contain

```
  -- Found ZLIB: idf::zlib (found version "1.3.2")
  -- Found PNG: idf::libpng (found version "1.6.58")
```


libpng/project_include.cmake
```diff
@@ -0,0 +1,14 @@
+# Expose the IDF component through FindPNG before component targets are created.
+# A user-provided PNG library takes precedence over this shim.
+get_filename_component(component_name ${CMAKE_CURRENT_LIST_DIR} NAME)
+set(idf_png_library idf::${component_name})
+if(NOT TARGET PNG::PNG AND
+   (NOT DEFINED PNG_LIBRARY OR "${PNG_LIBRARY}" STREQUAL "${idf_png_library}"))
+    set(PNG_PNG_INCLUDE_DIR ${CMAKE_CURRENT_LIST_DIR}/libpng)
+    set(PNG_LIBRARY ${idf_png_library})
```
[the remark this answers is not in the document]
suda-morris: The shim links idf::libpng through INTERFACE_LINK_LIBRARIES. That component declares both the component root and libpng subdirectory in idf_component_register(INCLUDE_DIRS ...), so its public usage requirements provide the root for pnglibconf.h and libpng/ for png.h transitively. A reconfigured FreeType build includes both paths; no change is needed.

libpng/project_include.cmake
```diff
@@ -0,0 +1,14 @@
...
+
+    add_library(PNG::PNG INTERFACE IMPORTED)
```
igrr, line 9: @suda-morris could you please check if the option suggested here (CMAKE_FIND_PACKAGE_REDIRECTS_DIR) works? https://github.com/espressif/idf-extra-components/pull/320#discussion_r1574882171

igrr: And is it okay that we aren't setting PNG_FOUND?

freetype/project_include.cmake
```diff
@@ -0,0 +1,16 @@
+# Expose the IDF component through FindFreetype before component targets are created.
+# A user-provided library takes precedence over this shim.
```
igrr: Not automatically, though, because the order of project_include.cmake evaluation is not guaranteed? One would have to define it before freetype/project_include.cmake is evalutated.

freetype/CMakeLists.txt
igrr: Could you please also check if the two -Wno flags below are still needed in the latest version?

igrr (approved):

libpng/project_include.cmake
suda-morris: CMAKE_FIND_PACKAGE_REDIRECTS_DIR was introduced since cmake 3.24, but IDF's minmal cmake version is 3.22

suda-morris: PNG_FOUND / ZLIB_FOUND are set automatically by FindPNG and FindZLIB

suda-morris: I tried the CMAKE_FIND_PACKAGE_REDIRECTS_DIR variable, and it also works, but I don't want to introduce a hybrid approach right now because of the minimal CMake version constraint.

PR #803, opened 2026-08-17
fix(flash_dispatcher): fix blocking when directly calling the flash i… (IEC-583)

udoudou:
Fixed an issue where directly calling the flash interface without first reading and loading the partition table would cause blocking.

The internal Flash erase and write interfaces check whether the operation range is protected. If the partition table has not yet been loaded, calling the `esp_flash_write`, `esp_flash_write_encrypted`, or `esp_flash_erase_region` APIs triggers an internal execution of `spi_flash_mmap` to load the partition table. Repeated execution of the scheduler results in a deadlock.

# Checklist

- [ ] CI passing

# Change description
Allow tasks whose own task stack resides in internal DRAM to directly invoke actual flash operations. Avoid repeatedly executing the scheduler and unnecessary task scheduling.

### DIFF
--- a/esp_flash_dispatcher/CHANGELOG.md
+++ b/esp_flash_dispatcher/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.3
+
+- fix: Bypass the dispatcher and call the real Flash operations directly when the stack resides in internal DRAM. This prevents deadlocks that could arise from nested calls to flash operation interfaces (e.g. specifically, calling a flash operation from within a flash operation already executing under the scheduler) and avoids unnecessary task scheduling overhead.
+
 ## 1.0.2
 
 - fix: Bypass the dispatcher and call the real Flash operations directly when it is not initialized yet, so early-boot Flash access (e.g. partition loading and core dump probing during system startup).
--- a/esp_flash_dispatcher/esp_flash_dispatcher.c
+++ b/esp_flash_dispatcher/esp_flash_dispatcher.c
@@ -16,6 +16,7 @@
 #include "esp_flash_dispatcher.h"
 #include "spi_flash_mmap.h"
 #include "esp_private/spi_flash_os.h"
+#include "esp_memory_utils.h"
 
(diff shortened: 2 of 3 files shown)


mythbuster5:
LGTM

mythbuster5:
@udoudou Thanks, I'm doing same thing as well. Let's merge your mr.

esp_flash_dispatcher/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.3
+
+- fix: Bypass the dispatcher and call the real Flash operations directly when the stack resides in internal DRAM. This prevents deadlocks that could arise from nested calls to flash operation interfaces (e.g. specifically, calling a flash operation from within a flash operation already executing under the scheduler) and avoids unnecessary task scheduling overhead.
```
mythbuster5, line 3: would it will cause a real deadlock or just a kind of waste of time?

udoudou: It will become completely blocked and seize up.
7d7e4507-5515-4418-8797-78b20b126893

esp_flash_dispatcher/esp_flash_dispatcher.c
```diff
@@ -103,7 +104,7 @@ static flash_dispatcher_context_t s_flash_dispatcher_ctx;
 // loading / core dump probing).
 static inline bool flash_dispatcher_should_bypass(void)
 {
-    return flash_dispatcher_is_no_os() || xTaskGetSchedulerState() != taskSCHEDULER_RUNNING;
+    return flash_dispatcher_is_no_os() || esp_ptr_in_dram(esp_cpu_get_sp()) || xTaskGetSchedulerState() != taskSCHEDULER_RUNNING;
```
igrr, line 103: Can also remove the comment above this function, because it is no longer matching what the function does.

mythbuster5 (approved):

PR #807, opened 2026-08-20
Sysview optimizations

erhankur:
These optimizations reduce per event SystemView tracing cost from ~9 µs to ~2 µs together with the esp_trace changes

- Compile SystemView hot path at -O2 (CONFIG_SEGGER_SYSVIEW_OPTIMIZE_SPEED, default on).
- Poll the down channel every 32nd event instead of on every event.
- Cache encoder pointer and link type at SetEncoder; load transport/ctx from the encoder.
- Inline those accessors; copy packets with a byte loop instead of memcpy.

Credits to @igrr

### DIFF
--- a/esp_sysview/src/esp/SEGGER_RTT_esp.c
+++ b/esp_sysview/src/esp/SEGGER_RTT_esp.c
@@ -31,6 +31,8 @@
 #define SEGGER_HOST_WAIT_TMO    CONFIG_SEGGER_SYSVIEW_BUF_WAIT_TMO
 #endif
 
+unsigned SEGGER_RTT_ESP_DownPollCnt;
+
 static uint8_t s_events_buf[SYSVIEW_EVENTS_BUF_SZ];
 static uint16_t s_events_buf_filled;
 static uint8_t s_down_buf[SYSVIEW_DOWN_BUF_SIZE];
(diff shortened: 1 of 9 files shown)


esp_sysview/src/esp/SEGGER_RTT_esp.c
```diff
@@ -215,7 +218,9 @@ unsigned SEGGER_RTT_WriteSkipNoLock(unsigned BufferIndex, const void *pBuffer, u
         }
         s_events_buf_filled = 0;
     }
-    memcpy(&s_events_buf[s_events_buf_filled], pBuffer, NumBytes);
+    for (unsigned k = 0; k < NumBytes; k++) {
```
sobuch, line 221: does this make any difference? comparing loop vs memcpy locally, I get about 10x slower results for loop without -O2 and just around the same with -O2 (tried p4,s3)

I would think in case we see a speedup here, that we need an optimized version of memcpy

sobuch (approved):
Looks ok to me, might be good to run the openocd pipeline with this version if possible

esp_sysview/src/esp/SEGGER_RTT_esp.c
erhankur: I haven't benchmarked it myself. According to Ivan's analysis [here](https://hedgedoc.espressif.tools/s/4VlVLhry6#Where-the-1024-cycles-went), the optimization saved around 80–300 cycles on ESP32-C3.

erhankur:
> Looks ok to me, might be good to run the openocd pipeline with this version if possible

I did a fix to catch START command.  Fast path (poll every 32nd event) must only run while tracing is active. Otherwise the target can miss the host's START command.

OpenOCD and esp-idf tests are fine with the latest version.

PR #808, opened 2026-08-20
feat(eigen): upgrade to v5.0.1 and improve component quality

suda-morris:
## Summary

- Upgrade the bundled Eigen submodule to upstream v5.0.1.
- Simplify the ESP-IDF component integration for Eigen's header-only API.
- Replace the SVD example with a focused `get_started` example.
- Add SBOM metadata and build-test rules.
- Refresh the README and example documentation.

## Example coverage

The get-started example demonstrates the following:

- Matrix and vector basics
- Linear system solving
- Sensor calibration with QR and SVD least-squares fitting
- Quaternion-based 3D rotation
- Fixed-size and dynamic-size matrices

The example does not build the optional upstream BLAS/LAPACK compatibility libraries.


suda-morris:
@igrr PTAL when you have time. Later we plan to port the IKFast library, which consumes this component.

PR #809, opened 2026-08-21
feat(lz4): add ESP-IDF LZ4 component

suda-morris:
## Summary

- Add the upstream LZ4 v1.10.0 library as an ESP-IDF component.
- Build the Block, HC, Frame, file, and checksum library sources as an IDF static component without pulling in upstream CLI, test, or install targets.
- Configure embedded-safe heap-backed workspaces and an optional PSRAM-aware allocator through Kconfig.
- Add a beginner-friendly Block and Frame streaming example.
- Add Unity integration tests, incremental Frame tests, HC coverage, PSRAM coverage for ESP32-S3, and benchmark metrics.
- Add pytest coverage for the test app and example.

## Validation

- `idf.py build` for `lz4/test_apps/test_lz4`
- `idf.py build` for `lz4/examples/pcm_lz4`
- Host compilation with GCC warnings enabled
- Python syntax checks for pytest scripts

The test app is enabled for representative Xtensa and RISC-V targets. HC and PSRAM paths are covered according to target memory capabilities and configuration.


suda-morris:
@igrr @mahavirj PTAL

lz4/README.md
```diff
@@ -0,0 +1,85 @@
...
+For function contracts, streaming, format details, and security notes, use the official [LZ4 documentation](https://github.com/lz4/lz4/tree/dev/doc) and the headers.
+
+> [!NOTE]
+> The bundled `xxhash.c` is an internal dependency of the Frame implementation.
+> Its API is not part of this component. Depend on a dedicated xxHash component
+> if the application needs xxHash directly.
+
+## On ESP-IDF
+
+Workspaces live on the heap so they do not consume task stack. Typical sizes:
```
igrr: One sentence here to introduce the term "workspace" in the context of LZ4 would be helpful

lz4/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+```c
+LZ4_stream_t *state = LZ4_createStream();
+int n = LZ4_compress_fast_extState(state, src, dst, src_size, dst_capacity, 1);
+LZ4_freeStream(state);
+```
+
+Do not place `LZ4_stream_t` or an HC stream on a small FreeRTOS task stack.
+
+Every task must use its own contexts. Caller buffers need no extra locking unless they are shared.
```
igrr: Is context here the same as workspace?

igrr (approved):

lz4/README.md
suda-morris: context >= workspace. I restructured this part in the README and removed the confusion. These details are not necessary for the user.

PR #811, opened 2026-08-27
feat(esp_lcd_screenshot): add virtual LCD panel driver for capturing screenshots

suda-morris:
## Summary

- add a virtual LCD panel driver for capturing display output as screenshots
- add an LVGL screenshot example and golden-image test
- register the component and CI/build-test metadata

## Testing

- Added the `lvgl_screenshot` example and pytest golden-result test.


esp_lcd_screenshot/src/esp_lcd_screenshot.c
```diff
@@ -0,0 +1,419 @@
...
+
+static esp_err_t screenshot_draw_bitmap(esp_lcd_panel_t *panel,
+                                        int x_start, int y_start,
+                                        int x_end, int y_end,
+                                        const void *color_data)
+{
+    ESP_RETURN_ON_FALSE(color_data, ESP_ERR_INVALID_ARG, TAG, "color_data is NULL");
+    screenshot_panel_t *drv = (screenshot_panel_t *)panel;
+    int bpp = drv->bytes_per_pixel;
+
```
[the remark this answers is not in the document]
suda-morris: this has been guaranteed by the esp_lcd interface

esp_lcd_screenshot/src/esp_lcd_screenshot.c
```diff
@@ -0,0 +1,496 @@
...
+    char fourcc_str[5];
+    screenshot_fourcc_str(drv->color_format, fourcc_str);
+    fprintf(stream, "FRAMEBUFFER_BEGIN %d %d %s\n", drv->width, drv->height, fourcc_str);
+
+    /* Encode in chunks: 3072 input bytes -> 4096 base64 chars per line.
+     * Each payload line carries the FB_BASE64 prefix so that a host parsing
+     * the serial output can tell it apart from interleaved log lines. */
+    const size_t chunk_in = 3072;
+    char b64_line[4100];
```
igrr: Can we allocate this from heap? Looks like a nontrivial amount to reserve on the stack, this could easily lead to a stack overflow. Or, after using this feature, someone might forget to reduce the stack back to the original size, wasting RAM.

esp_lcd_screenshot/examples/lvgl_screenshot/main/lvgl_screenshot_main.c
```diff
@@ -0,0 +1,170 @@
...
+    lv_display_set_buffers(display, draw_buf, NULL, sizeof(draw_buf), LV_DISPLAY_RENDER_MODE_PARTIAL);
+
+    ESP_LOGI(TAG, "Create a static demo UI and render it once");
+    example_create_ui();
+    /* Force an immediate refresh instead of waiting for an LVGL timer tick. */
+    lv_refr_now(display);
+
+    /* The PNG is written to the LittleFS mount initialized above. */
+    ESP_LOGI(TAG, "Save the captured frame as a PNG file");
+    ESP_ERROR_CHECK(esp_lcd_screenshot_save_png(panel_handle, EXAMPLE_PNG_PATH));
```
igrr, line 159: It looks okay when we rely on lv_refr_now rather than a refresh happening in LVGL task.

In a real scenario where LVGL or another library works in a task, how do we know when to call esp_lcd_screenshot_save_png or esp_lcd_screenshot_dump_base64? Seems like we need to have some post-draw hook to take the screenshot when the frame is done.

igrr (approved):

PR #813, opened 2026-08-27
fix(spi_nand_flash): correct ECC status extraction in PACK_2BITS_STATUS and PACK_3BITS_STATUS (IEC-591)

worblehat:
# Summary

The number of corrected ECC bit is read from two or three bits of the STATUS register.
`PACK_2BITS_STATUS` and `PACK_3BITS_STATUS` in `spi_nand_flash/src/nand_impl.c` shifted the masked ECC status bits in the wrong direction. The resulting value never equals any `nand_ecc_status_t` enumerator, so every comparison against that enum fails. As a consequence the driver never reports an uncorrectable read error, and the data refresh threshold never triggers.

# Root cause

The ECC status bits are defined at bit 4 and above in the status register (`spi_nand_oper.h`):

```c
#define STAT_ECC0           1 << 4
#define STAT_ECC1           1 << 5
#define STAT_ECC2           1 << 6
```

The macros to extract these bits from the register value masks them and then shift them further
left instead of down to positions 0-2:

```c
#define PACK_3BITS_STATUS(status, bit2, bit1, bit0) \
    ((((status) & (bit2)) << 2) | (((status) & (bit1)) << 1) | ((status) & (bit0)))
```

For the five defined values in the ECC status bits this results in: 

| ECC Bits  | Meaning                               | Macro result should be  | Macro result  is|
|-----------|---------------------------------------|---------------------|--------------|
| `000`     | no errors                             | 0                   | 0            |
| `001`     | 1-3 bits corrected                    | 1                   | **16**       |
| `010`     | uncorrectable                         | 2                   | **64**       |
| `011`     | 4-6 bits corrected                    | 3                   | **80**       |
| `101`     | 7-8 bits corrected                    | 5                   | **272**      |

The `is_ecc_error()` tries to map the extracted values to `nand_ecc_status_t` values:

```
NAND_ECC_OK = 0,
NAND_ECC_1_TO_3_BITS_CORRECTED = 1,
NAND_ECC_NOT_CORRECTED = 2,
NAND_ECC_4_TO_6_BITS_CORRECTED = 3,
NAND_ECC_7_8_BITS_CORRECTED = 5,
```

Only `000` can be mapped correctly.

# Impact

`is_ecc_error()` cannot match `NAND_ECC_NOT_CORRECTED`, so `nand_read()` returns `ESP_OK` for a
page the ECC could not correct, and passes the corrupt data to the caller.
Therefore `dhara_glue.c` never refreshes pages with faulty bits and never detects an uncorrectable read.
An unrecoverable read error is reported as success.

# Checklist

- [ ] CI passing

### DIFF
--- a/spi_nand_flash/CHANGELOG.md
+++ b/spi_nand_flash/CHANGELOG.md
@@ -4,9 +4,14 @@ All notable changes to this component will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). See [VERSIONING.md](VERSIONING.md) for this component's versioning policy.
 
-## [Unreleased]
+## [1.4.4] - 2026-09-10
+
+### Fixed
+
+- Extraction of ECC status bits
 
 ### Changed
+
 - Demote per-page/per-block traces (`is_bad`, `erase_block`, `is_free`) from debug to verbose log level, since they flooded the log during any full-chip scan or erase
 
 ## [1.4.3]
@@ -147,4 +152,3 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Initial release with basic NAND flash support based on dhara nand library
 - Support for Winbond, Alliance and Gigadevice devices
 - Added test application for the same
-
--- a/spi_nand_flash/src/nand_impl.c
+++ b/spi_nand_flash/src/nand_impl.c
@@ -442,8 +442,13 @@ fail:
     return ret;
 }
 
-#define PACK_2BITS_STATUS(status, bit1, bit0)         ((((status) & (bit1)) << 1) | ((status) & (bit0)))
-#define PACK_3BITS_STATUS(status, bit2, bit1, bit0)   ((((status) & (bit2)) << 2) | (((status) & (bit1)) << 1) | ((status) & (bit0)))
+// Collapse each ECC status flag down to 0b0 or 0b1 regardless of its original position,
+// shift it into the correct place (bit n) and pack the result into a nand_ecc_status_t value.
+#define PACK_2BITS_STATUS(status, bit1, bit0)         ((!!((status) & (bit1)) << 1) | \
+                                                        !!((status) & (bit0)))
+#define PACK_3BITS_STATUS(status, bit2, bit1, bit0)   ((!!((status) & (bit2)) << 2) | \
+                                                       (!!((status) & (bit1)) << 1) | \
+                                                        !!((status) & (bit0)))
 
 static bool is_ecc_error(spi_nand_flash_device_t *dev, uint8_t status)
 {
--- a/spi_nand_flash/idf_component.yml
+++ b/spi_nand_flash/idf_component.yml
@@ -1,4 +1,4 @@
-version: "1.4.3"
+version: "1.4.4"
 description: Driver for accessing SPI NAND Flash
 url: https://github.com/espressif/idf-extra-components/tree/master/spi_nand_flash
 issues: https://github.com/espressif/idf-extra-components/issues
--- a/spi_nand_flash/priv_include/spi_nand_oper.h
+++ b/spi_nand_flash/priv_include/spi_nand_oper.h
@@ -50,13 +50,13 @@ typedef struct spi_nand_transaction_t spi_nand_transaction_t;
 #define REG_CONFIG          0xB0
 #define REG_STATUS          0xC0
 
-#define STAT_BUSY           1 << 0
-#define STAT_WRITE_ENABLED  1 << 1
-#define STAT_ERASE_FAILED   1 << 2
-#define STAT_PROGRAM_FAILED 1 << 3
-#define STAT_ECC0           1 << 4
-#define STAT_ECC1           1 << 5
-#define STAT_ECC2           1 << 6
+#define STAT_BUSY           (1 << 0)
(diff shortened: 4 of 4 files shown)


spi_nand_flash/src/nand_impl.c
```diff
@@ -412,8 +412,9 @@ esp_err_t nand_is_free(spi_nand_flash_device_t *handle, uint32_t page, bool *is_
     return ret;
 }
 
-#define PACK_2BITS_STATUS(status, bit1, bit0)         ((((status) & (bit1)) << 1) | ((status) & (bit0)))
-#define PACK_3BITS_STATUS(status, bit2, bit1, bit0)   ((((status) & (bit2)) << 2) | (((status) & (bit1)) << 1) | ((status) & (bit0)))
+#define ECC_STATUS_SHIFT 4
+#define PACK_2BITS_STATUS(status, bit1, bit0)         (((status) & ((bit1) | (bit0))) >> ECC_STATUS_SHIFT)
+#define PACK_3BITS_STATUS(status, bit2, bit1, bit0)   (((status) & ((bit2) | (bit1) | (bit0))) >> ECC_STATUS_SHIFT)
```
omnitex, line 417: Hard-coded shift agnostic version which uses `!!` to collapse the masked bits down to 0/1 and than shift them to packed positions. 

This should be a cleaner way to achieve the correct shift?

```suggestion
// Each STAT_ECCx bit flag may live at an arbitrary bit position in the status byte.
// `!!(status & bitN)` collapses that flag down to 0b0 or 0b1 regardless of its
// original position, so it can then be shifted into its correct place (bit N)
// in the packed result, producing a plain 0..3 / 0..7 value matching nand_ecc_status_t.
#define PACK_2BITS_STATUS(status, bit1, bit0)         ((!!((status) & (bit1)) << 1) | !!((status) & (bit0)))
#define PACK_3BITS_STATUS(status, bit2, bit1, bit0)   ((!!((status) & (bit2)) << 2) | (!!((status) & (bit1)) << 1) | !!((status) & (bit0)))
```

RathiSonika:
@worblehat Thank you for finding this bug and for the PR — the root cause analysis looks correct.

I agree with @omnitex suggestion: rather than a hard-coded ECC_STATUS_SHIFT, using !! to collapse each masked bit to 0/1 and then packing into the result is clearer and stays correct if the STAT_ECC* positions ever change. Could you update the macros to that form?

Also, could you please:
1. Rebase the branch onto the latest master
2. Bump the component version in idf_component.yml
3. Add a corresponding entry under ## [1.4.x] in CHANGELOG.md

worblehat:
Done :+1: Please review again.

worblehat:
Another thing I noticed:
The `STAT_*` bit definitions in [spi_nand_oper.h](https://github.com/espressif/idf-extra-components/blob/master/spi_nand_flash/priv_include/spi_nand_oper.h#L53-L59) should probably have parantheses around the shifts (e.g. `(1 << 0)` instead of `1 << 0`) to avoid unintended operator precedence when the caller does not put parantheses around the macro.

What do you think? Can I add this to this PR?

omnitex:
> Another thing I noticed: The `STAT_*` bit definitions in [spi_nand_oper.h](https://github.com/espressif/idf-extra-components/blob/master/spi_nand_flash/priv_include/spi_nand_oper.h#L53-L59) should probably have parantheses around the shifts (e.g. `(1 << 0)` instead of `1 << 0`) to avoid unintended operator precedence when the caller does not put parantheses around the macro.
> 
> What do you think? Can I add this to this PR?

It is related to the main issue at hand, and is worth to harden going forward, so yes, please go ahead, thank you

worblehat:
Done.

omnitex (approved):
Please reword the first commit's message to something like 

`fix(spi_nand_flash): correct ECC status extraction in PACK_*BITS_STATUS` 

so it adheres to the standard structure. Otherwise LGTM, thank you!

omnitex:
@RathiSonika PTAL

spi_nand_flash/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 Versioning policy: see [VERSIONING.md](VERSIONING.md). From **v1.0.0** onward this component follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.4]
+- fix: extraction of ECC status bits
+
```
omnitex:
```suggestion
## [1.4.4] - 2026-09-10

### Fixed

- extraction of ECC status bits

```

Hi @worblehat, can you please rebase on the latest master, and see if you can tweak the changelog entry to align with https://keepachangelog.com/en/1.1.0/ ? Ideally keep the date correct to when you actually do the change. Thank you!

worblehat: Done. Thanks for fixing the tests!

omnitex: Thank you for quick response! You might have to sync your fork first, then rebase again. Right now it still shows `This branch is out-of-date with the base branch`

omnitex: @worblehat When you rebase on synced fork, there might be a conflict here, https://github.com/espressif/idf-extra-components/blob/master/spi_nand_flash/CHANGELOG.md?plain=1#L7

Please change the `[Unreleased]` -> `[1.4.4] - 2026-09-10` suggested above. This release will also contain a change made during fixing the tests, together with your fix.

worblehat: Sorry. My fault. I rebased onto the wrong branch.

RathiSonika (approved):

PR #818, opened 2026-09-01
fix(spi_nand_flash): add STAT_BUSY hang timeout in wait_for_ready

RathiSonika:
# Change description
## Summary
- Bound `wait_for_ready()` so a stuck/missing chip cannot hang forever on `STAT_BUSY` (e.g. after `PAGE_READ`).
- Hang limit is **10×** the vendor-table typical delay (`tR` / `tPROG` / `tBERS`), ceiled to FreeRTOS ticks, plus one tick; success path still polls until `BUSY` clears and returns `ESP_ERR_TIMEOUT` on expiry.
## Why
`STAT_BUSY` is not guaranteed to clear (no chip, stuck status, etc.). An unbounded poll loops forever. Using typical delay alone as the timeout is too tight (e.g. Winbond erase typ ~2.5 ms, datasheet max ~10 ms), so the bound is 10× typical.

### DIFF
--- a/spi_nand_flash/CHANGELOG.md
+++ b/spi_nand_flash/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Versioning policy: see [VERSIONING.md](VERSIONING.md). From **v1.0.0** onward this component follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+## [1.4.3]
+- fix: add STAT_BUSY hang timeout in wait_for_ready
+
 ## [1.4.2]
 - fix: use Internal Data Move same-parity handling for GigaDevice chips that require it (GD5F2GQ5, GD5F2GM7, GD5F4GQ6, GD5F4GM8, GD5F4GM7, GD5F8GM8), without treating them as hardware dual-plane
 
--- a/spi_nand_flash/src/nand_impl.c
+++ b/spi_nand_flash/src/nand_impl.c
@@ -3,10 +3,11 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  *
- * SPDX-FileContributor: 2015-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileContributor: 2015-2026 Espressif Systems (Shanghai) CO LTD
  */
 
 #include <string.h>
+#include <inttypes.h>
 #include "esp_check.h"
 #include "esp_err.h"
 #include "spi_nand_oper.h"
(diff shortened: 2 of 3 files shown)


spi_nand_flash/src/nand_impl.c
```diff
@@ -180,6 +195,11 @@ static esp_err_t wait_for_ready(spi_nand_flash_device_t *dev, uint32_t expected_
             break;
         }
 
+        if ((xTaskGetTickCount() - start_tick) >= timeout_ticks) {
+            ESP_LOGE(TAG, "STAT_BUSY timeout (expected %" PRIu32 " us)", expected_operation_time_us);
```
omnitex: slightly more helpful debug print

```suggestion
            ESP_LOGE(TAG, "STAT_BUSY timeout: status=0x%02x, expected_op=%" PRIu32 " us, waited=%" PRIu32 " ms",
              status, expected_operation_time_us, timeout_ms);
```

omnitex (approved):

spi_nand_flash/src/nand_impl.c
```diff
@@ -180,6 +195,12 @@ static esp_err_t wait_for_ready(spi_nand_flash_device_t *dev, uint32_t expected_
             break;
         }
 
+        if ((xTaskGetTickCount() - start_tick) >= timeout_ticks) {
+            ESP_LOGE(TAG, "STAT_BUSY timeout: status=0x%02x, expected_op=%" PRIu32 " us, waited=%" PRIu32 " ms",
```
pacucha42, line 199: I think this is technically wrong - the actual time elapsed is `timeout_ticks * portTICK_PERIOD_MS` (parameter `waited=%" PRIu32 " ms"`)

pacucha42 (changes requested):
LGTM, needs just one small fix. Thanks

spi_nand_flash/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 Versioning policy: see [VERSIONING.md](VERSIONING.md). From **v1.0.0** onward this component follows [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html).
 
+# [1.4.3]
```
pacucha42: Sorry, missed this one before: all previous updates are '##', this one is just '#'... pure nit, however, have a look. Thanks

RathiSonika: Yes, thank you for noticing. Fixed

pacucha42 (approved):
Thanks for addressing my comments. Approved

PR #823, opened 2026-09-03
add Raylib (IEC-600)

georgik:
The Raylib is very popular graphical library and thanks to software renderer now works nicely with ESP devices with PSRAM.

Formerly it was hosted here: https://github.com/georgik/esp-idf-component-raylib

Component Registry: https://components.espressif.com/components/georgik/raylib/

### DIFF
--- a/raylib/examples/esp32/m5stack-core2_hello/main/idf_component.yml
+++ b/raylib/examples/esp32/m5stack-core2_hello/main/idf_component.yml
@@ -1,0 +1,13 @@
+## IDF Component Manager Manifest File
+dependencies:
+  idf:
+    version: ">=5.5.0"
+  # Local raylib component - relative path from examples/esp_board_hello/main/
+  # Structure: raylib/examples/esp_board_hello/main/ -> raylib/ (up 3 levels)
+  georgik/raylib:
+    version: "*"
+    override_path: "../../../.."
+  # New boards from esp-bsp (ESP32-S3 only)
+  espressif/m5stack_core_2:
+    version: "*"
+  # P4 boards skipped - esp-generate doesn't support esp32p4 chip yet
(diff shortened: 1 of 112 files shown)


georgik:
@raysan5 The PR is tracking working on meging ESP-IDF wrapper component for Raylib into set of official components which are available under `espressif` namespace - https://components.espressif.com/components?q=namespace:espressif

The PR is following simplified architecture without "port" layer as was discussed in June.

I've updated the repo to current master to include latest changes.

raylib/examples/esp32/m5stack-core2_hello/main/idf_component.yml
```diff
@@ -0,0 +1,13 @@
+## IDF Component Manager Manifest File
+dependencies:
+  idf:
+    version: ">=5.5.0"
+  # Local raylib component - relative path from examples/esp_board_hello/main/
+  # Structure: raylib/examples/esp_board_hello/main/ -> raylib/ (up 3 levels)
+  georgik/raylib:
```
adokitkat, line 7: I think this should change to `espressif/raylib`? Also in all other instances.

suda-morris (commented):
Thanks for your work on this PR. I appreciate the effort. My view is that examples should mainly highlight the component’s API usage, not aim to run out‑of‑the‑box on every board. Having multiple board directories essentially repeats the same demonstration — supporting one generic board (or a single official board) achieves the same purpose. Also, since you’ve already provided usage instructions for Wokwi simulation, the example is “out‑of‑the‑box” without hardware. Given that, I’d recommend keeping just one board example.

igrr:
I second this, let's keep one simple example here for now. We can create another repo for the more comprehensive examples, esp-generate usage and multi-board support (e.g. espressif/raylib-examples).

Also please take a look at build failures.

raysan5:
> The PR is tracking working on meging ESP-IDF wrapper component for Raylib into set of official components which are available under espressif namespace

@georgik That's fantastic! Thanks for all the hard work to make it possible and on the new portable architecture! Hopefully it will allow new boards to be added more easely! 💯

PR #838, opened 2026-09-13
feat(esp_ext_part_tables): Release stable v1.0.0 version

adokitkat:
# esp_ext_part_tables: release 1.0.0

Prepares the first stable release. The API is frozen at 1.0, so the renames and
signature changes have to happen now. Also fixes four bugs and adds format detection
and on-target test coverage.

## Commits

Four commits, each building and passing the suite on its own:

1. `refactor!: freeze and clean up the public API` — renames, `const`, typed signature accessors.
2. `fix: make the generated table a pure function of the list` — the behavioural fixes.
3. `refactor!: replace the type-erased block device helpers` — typed helpers + `esp_ext_part_probe`.
4. `chore: test on target under QEMU and release 1.0.0` — coverage, version bump, CHANGELOG.

## Breaking changes

| Before | After |
|---|---|
| `MBR_SIZE`, `MBR_SIGNATURE`, `MBR_CHS_*`, ... | `ESP_MBR_*` |
| `mbr_t`, `mbr_partition_t` | `esp_mbr_t`, `esp_mbr_partition_t` |
| `esp_mbr_remove_gaps_between_partiton_entries` | `..._partition_entries` (typo) |
| `..._signature_get/set(list, void *buf, [type])` | typed `esp_ext_part_list_signature_t *` |
| `esp_ext_part_list_bdl_read/write(h, list, TYPE_MBR, void *)` | `esp_mbr_bdl_read/write(h, list, <typed args>)` |

Both `void *` parameters are gone. A second table format would now be added as new
functions rather than a new enumerator — additive, not breaking.

Three functions now reject input they used to accept silently:

- `esp_mbr_parse` and `esp_ext_part_list_deep_copy` return `ESP_ERR_INVALID_STATE` on a
  non-empty destination list. Both previously overwrote the list head, leaking its items.
- `esp_mbr_partition_set` returns `ESP_ERR_INVALID_ARG` when `sector_size` is unset; it
  used to emit a zero-length partition at LBA 0. `esp_mbr_generate` is unaffected.

## Bug fixes

Generating into a buffer that already held an MBR left parts of the old table behind —
the read-modify-write case `keep_signature` exists for:

- Entries not covered by the list are zeroed; removed partitions used to reappear on the
  next parse.
- Each entry is built in a zeroed local and committed only after all checks pass, so no
  stale `status`/CHS bytes survive and a rejected entry leaves the MBR untouched.
- The copy-protection marker now follows `ESP_EXT_PART_LIST_FLAG_READ_ONLY` both ways.

Bootstrap code is still preserved. Separately, `esp_hw_support` (for `esp_random`) and
`<stdlib.h>` were never declared and only worked transitively.

## Format detection

The old read function took the format as an input, which a caller looking at an unknown
card cannot supply. `esp_ext_part_probe(handle, &type)` reads the first sector and
reports MBR or GPT (a GPT disk's protective MBR has a `0xEE` entry).

This also documents previously unstated behaviour: **parsing a GPT disk as MBR
succeeds**, returning one `ESP_EXT_PART_TYPE_GPT_PROTECTIVE_MBR` partition that is not
the real table. Callers who may see GPT media must check for it.

## Testing

The suite only ran on the linux host. Adding a QEMU run found three target-only problems,
all fixed here: `TEST_ASSERT_*_UINT64` expanded to an unconditional failure on chips
(`CONFIG_UNITY_ENABLE_64BIT` off), two assertions compared pointers through an integer
macro, and the BDL test leaked 568 bytes by never releasing its handle.

Suite grew 37 → 44 cases, including three functions that previously had no test at all
(`esp_ext_part_list_deep_copy`, `esp_mbr_lba_to_chs_arr`, `esp_mbr_chs_arr_val_set/get`).

| Configuration | Result |
|---|---|
| linux host, from clean | 44/44, no warnings |
| esp32s3 under QEMU | 40/40, no warnings under `-Werror` — run at an earlier revision, before the newest four cases |
| `examples/basic` on linux | builds and runs as its pytest expects |

Minimum ESP-IDF raised 5.1 → 5.2, the oldest version CI actually builds.

## For reviewers

1. The copy-protection marker is now cleared as well as set. Those two bytes overlap the
   classical bootstrap area, so on a disk with classical boot code this writes two zero
   bytes previously left alone.
2. `ESP_EXT_PART_LIST_SIGNATURE_GPT` names a format this component cannot read or write.
   It exists so `probe` can return something other than "MBR". The alternative is to drop
   `probe` and have callers check the parsed result for a protective MBR.
3. `esp32c3` is in the QEMU parameters but was only verified on `esp32s3` locally.

# Checklist

- [x] CI passing


esp_ext_part_tables/include/esp_mbr.h
RathiSonika, line 157: Stale reference to esp_ext_part_list_bdl_write in the public header (also around line 155). Please replace with esp_mbr_bdl_write (or remove if no longer needed).

esp_ext_part_tables/src/esp_mbr.c
RathiSonika, line 141: On insert failure mid-parse, esp_mbr_parse returns an error but can leave a partially filled list (signature/flags may already be set, some items already inserted). esp_ext_part_list_deep_copy already cleans up with deinit on failure — please do the same here, or document that the caller must esp_ext_part_list_deinit() after any non-OK return.

RathiSonika, line 415: On error here, earlier slots are already committed and leftover slots are not zeroed yet, so an RMW into this buffer can leave a mix of new and stale entries. Overlap/total_size checks also run only after the full rewrite. Please either generate into a local staging buffer and copy into mbr only on success, or document that on any error mbr contents are undefined and must not be written / must be re-read.

esp_ext_part_tables/src/esp_mbr.c
```diff
@@ -323,8 +356,8 @@ esp_err_t esp_mbr_generate(mbr_t *mbr,
     // lands on the first aligned LBA (e.g. sector 2048 for 1 MiB / 512 B).
     uint32_t next_free_lba = 1;
     SLIST_FOREACH(it, &part_list->head, next) {
-        if (i >= MBR_MAX_PARTITION_COUNT) {
-            ESP_LOGW(TAG, "More than %d partitions in the list, only the first %d will be added to the MBR", MBR_MAX_PARTITION_COUNT, MBR_MAX_PARTITION_COUNT);
+        if (i >= ESP_MBR_MAX_PARTITION_COUNT) {
+            ESP_LOGW(TAG, "More than %d partitions in the list, only the first %d will be added to the MBR", ESP_MBR_MAX_PARTITION_COUNT, ESP_MBR_MAX_PARTITION_COUNT);
```
RathiSonika, line 360: Lists with more than 4 partitions are truncated with ESP_OK. At 1.0, prefer returning an error (or document this as a hard limitation). Silent drop is easy to miss on the write path.

PR #839, opened 2026-09-14
fix(led_strip): use official SPI data-out inversion when available

StevenYang77:
## Summary
- SPI backend uses `SPICOMMON_BUSFLAG_DATA_OUT_INV` when provided by the ESP-IDF SPI driver (IDF-15546).
- Older IDF versions retain the previous GPIO-matrix inversion fallback.

## Test plan
- [x] ESP32-H2-DevKitM-1, `invert_out = false`: LED blinks normally.
- [x] ESP32-H2-DevKitM-1, `invert_out = true`, ESP-IDF master & release/v6.0: inverted waveform with physical idle high.
- [x] Logic analyzer: approximately 28.8 μs per frame, 400/800 ns pulses, and approximately 500 ms between frames.


led_strip/idf_component.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.0.3"
+version: "3.0.4"
```
suda-morris: don't need to bump the version for now, let's collect more features before releasing a new version

StevenYang77: Thanks, reverted the version bump. Need changelog entry being remained under the upcoming 3.0.4 section?

StevenYang77: Thanks, reverted the version bump. Need changelog entry remained under the upcoming 3.0.4 section?

led_strip/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.0.4
```
suda-morris:
```suggestion
## Unreleased
```

led_strip/src/led_strip_spi_dev.c
```diff
@@ -190,12 +190,19 @@ esp_err_t led_strip_new_spi_device(const led_strip_config_t *led_config, const l
         .quadhd_io_num = -1,
         .max_transfer_sz = led_config->max_leds * bytes_per_pixel * SPI_BYTES_PER_COLOR_BYTE,
     };
+#ifdef SPICOMMON_BUSFLAG_DATA_OUT_INV
+    if (led_config->flags.invert_out == true) {
+        spi_bus_cfg.flags |= SPICOMMON_BUSFLAG_DATA_OUT_INV;
+        spi_bus_cfg.data_io_default_level = 1; /* keep physical idle high, same as old GPIO invert */
```
suda-morris, line 196: Ah, sorry, it looks counterintuitive. Maybe we should revert back to your original approach, that let SPICOMMON_BUSFLAG_DATA_OUT_INV and data_io_default_level work in parallel 🤔

StevenYang77: updated.

StevenYang77:
@suda-morris PTAL

suda-morris (approved):

PR #843, opened 2026-09-17
perf(dhara): add sequential-locality path cache to FTL map lookups

omnitex:
## Sequential-locality path cache for Dhara FTL lookups

**What:** Adds an opt-in read-path cache to `dhara_map_find()` (`CONFIG_DHARA_MAP_PATH_CACHE`, default `y`) that remembers the last successful trace and lets sequential sector reads sharing a radix-tree prefix skip already-known levels, cutting `dhara_journal_read_meta()` calls. Costs +140 bytes in `struct dhara_map`. Bumps `dhara` to 1.1.0.

**Commits:**
- `perf(dhara): add sequential-locality path cache to FTL map lookups` — the optimization itself (`dhara/dhara/dhara/map.c`, `map.h`), Kconfig toggle, changelog, SBOM/version bump.
- `test(spi_nand_flash): add host-test coverage for dhara path-cache` — black-box regression tests in `spi_nand_flash/host_test` driven only through the public `spi_nand_flash_*` API: same-sector repeat-read hit, invalidation after an intervening write, plus existing overwrite/GC-pressure/journal-wrap correctness checks. Assertions are gated on `CONFIG_DHARA_MAP_PATH_CACHE` so both on/off builds get a meaningful, config-specific check rather than a toggle with no observable effect. Also implements `nand_emul_get_stats()` in the Linux mmap NAND emulation (declared but previously unused).

**Notes:**
- The `map.h` Kconfig guard now explicitly pulls in `sdkconfig.h` (`__has_include`) — without it the option was silently always-on regardless of menuconfig, since nothing else in the vendored `dhara` tree reaches ESP-IDF's config header transitively.
- `spi_nand_flash`'s own dependency constraint on `espressif/dhara` (`"1.*"`) already covers 1.1.0; no manifest change needed there.

## TODO
- [x] ~~the host test introduced to specifically exercise Dhara internals, rethink? double check~~ removed the "white box" tests, now only extending `spi_nand_flash` host test
- [x] ~~as Dhara is ours to maintain, do we need the sbom?~~ yeah keep it ig

# Checklist

- [ ] CI passing


RathiSonika:
Please add host-side tests that at-least covers:
1. sequential dhara_map_find (cache hit / prefix skip)
2. write or GC, then find again (must not return the pre-write page)
3. journal wrap (epoch bump / root reuse), then find again

dhara/dhara/dhara/map.h
```diff
@@ -19,6 +19,17 @@
 
 #include "journal.h"
 
+/* Enable sequential-locality shortcut in trace_path().
+ * Caches the last successful read-only path traversal (136 bytes in
```
RathiSonika: Please fix this comment in the public header and the same number in the PR body / commit message. 
The 136-byte figure is short. The new fields are prev_target (4) + prev_path[32] (128) + prev_root (4) + prev_epoch (1), plus 3 bytes of trailing alignment on a 32-bit ABI, so sizeof(struct dhara_map) grows by 140, not 136?

dhara/dhara/dhara/map.h
```diff
@@ -19,6 +19,17 @@
...
+ * dhara_map). Consecutive sector lookups that share a common radix-tree
+ * prefix skip the shared levels entirely, saving up to 31 of 32
+ * dhara_journal_read_meta calls for sequential sector access.
+ * Set to 0 to disable (no RAM overhead, original behaviour).
+ */
+#ifndef DHARA_MAP_PATH_CACHE
+#define DHARA_MAP_PATH_CACHE  1
```
RathiSonika, line 38: Please expose DHARA_MAP_PATH_CACHE as a dhara Kconfig option. The cache grows struct dhara_map by 140 bytes, and the only way to opt out today is a compile macro. Users who want the original RAM footprint should be able to disable it from the component config, with the size called out in the help text (?)

dhara/dhara/dhara/map.h
omnitex: Fixed to mention 140 bytes everywhere, thank you

dhara/dhara/dhara/map.h
omnitex: Exposed as Kconfig, turned on by default, with explanation it can be disabled for reclaiming the RAM, at the cost of read performance

omnitex:
The host side tests are WIP, might need to rethink the approach, how to correctly test.

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
```diff
@@ -788,3 +788,248 @@ TEST_CASE("FTL write/read alternating 0xAA/0x55 pattern", "[ftl][rw][patterns]")
...
+     * so the repeat costs exactly the same as the cold read. */
+    REQUIRE(warm_reads == cold_reads);
+#endif
+
+    free(wbuf);
+    free(rbuf);
+    destroy_ftl_dev(dev);
+#else
+    SUCCEED("CONFIG_NAND_ENABLE_STATS disabled; skipping read-count assertion.");
+#endif
```
omnitex: add comments for which `#if` etc is closed

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
```diff
@@ -788,3 +788,255 @@ TEST_CASE("FTL write/read alternating 0xAA/0x55 pattern", "[ftl][rw][patterns]")
...
+    printf("[path_cache] DHARA_MAP_PATH_CACHE=0 (off): reads_after_invalidation=%zu reads_recached=%zu\n",
+           reads_after_invalidation, reads_recached);
+    REQUIRE(reads_recached == reads_after_invalidation);
+#endif // CONFIG_DHARA_MAP_PATH_CACHE
+
+    free(wbuf);
+    free(rbuf);
+    destroy_ftl_dev(dev);
+#else
+    SUCCEED("CONFIG_NAND_ENABLE_STATS disabled; skipping read-count assertion.");
```
RathiSonika:
```suggestion
    SKIP("CONFIG_NAND_ENABLE_STATS disabled; skipping read-count assertion.");
```

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
```diff
@@ -788,3 +788,255 @@ TEST_CASE("FTL write/read alternating 0xAA/0x55 pattern", "[ftl][rw][patterns]")
...
+    /* No persistent cache -- trace_path() re-derives the identical walk,
+     * so the repeat costs exactly the same as the cold read. */
+    REQUIRE(warm_reads == cold_reads);
+#endif // CONFIG_DHARA_MAP_PATH_CACHE
+
+    free(wbuf);
+    free(rbuf);
+    destroy_ftl_dev(dev);
+#else
+    SUCCEED("CONFIG_NAND_ENABLE_STATS disabled; skipping read-count assertion.");
```
RathiSonika:
```suggestion
    SKIP("CONFIG_NAND_ENABLE_STATS disabled; skipping read-count assertion.");
```

dhara/dhara/dhara/map.h
```diff
@@ -19,6 +19,26 @@
...
+ * Kconfig guard below never sees CONFIG_ and always enables the cache.
+ */
+#if !defined(DHARA_MAP_PATH_CACHE) && __has_include("sdkconfig.h")
+#include "sdkconfig.h"
+#endif
+
+#ifndef DHARA_MAP_PATH_CACHE
+#if defined(CONFIG_IDF_TARGET) && !defined(CONFIG_DHARA_MAP_PATH_CACHE)
+/* CONFIG_ bools are only defined when 'y'; absence here means 'n'. */
+#define DHARA_MAP_PATH_CACHE  0
```
RathiSonika: dhara is under the Espressif namespace and we own this component. Can we use CONFIG_DHARA_MAP_PATH_CACHE directly in map.c / map.h instead of mapping it through DHARA_MAP_PATH_CACHE?

With sdkconfig.h already pulled in, #if CONFIG_DHARA_MAP_PATH_CACHE should be enough for IDF builds. The extra macro mainly adds a non-IDF / -DDHARA_MAP_PATH_CACHE= escape hatch — if we do not need that, the shim looks unnecessary.

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
omnitex: changed, ty

spi_nand_flash/host_test/main/test_nand_flash_ftl.cpp
omnitex: changed, ty

dhara/dhara/dhara/map.h
omnitex: Right, removed the shim, clean CONFIG_DHARA_MAP_PATCH_CACHE via sdkconfig.h is enough, thank you

RathiSonika (approved):

PR #844, opened 2026-09-18
thorvg: do not define __linux__ when cross-compiling for ESP-IDF

vanBassum:
## The problem

`thorvg/cross_file.txt.in` passes `-D__linux__` to ThorVG's meson build on every ESP-IDF target.

Two files in ThorVG read that token, and one of them is built here:

- `src/loaders/sfnt/tvgSfntLoader.cpp` takes it to mean mmap is available and includes `<sys/mman.h>`, which ESP-IDF does not have.
- `src/renderer/gpu_engine/gl/tvgGl.cpp` — not built for ESP-IDF.

So `CONFIG_THORVG_TTF_LOADER_SUPPORT=y`, which the Kconfig help actively invites, does not compile:

```
fatal error: sys/mman.h: No such file or directory
```

Without a font loader an SVG `<text>` element renders as nothing at all — silently — so for anyone rendering labels or UI text this is not an optional feature.

## The fix

Drop the one token. ThorVG's font loader then takes its own portable `fopen` path, which is what every other non-Linux target already uses.

## Verification

Built and run on an ESP32-S3 (IDF v6.0) with `CONFIG_THORVG_TTF_LOADER_SUPPORT=y`: TrueType fonts load from a FAT volume via `tvg_font_load_data`, and SVG `<text>` renders. Before this change the build failed inside the dependency.


suda-morris (approved):
Thanks for the fix, LGTM.

PR #821, opened 2026-09-02
feat(spi_nand_flash): refine GigaDevice ECC status with the ECCSE field (IEC-599)

ljodal:
# Checklist

- [ ] Component contains License
- [ ] Component contains README.md
- [ ] Component contains idf_component.yml file with `url` field defined
- [ ] Component was added to [upload job](https://github.com/espressif/idf-extra-components/blob/master/.github/workflows/upload_component.yml#L18)
- [ ] Component was added to [build job](https://github.com/espressif/idf-extra-components/blob/master/test_app/CMakeLists.txt#L8)
- [ ] _Optional:_ Component contains unit tests
- [ ] CI passing

# Change description
On the GD5F4GM8 the 2-bit ECCS field reports 01b for anything from 1 to 7 corrected bits and 11b for exactly 8, so the data-refresh threshold of 4 bits only triggers once every ECC bit is already in use. The ECCSE field in status register F0h refines ECCS=01b (00b = 1-4, 01b = 5, 10b = 6, 11b = 7 corrected bits), letting the driver rewrite decaying pages from 5 corrected bits onward.

Enabled only for the GD5F4GM8 device IDs (95h/85h), the parts this was verified against. Other GM-series devices likely share the encoding but are left untouched.

Depends on espressif/idf-extra-components#813 to parse the ECCS bits correctly

PR #845, opened 2026-09-18
refactor(spi_nand_flash): decode GigaDevice ECCSE via a per-chip decoder

omnitex:
## Description

Follow-up to #821, refactoring the overall approach to ECC status decoding, making it easier to maintain, add vendor/model specific behavior and correctly map the GigaDevice extended status to driver page refreshing mechanism threshold.

### What changed

- Replaced the `has_ecc_status_extension` flag and inline refine helper with a
  `decode_ecc_status` function pointer on `spi_nand_flash_device_t`, so each
  chip owns its own C0h ECC status decoder.
- Default decoder stays 2-bit C0h packing; Micron and FM use 3-bit.
- GD5F4GM8 (device IDs 95h/85h) gets a dedicated decoder that dispatches on
  the C0h ECCS prefix and reads F0h (ECCSE) only when ECCS is `01b`, reporting
  exact 5/6/7/8 corrected-bit counts instead of the generic 4-6 bucket.
- Added `nand_ecc_min_bits_corrected()` to replace the hand-rolled threshold
  if/else chain in `nand_ecc_exceeds_data_refresh_threshold()`.
- Added unit tests (`test_nand_ecc_decode.cpp`) covering 2-bit/3-bit packing,
  GD prefix dispatch, and the F0h-read-failure fallback path.

## TODOs
- [ ] check all gigadevice chips to determine if the extended status is applicable in general to gigadevice, or is chip model specific

### Gigadevice IDs to check for extended ECC status
- [x] `0x51` — GD5F1GQ5
- [x] `0x41` — GD5F1GQ5 (variant)
- [ ] `0x31` —
- [ ] `0x21` — 
- [x] `0x91` — GD5F1GQ5 (variant)
- [x] `0x81` — GD5F1GQ5 (variant)
- [ ] `0x32` — 
- [ ] `0x22` — 
- [x] `0x52` — GD5F2GQ5
- [x] `0x42` — GD5F2GQ5 (variant)
- [x] `0x92` — GD5F2GM7
- [x] `0x82` — GD5F2GM7 (variant)
- [ ] `0x35` — 
- [ ] `0x25` — 
- [x] `0x55` — GD5F4GQ6
- [x] `0x45` — GD5F4GQ6 (variant)
- [x] `0x95` — GD5F4GM8 (already enabled in current commit)
- [x] `0x85` — GD5F4GM8 (variant, already enabled in current commit)
- [x] `0x94` — GD5F4GM7UExxG
- [x] `0x99` — GD5F8GM8UExxG

# Checklist

- [ ] CI passing


spi_nand_flash/src/nand_ecc_decode.c
```diff
@@ -0,0 +1,124 @@
...
+/* 4-bit/528B internal-ECC-strength family: ECCSE resolves 1/2/3/4 exactly.
+ * ECCS=11b is reserved/invalid for this family. */
+static const nand_ecc_status_t s_ecc_gd_eccse_4bit_strength_map[4] = {
+    [0] = NAND_ECC_1_BIT_CORRECTED,
+    [1] = NAND_ECC_2_BITS_CORRECTED,
+    [2] = NAND_ECC_3_BITS_CORRECTED,
+    [3] = NAND_ECC_4_BITS_CORRECTED,
+};
+
+nand_ecc_status_t nand_ecc_pack_decode_2bit(uint8_t status_c0)
```
RathiSonika: nand_ecc_pack_decode_2bit / _3bit look like test-facing wrappers over the same maps as nand_ecc_decode_2bit / _3bit. Can we keep only the device-bound pair (for ecc_status_decoder) and make pack_* static in nand_ecc_decode.c, or drop them and have tests call nand_ecc_decode_*? Also, should nand_ecc_decode.h live under PRIV_INCLUDE_DIRS so none of this widens the public surface?

spi_nand_flash/src/nand_ecc_decode.c
```diff
@@ -0,0 +1,124 @@
...
+{
+    return s_ecc_3bit_status_map[PACK_3BITS_STATUS(status_c0, STAT_ECC2, STAT_ECC1, STAT_ECC0)];
+}
+
+bool nand_ecc_gd_needs_status_ext(uint8_t status_c0)
+{
+    return PACK_2BITS_STATUS(status_c0, STAT_ECC1, STAT_ECC0) == 1;
+}
+
+nand_ecc_status_t nand_ecc_decode_gd_eccse_8bit_strength(uint8_t raw_status_c0, uint8_t raw_status_f0, bool f0_read_ok)
```
RathiSonika: nand_ecc_decode.c is named/used as a generic ECC status decode module, but it also contains GD-specific pieces (nand_ecc_gd_needs_status_ext, nand_ecc_decode_gd_eccse_*, and the GD ECCSE maps). As far as I can tell that F0h/ECCS==01b scheme is GigaDevice-specific (other vendors use the shared 2-bit/3-bit paths, or still need follow-up verification). Should we move the GD maps + nand_ecc_decode_gd_* / nand_ecc_gd_needs_status_ext out of the generic decode module (e.g. into nand_gigadevice.c or a GD-only decode file), and keep nand_ecc_decode.c limited to the shared 2-bit/3-bit decoders?

spi_nand_flash/include/nand_device_types.h
```diff
@@ -7,30 +7,86 @@
...
+    NAND_ECC_7_8_BITS_CORRECTED,           /*!< 7-8 bits corrected */
+    NAND_ECC_1_TO_4_BITS_CORRECTED,        /*!< 1-4 bits corrected (GigaDevice 8-bit/528B-strength ECCSE: cannot resolve further) */
+    NAND_ECC_5_BITS_CORRECTED,             /*!< exactly 5 bits corrected (GigaDevice 8-bit/528B-strength ECCSE) */
+    NAND_ECC_6_BITS_CORRECTED,             /*!< exactly 6 bits corrected (GigaDevice 8-bit/528B-strength ECCSE) */
+    NAND_ECC_7_BITS_CORRECTED,             /*!< exactly 7 bits corrected (GigaDevice 8-bit/528B-strength ECCSE) */
+    NAND_ECC_8_BITS_CORRECTED,             /*!< exactly 8 bits corrected (GigaDevice 8-bit/528B-strength ECCSE) */
+    NAND_ECC_1_BIT_CORRECTED,              /*!< exactly 1 bit corrected (GigaDevice 4-bit/528B-strength ECCSE) */
+    NAND_ECC_2_BITS_CORRECTED,             /*!< exactly 2 bits corrected (GigaDevice 4-bit/528B-strength ECCSE) */
+    NAND_ECC_3_BITS_CORRECTED,             /*!< exactly 3 bits corrected (GigaDevice 4-bit/528B-strength ECCSE) */
+    NAND_ECC_4_BITS_CORRECTED,             /*!< exactly 4 bits corrected (GigaDevice 4-bit/528B-strength ECCSE) */
```
RathiSonika: Nit: since nand_ecc_status_t values are explicitly unordered IDs, could we list the new exact statuses as 1,2,3,4,5,6,7,8 (after the legacy range values) for easier reading? No functional change intended.

spi_nand_flash/include/nand_device_types.h
```diff
@@ -7,30 +7,86 @@
...
 
+/**
+ * @brief Highest corrected-bit count implied by a status class (worst case within the range).
+ *
+ * Used to decide whether a page's ECC status may have reached a configurable refresh
+ * threshold. Comparing against the range's max (rather than its min) is the safe choice for
+ * an arbitrary/tunable threshold: if the threshold falls inside a range, treating it as
+ * "exceeded" risks only an unnecessary extra refresh, whereas comparing against the min would
+ * risk silently skipping a refresh that was actually needed.
+ */
```
RathiSonika: Nit: this comment can be a bit tighter, e.g.
```suggestion
 /**
 * @brief Worst-case corrected-bit count for a status class (range ceiling).
 *
 * Refresh compares this to the tunable threshold. Using max (not min) may
 * refresh early if the threshold sits inside a range; using min can skip a
 * needed refresh.
 */
```

Just a suggestion — feel free to change or ignore.

spi_nand_flash/src/devices/nand_gigadevice.c
```diff
@@ -11,9 +11,41 @@
...
+
+static nand_ecc_status_t gd_decode_ecc_status_8bit_strength(spi_nand_flash_device_t *dev, uint8_t status_c0)
+{
+    uint8_t ext = 0;
+    bool f0_ok = false;
+
+    if (nand_ecc_gd_needs_status_ext(status_c0)) {
+        f0_ok = (spi_nand_read_register(dev, REG_STATUS_EXT, &ext) == ESP_OK);
+        if (!f0_ok) {
+            ESP_LOGW(TAG, "%s: failed to read ECC status extension register", __func__);
```
RathiSonika: When spi_nand_read_register() fails for F0h (REG_STATUS_EXT), the decoder still returns NAND_ECC_4_TO_6_BITS_CORRECTED / NAND_ECC_4_BITS_CORRECTED. That is not the real ECC status — it is a fail-soft stand-in to force refresh. Please confirm. For register read failure it should return failure error code.

Same applies to gd_decode_ecc_status_4bit_strength

spi_nand_flash/src/devices/nand_gigadevice.c
```diff
@@ -11,9 +11,41 @@
 #include "nand.h"
 #include "spi_nand_oper.h"
 #include "nand_flash_devices.h"
+#include "nand_private/nand_ecc_decode.h"
 
 static const char *TAG = "nand_gigadevice";
 
+/* GD5F4GM8 ECCSE lives in feature register F0h bits [5:4]; only meaningful when C0h ECCS is 01b. */
+#define REG_STATUS_EXT  0xF0
```
RathiSonika: Is F0h/ECCSE considered GD-only for now? If yes, defining REG_STATUS_EXT in nand_gigadevice.c looks right. If we expect other vendors to share the same meaning, it may belong in spi_nand_oper.h — but only once that common meaning is confirmed.

spi_nand_flash/include/nand_private/nand_ecc_decode.h
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * True when C0h ECCS is 01b, the only prefix where F0h ECCSE is meaningful.
+ */
+bool nand_ecc_gd_needs_status_ext(uint8_t status_c0);
+
+/**
+ * GigaDevice ECCS+ECCSE prefix dispatch into nand_ecc_status_t, for the
+ * 8-bit/528B internal-ECC-strength family:
+ * GD5F1GM7/GD5F2GM7/GD5F4GM8/GD5F2GQ5xExxH. ECCSE collapses the 1-4 bit range into
```
RathiSonika: Nit: please drop the specific part list from this docstring (e.g. GD5F1GM7/...). Describe only the ECCS/ECCSE encoding scheme here. Which chips use this decoder should stay in nand_gigadevice.c, so we do not have to update this comment every time a device is added.

Please update the comment for nand_ecc_decode_gd_eccse_4bit_strength as well

RathiSonika (changes requested):

PR #847, opened 2026-09-20
feat(catch2): update Catch2 to v3.16.0

suda-morris:
## Summary
- Update Catch2 from v3.7.0 to upstream v3.16.0 and bump the component/SBOM versions.
- Fix the IDF 5.2/5.3 link failure (`undefined reference to esp_fill_random`) by forcing both `getentropy` and `esp_fill_random` into the binary.
- Keep `catch2-console` target tests on ESP32 and ESP32-C3 only (Xtensa + RISC-V). 
- Move the former on-target `catch2-test` example to `host_test` and keep the console example run on real targets.


igrr:
Just noticed that we don't seem to have build-test-rules for catch2 for some reason...

suda-morris:
> Just noticed that we don't seem to have build-test-rules for catch2 for some reason...

@igrr Done.

